PR #1, opened 2023-09-26
Add RPC service for remote capture

RenfengLiu:


capture_service/trace_mgr.cc
```diff
@@ -0,0 +1,41 @@
...
+#include "trace_mgr.h"
+
+namespace Dive
+{
+namespace
+{
+constexpr uint32_t kNumFrameToTrace = 1;
+}
+
+TraceManager* GetTraceMgr()
```
wangra-google, line 26: would it be better to return TraceManager& instead? always worry about nullptr when i see pointers.

wangra-google (approved):

PR #6, opened 2023-09-26
Get return code from the command

RenfengLiu:


capture_service/log.h
```diff
@@ -19,15 +19,26 @@ limitations under the License.
...
-#    define LOGI(...) printf(__VA_ARGS__)
-#    define LOGD(...) printf(__VA_ARGS__)
+#    define __LOGE(...) printf(__VA_ARGS__)
+#    define __LOGW(...) printf(__VA_ARGS__)
+#    define __LOGI(...) printf(__VA_ARGS__)
+#    define __LOGD(...) printf(__VA_ARGS__)
+#endif
 
+#ifndef NDEBUG
+#    define LOGE(...) __LOGE(__VA_ARGS__)
```
wangra-google, line 35: same for LOGE/W/I in both cases?

scripts/build_android.sh
wangra-google, line 1: this file is empty?

wangra-google: aha

wangra-google (dismissed):

wangra-google (approved):

PR #19, opened 2023-10-19
Fix crash due to unintialized variable

shanminchao:


wangra-google:
oh, i think this fixed is already in main 
https://github.com/google/dive/pull/16/files

shanminchao:
Oh! Well, thank you then :)

PR #23, opened 2023-10-20
change the file extension to rd

wangra-google:
change the file extension to rd


shanminchao (approved):
We may want to support .dive in the future as well, but no need for that now

PR #24, opened 2023-10-23
Refector the script that generates adreno.h

wangra-google:
- use the same funciton to generate union for both pm4 and reg
- fix the case where both bitfield is used to extend bitset for a reg
- use union/enum instead of uint32 for non-buildin types
- [Mesa] fix bitfield of vgt_draw_initiator

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ prebuild/grpc/Debug/lib/grpc_authorization_provider.lib
 prebuild/grpc/Debug/lib/grpc_unsecure.lib
 prebuild/grpc/Debug/lib/libprotobufd.lib
 prebuild/grpc/Debug/lib/libprotocd.lib
+install/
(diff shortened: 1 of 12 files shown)


third_party/mesa/src/freedreno/registers/adreno/adreno_pm4.xml
```diff
@@ -831,8 +831,8 @@ opcode: CP_LOAD_STATE4 (30) (4 dwords)
 <bitset name="vgt_draw_initiator" inline="yes">
 	<bitfield name="PRIM_TYPE" low="0" high="5" type="pc_di_primtype"/>
 	<bitfield name="SOURCE_SELECT" low="6" high="7" type="pc_di_src_sel"/>
-	<bitfield name="VIS_CULL" low="9" high="10" type="pc_di_vis_cull_mode"/>
-	<bitfield name="INDEX_SIZE" pos="11" type="pc_di_index_size"/>
+	<bitfield name="VIS_CULL" low="8" high="9" type="pc_di_vis_cull_mode"/>
```
shanminchao, line 834: Is this a change that should be upstreamed?

shanminchao (approved):

third_party/mesa/src/freedreno/registers/adreno/adreno_pm4.xml
wangra-google: yes, that is the plan.

PR #25, opened 2023-10-24
Update ib count

wangra-google:
update the max ib count to 256

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,49 @@ And on Windows, Open Developer Command Prompt for VS 2022(or 2019) and run
 scripts\build_android.bat
 ```
 
-
-
 It will build both debug and release version of the libraries under `build_android` folder.
 
+### CLI Tool for capture and cleanup
+#### Capture with command line tool for Android applications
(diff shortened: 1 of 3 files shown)


bjoeris (approved):

dive_core/common/emulate_pm4.h
```diff
@@ -191,7 +191,7 @@ class EmulatePM4
     // Keep all emulation state together
     struct EmulateState
     {
-        static const uint32_t kMaxPendingIbs = 100;
```
shanminchao, line 194: Can you give context on this?
The reason it's 100 is because I didn't think set_draw_state ever exceeded 100 count. Are you saying that's possible?

If we're exceeding this # for a non-set-draw-state, then we could be dealing with a bug.

wangra-google: replied in the chatroom, we can continue the discussion there. Benson has more info on the capture.

shanminchao:
PR 32 fixes this issue properly. We can probably close this PR.

PR #27, opened 2023-10-25
Capture the chip ID in libwrap

RenfengLiu:
As I tested, I still have problem for 1 frame capture for the latest commit. So put up this PR to dump the chip ID first as a working solution first. Future PR will fix the problem of using latest commit from freedreno repo.

### DIFF
--- a/dive_core/command_hierarchy.cpp
+++ b/dive_core/command_hierarchy.cpp
@@ -1571,6 +1571,118 @@ void CommandHierarchyCreator::ParseVulkanCallMarker(char    *marker_ptr,
     return;
 }
 
+static const std::string ConvertPrimTypeToStr(pc_di_primtype pt)
+{
+    std::string s;
+    switch (pt)
+    {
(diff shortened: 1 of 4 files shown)


third_party/freedreno/wrap/wrap-util.c
```diff
@@ -106,6 +108,11 @@ void rd_start(const char *name, const char *fmt, ...)
 		 */
 		rd_write_section(RD_GPU_ID, &gpu_id, sizeof(gpu_id));
 	}
+
+	// GOOGLE: Write out the chip id.
+	if(chip_id) {
```
wangra-google, line 113: should here both be chip_id instead of gpu_id?

wangra-google: out of curiosity, why do we add this write_section here? it seems that the latest freedreno change doesn't have this part in this file?

wangra-google (approved):

PR #29, opened 2023-10-26
Update README for capture and cleanup

RenfengLiu:


README.md
```diff
@@ -93,8 +93,45 @@ And on Windows, Open Developer Command Prompt for VS 2022(or 2019) and run
...
+cmake -GNinja ..
+ninja
+ninja install
+```
+
+And the libraries and CLI will be installed under the `install` folder.
+Run `./dive_client_cli  --help` for help.
+
+Examples:
+ - Install the dependencies on device and start the package and do a capture
```
wangra-google, line 119: sorry, a bit confused here:
- `9A221FFAZ004TL` is the device id? how do i get this id?
- the 1st command is to both run the apk and do a capture? if this is the case, how do i decide which frame to capture?
- the 2nd command is to only run the apk without capture but still enable dive layer? What would be the use case for this one?
- can you add some comment on what does `type 1` and `type 0` mean?
Thanks :)

wangra-google: cool, thanks for the info. Great job!

wangra-google (approved):

wangra-google (approved):

PR #30, opened 2023-10-26
Add support of the attribute of "shr"

wangra-google:
- this is used to shift the value, for example GRAS_BIN_CONTROL::BINW

### DIFF
--- a/dive_core/generatePm4Info_adreno.py
+++ b/dive_core/generatePm4Info_adreno.py
@@ -54,7 +54,8 @@ struct RegField {
     uint32_t m_type         : 8;  // ValueType enum
     uint32_t m_enum_handle  : 8;
     uint32_t m_shift        : 6;
-    uint32_t                : 10;
+    uint32_t m_shr          : 5; // used to shift left to extract the "original" value, range: [0, 31]
+    uint32_t                : 5;
     uint64_t m_mask;
     const char* m_name;
 };
@@ -91,6 +92,7 @@ void Pm4InfoInit();
 const char *GetOpCodeString(uint32_t op_code);
 const RegInfo *GetRegInfo(uint32_t reg);
 const RegInfo *GetRegByName(const char *);
+const RegField *GetRegFieldByName(const char *name, const RegInfo *info);
 uint32_t GetRegOffsetByName(const char *name);
 const char *GetEnumString(uint32_t enum_handle, uint32_t val);
 const PacketInfo *GetPacketInfo(uint32_t op_code);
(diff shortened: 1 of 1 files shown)


dive_core/generateAdrenoHeader.py
```diff
@@ -257,19 +263,22 @@ def outputRegUnions(pm4_info_file, a6xx_domain, name, reg, for_pm4, postfix):
...
       pm4_info_file.write(extra_front_tab_str + "\tstruct\n") 
       pm4_info_file.write(extra_front_tab_str + "\t{\n")
-    outputBitfields(pm4_info_file, bitfields, extra_front_tab_str)
+    outputBitfields(pm4_info_file, bitfields, extra_front_tab_str, shr_vars)
 
   if use_bitset_as_type or bitfields:
     pm4_info_file.write(extra_front_tab_str + "\t}bitfields" + postfix + ";\n\n")
 
+  for index, shr_var in enumerate(shr_vars):
+    pm4_info_file.write(extra_front_tab_str + "\tenum {" + shr_var[0] + " = " + shr_var[1] + "};\n")
```
wangra-google, line 280: using enum so that i can define const inside union

shanminchao: Yay! Another one of those things on my list that you're knocking out!

Can you show a before and after of the script results (i.e. code snippet from adreno.h) that shows what your change did?

wangra-google: sure
image

wangra-google: ok, gcc doesn't like having enum declarations within the anonymous union, so i have to move the enum declarations outside, here is with the latest change:
image

shanminchao: Thanks.
Is there a reason this isn't done as part of the script that generates pm4_info.h? That's usually where things like register "type" and other attributes are stored. It seems to be weird to put it into adreno.h instead.

shanminchao: Specifically, I mean adding the field to this struct:

image

wangra-google: good point, i didn't know that we can put shift here. it is a much cleaner solution. will make the change, thanks!

wangra-google: updated the change as you suggested

shanminchao: Nitpicky: Do you mind cleaning up the history of this change?
You have a commit [Add support of the attribute of "shr"] which is reverted back soon after. It'd be better if that first commit didn't exist at all, if possible. Just to keep things clean.

wangra-google: with squash, the history of those 2 commits would be gone?

shanminchao: Oh that's true, if you're planning to squash *and* the latter commit 100% reverts back all the previous redundant code, then the squash would effectively eliminate that.

You can always squash manually on the git side to see if that's the case as well.

shanminchao (approved):

PR #32, opened 2023-10-30
Fix queue size not reset to 0

shanminchao:


wangra-google (approved):

dive_core/common/emulate_pm4.cpp
```diff
@@ -735,6 +735,8 @@ bool EmulatePM4::AdvanceCb(const IMemoryManager &mem_manager,
         uint64_t ib_addr = ((uint64_t)ib_packet.ADDR_HI << 32) | (uint64_t)ib_packet.ADDR_LO;
         IbType   ib_type = (type7_header.opcode == CP_INDIRECT_BUFFER_CHAIN) ? IbType::kChain :
                                                                                IbType::kCall;
+        emu_state_ptr->GetCurIb()->m_ib_queue_index = 0;
```
wangra-google, line 738: actually, would it be better to add a reset function in IbStack to reset those members?

shanminchao: I'm opting to avoid the abstraction here so developers can follow the simple logic without having to jump around
More of a personal preference

PR #35, opened 2023-10-31
[mesa] Update freedreno python script to fix a warning of the header

wangra-google:
[mesa] Update freedreno python script to fix a warning of the generated header

- we dont really care a2xx gpus, and the default value of ~0 is not correct anyways, so set the value to 4294967295 to avoid the warning the prevent compiling the header

### DIFF
--- a/dive_core/command_hierarchy.cpp
+++ b/dive_core/command_hierarchy.cpp
@@ -361,12 +361,12 @@ uint8_t CommandHierarchy::GetPacketNodeOpcode(uint64_t node_index) const
 }
 
 //--------------------------------------------------------------------------------------------------
-bool CommandHierarchy::GetPacketNodeIsCe(uint64_t node_index) const
+uint8_t CommandHierarchy::GetPacketNodeIbLevel(uint64_t node_index) const
 {
     DIVE_ASSERT(node_index < m_nodes.m_aux_info.size());
     DIVE_ASSERT(m_nodes.m_node_type[node_index] == Dive::NodeType::kPacketNode);
     const AuxInfo &info = m_nodes.m_aux_info[node_index];
-    return info.packet_node.m_is_ce_packet;
+    return info.packet_node.m_ib_level;
 }
 //--------------------------------------------------------------------------------------------------
 bool CommandHierarchy::GetRegFieldNodeIsCe(uint64_t node_index) const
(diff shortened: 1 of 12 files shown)


third_party/mesa/src/freedreno/common/freedreno_devices.py
```diff
@@ -204,7 +204,7 @@ def __str__(self):
         gmem_align_w = 32,  gmem_align_h = 32,
         tile_align_w = 32,  tile_align_h = 32,
         tile_max_w   = 512,
-        tile_max_h   = ~0, # TODO
+        tile_max_h   = 0, #GOOGLE, change this to 0 to fix a warning in generated header, original value: ~0, # TODO
```
wangra-google, line 207: i am using the generated header in Dive to check the gpu id (will submit in a different PR), and because of the warning, the compiling failed.

wangra-google: in the generated header, tile_max_h has a type of uint32_t, and ~0 will generate -1, so the warning is implicit converting from int to uint32_t. without my change, it also has #TODO (so i assume it is a temp one), and i assume eventually someone from mesa side will update the value to a correct one.

shanminchao: Setting the max tile height to 0 doesn't cause any issues?

wangra-google: no, it is not used for now. `~0` is already a broken value

shanminchao: Wait, doesn't setting a -1 to a uint32_t essentially sets it to UINT32_MAX? In that case, that means setting an unbounded max tile height?

wangra-google: change the value to 4294967295 which is UINT32_MAX, so should be safe now :)

shanminchao: Nitpicky: Is it possible to set it to 0xffffffff instead or  2^32–1?
Seems less error prone in case you get one of those digits wrong.
Nitpicking because eventually we'll upstream this change, so want to make sure we do the cleanest job possible.

wangra-google: updated the file as suggested. :)

shanminchao (approved):

PR #36, opened 2023-10-31
Set draw state ibs

shanminchao:
Adding all set_draw_state IB packets as children of the set_draw_state packet:

image

### DIFF
--- a/dive_core/command_hierarchy.cpp
+++ b/dive_core/command_hierarchy.cpp
@@ -725,6 +725,10 @@ bool CommandHierarchyCreator::OnIbStart(uint32_t                  submit_index,
                                         const IndirectBufferInfo &ib_info,
                                         IbType                    type)
 {
+    // Make all subsequent shared node parent the actual IB-packet
+    if (m_cur_ib_packet_node_index != UINT64_MAX)
+        m_cur_shared_node_parent_index = m_cur_ib_packet_node_index;
+
     // Create IB description string
     std::ostringstream ib_string_stream;
     if (type == IbType::kNormal)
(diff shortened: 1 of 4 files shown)


dive_core/common/emulate_pm4.cpp
```diff
@@ -860,7 +860,11 @@ bool EmulatePM4::QueueIB(uint64_t      ib_addr,
     // Grab current stack level info
     EmulateState::IbStack *cur_ib_level = emu_state->GetCurIb();
 
-    if (cur_ib_level->m_ib_queue_size >= EmulateState::kMaxPendingIbs)
+    // Queue up only non-0-sized IBs
```
wangra-google, line 863: in what case we have 0 sized ibs?

dive_core/command_hierarchy.cpp
```diff
@@ -755,6 +759,19 @@ bool CommandHierarchyCreator::OnIbStart(uint32_t                  submit_index,
         ib_string_stream << "DrawState IB"
                          << ", Address: 0x" << std::hex << ib_info.m_va_addr
                          << ", Size (DWORDS): " << std::dec << ib_info.m_size_in_dwords;
+
+        // Make all subsequent shared node parent be the relevant group in the set_draw_state packet
```
wangra-google, line 763: sorry still a bit confused: what is shared node parent?

wangra-google (approved):

dive_core/command_hierarchy.cpp
shanminchao: A "shared node" is a node that can have multiple parents (i.e. shared among many different topologies). This is used exclusively for nodes that represent PM4 packets. PM4 packets can be children of IB-nodes, or children of other shared-nodes (e.g. children of INDIRECT_BUFFER packets), or children of draw call nodes, or even children of submit-nodes. So we needed to treat them a little differently.

dive_core/common/emulate_pm4.cpp
wangra-google: good to know, thanks

dive_core/command_hierarchy.cpp
wangra-google: aha, thanks!

PR #37, opened 2023-10-31
use standard 2 core instead of the 16 cores to avoid being charged

wangra-google:
use standard 2 core instead of the 16 cores to avoid being charged for extra fees
we are currently charged for 136$ which is closed to the limit of 150$/m

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ concurrency:
 
 jobs:
   lint:
-    runs-on: ubuntu-22.04-16core
+    runs-on: ubuntu-22.04
     steps:
     - name: Check out code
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -14,7 +14,7 @@ concurrency:
 
 jobs:
   lint:
-    runs-on: ubuntu-22.04-16core
+    runs-on: ubuntu-22.04-2core
```
wangra-google, line 17: @RenfengLiu : is this the right way to change to a 2 core runner? where can i see all the available options? thanks :)

wangra-google: updated the yml file as suggested, now the ci is back to normal, thanks!

PR #38, opened 2023-11-01
Add support of variant

wangra-google:
- support variant of A?xx for registers and bitfield

### DIFF
--- a/dive_core/command_hierarchy.cpp
+++ b/dive_core/command_hierarchy.cpp
@@ -725,6 +725,10 @@ bool CommandHierarchyCreator::OnIbStart(uint32_t                  submit_index,
                                         const IndirectBufferInfo &ib_info,
                                         IbType                    type)
 {
+    // Make all subsequent shared node parent the actual IB-packet
+    if (m_cur_ib_packet_node_index != UINT64_MAX)
+        m_cur_shared_node_parent_index = m_cur_ib_packet_node_index;
+
     // Create IB description string
     std::ostringstream ib_string_stream;
     if (type == IbType::kNormal)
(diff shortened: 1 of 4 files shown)


dive_core/capture_data.cpp
```diff
@@ -955,9 +957,26 @@ CaptureData::LoadResult CaptureData::LoadAdrenoRdFile(std::istream &capture_file
         case RD_PROGRAM:
         case RD_VERT_SHADER:
         case RD_FRAG_SHADER:
-        // case RD_BUFFER_CONTENTS:
+            // case RD_BUFFER_CONTENTS:
```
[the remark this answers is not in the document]
wangra-google: fixed

dive_core/CMakeLists.txt
```diff
@@ -85,20 +117,27 @@ endif()
 file(GLOB_RECURSE COMMON_HDR_FILES "${CMAKE_SOURCE_DIR}/dive_core/common/*.h")
 file(GLOB_RECURSE COMMON_SRC_FILES "${CMAKE_SOURCE_DIR}/dive_core/common/*.cpp")
 
+file(GLOB_RECURSE FREEDRENO_SRC_FILES "${THIRDPARTY_DIRECTORY}/mesa/src/freedreno/common/freedreno_dev_info.c")
```
shanminchao, line 120: For my info and out of curiosity... you did all this modification to CMakeLists.txt in order to support what? Building a header or .c file? And why is it necessary to have that header and/or source file to deal with variants?

wangra-google: to call fd_dev_info to get gpu id based on chip id defined in mesa\src\freedreno\common\freedreno_devices.py

shanminchao: And I take it fd_dev_info is non-trivial code that is best not duplicated within Dive, right?

wangra-google: it is using generated header from python, my worry is that if we duplicate the code, we need to track chip id of new devices. also the code is in c which violates some cpp rules. :( so simpler compiling their code instead of duplicating in Dive.

shanminchao: Okie dokie, just checking!

shanminchao (approved):

PR #41, opened 2023-11-07
Fix bugs with incorrect IB packet hierarchy

shanminchao:
This includes adding support for CP_SET_CTXSWITCH_IB, as well as fixing the bug where sometimes INDIRECT_BUFFER child packets get incorrectly added to their parent nodes

### DIFF
--- a/dive_core/common/emulate_pm4.h
+++ b/dive_core/common/emulate_pm4.h
@@ -177,6 +177,21 @@ private:
 class EmulatePM4
 {
 public:
+    enum IbLevel
+    {
+        kPrimaryRing,
+        kIb1,
+        kIb2,
+        kIb3,
+
+        // There's no IB4 in the GPU, but it's useful to have one in the emulator in case a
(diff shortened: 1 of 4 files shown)


dive_core/common/emulate_pm4.h
```diff
@@ -246,6 +248,7 @@ class EmulatePM4
         IbLevel  m_top_of_stack;
         IbStack *GetCurIb() { return &m_ib_stack[m_top_of_stack]; }
     };
+    static_assert(kMaxNumIbLevel == EmulateState::kTotalIbLevels);
```
wangra-google, line 251: why not just use either one of them if they have the same value?

shanminchao: Oh you're right! I was originally hesitant to expose the enum outside, but it makes sense to in this case. Much better than having the same value duplicated!

wangra-google (approved):

wangra-google (approved):

PR #49, opened 2023-11-10
Add support for CP_LOAD_STATE6_* packets

shanminchao:
This includes showing not only the packet contents, but also any constant data and V#/T#/S# data
Missing support for bindless (no test case!) and showing shader disassembly (probably shouldn't be part of the packet disassembly)

### DIFF
--- a/dive_core/command_hierarchy.cpp
+++ b/dive_core/command_hierarchy.cpp
@@ -981,6 +981,11 @@ bool CommandHierarchyCreator::OnPacket(const IMemoryManager &mem_manager,
     {
         m_cur_ib_packet_node_index = packet_node_index;
     }
+    else if (opcode == CP_LOAD_STATE6 || opcode == CP_LOAD_STATE6_GEOM ||
+             opcode == CP_LOAD_STATE6_FRAG)
+    {
+        AppendLoadStateExtBufferNode(mem_manager, submit_index, va_addr, packet_node_index);
+    }
     // vulkan call NOP packages. Currently contains call parameters(except parameters in array),
     // each call is in one NOP packet.
     else if (opcode == CP_NOP)
@@ -1302,13 +1307,19 @@ uint64_t CommandHierarchyCreator::AddPacketNode(const IMemoryManager &mem_manage
         else
         */
         {
+            // If there are missing packet fields, then output the raw DWORDS directly
+            // Some packets, such as CP_LOAD_STATE6_* handle this explicitly elsewhere
+            bool append_extra_dwords = (type7_header.opcode != CP_LOAD_STATE6 &&
+                                        type7_header.opcode != CP_LOAD_STATE6_GEOM &&
+                                        type7_header.opcode != CP_LOAD_STATE6_FRAG);
+
             const PacketInfo *packet_info_ptr = GetPacketInfo(type7_header.opcode);
             DIVE_ASSERT(packet_info_ptr != nullptr);
             AppendPacketFieldNodes(mem_manager,
                                    submit_index,
                                    va_addr,
-                                   is_ce_packet,
-                                   type7_header,
+                                   type7_header.count,
+                                   append_extra_dwords,
                                    packet_info_ptr,
                                    packet_node_index);
         }
@@ -1990,18 +2001,12 @@ void CommandHierarchyCreator::AppendEventWriteFieldNodes(const IMemoryManager   
 void CommandHierarchyCreator::AppendPacketFieldNodes(const IMemoryManager &mem_manager,
                                                      uint32_t              submit_index,
                                                      uint64_t              va_addr,
-                                                     bool                  is_ce_packet,
-                                                     Pm4Type7Header        type7_header,
+                                                     uint32_t              dword_count,
+                                                     bool                  append_extra_dwords,
                                                      const PacketInfo     *packet_info_ptr,
                                                      uint64_t              packet_node_index,
-                                                     size_t                field_start,
-                                                     size_t                field_last)
+                                                     const char           *prefix)
 {
-    // Do a min(), since field_last defaults to UINT64_MAX
-    size_t end_field = (packet_info_ptr->m_fields.size() < field_last + 1) ?
-                       (uint32_t)packet_info_ptr->m_fields.size() :
-                       field_last + 1;
-
     // Loop through each field and append it to packet
     uint32_t base_dword = 0;  // For tracking non-0 array fields
     uint32_t end_dword = UINT32_MAX;
@@ -2015,7 +2020,7 @@ void CommandHierarchyCreator::AppendPacketFieldNodes(const IMemoryManager &mem_m
         // If this is a packet with arrays, and there are more dwords left,
         // then add a parent node for each index
         uint64_t parent_node_index = packet_node_index;
-        if ((packet_info_ptr->m_max_array_size > 1) && (base_dword < type7_header.count))
+        if ((packet_info_ptr->m_max_array_size > 1) && (base_dword < dword_count))
         {
             std::ostringstream field_string_stream;
             field_string_stream << array;
@@ -2032,7 +2037,7 @@ void CommandHierarchyCreator::AppendPacketFieldNodes(const IMemoryManager &mem_m
             parent_node_index = array_node_index;
         }
 
-        for (size_t field = field_start; field < end_field; ++field)
+        for (size_t field = 0; field < packet_info_ptr->m_fields.size(); ++field)
         {
             const PacketField &packet_field = packet_info_ptr->m_fields[field];
 
@@ -2041,7 +2046,7 @@ void CommandHierarchyCreator::AppendPacketFieldNodes(const IMemoryManager &mem_m
 
             // Some packets end early sometimes and do not use all fields (e.g. CP_EVENT_WRITE with
             // CACHE_CLEAN)
-            if (field_dword > type7_header.count)
+            if (field_dword > dword_count)
             {
                 packet_end_early = true;
                 break;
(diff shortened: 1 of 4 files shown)


shanminchao:
Screenshots, 1st showing constant data, the 2nd one showing UBOs:
image
image

wangra-google:
> Screenshots, 1st showing constant data, the 2nd one showing UBOs: image image

in the 2nd screenshot, each elem in the array is a ubo (with address and size)? where there are some ones with size of 0x0?

dive_core/command_hierarchy.cpp
```diff
@@ -2089,32 +2093,157 @@ void CommandHierarchyCreator::AppendPacketFieldNodes(const IMemoryManager &mem_m
...
+                CommandHierarchy::AuxInfo aux_info = CommandHierarchy::AuxInfo::RegFieldNode(false);
+                uint64_t                  field_node_index = AddNode(NodeType::kFieldNode,
+                                                    field_string_stream.str(),
+                                                    aux_info);
+
+                // Add it as child to packet_node
+                AddChild(CommandHierarchy::kEngineTopology, packet_node_index, field_node_index);
+                AddChild(CommandHierarchy::kSubmitTopology, packet_node_index, field_node_index);
+                AddChild(CommandHierarchy::kAllEventTopology, packet_node_index, field_node_index);
+                AddChild(CommandHierarchy::kRgpTopology, packet_node_index, field_node_index);
```
wangra-google, line 2124: not specific to your change, but i think we should remove those unused topology asap. i guess `kRgpTopology` is not used for sure, how about `kEngineTopology` and `kSubmitTopology`?

dive_core/command_hierarchy.cpp
```diff
@@ -2089,32 +2093,157 @@ void CommandHierarchyCreator::AppendPacketFieldNodes(const IMemoryManager &mem_m
...
+    case SB6_CS_TEX: cat = StateBlockCat::kTex; break;
+    case SB6_VS_SHADER:
+    case SB6_HS_SHADER:
+    case SB6_DS_SHADER:
+    case SB6_GS_SHADER:
+    case SB6_FS_SHADER:
+    case SB6_CS_SHADER: cat = StateBlockCat::kShader; break;
+    case SB6_IBO:
+    case SB6_CS_IBO:
+        DIVE_ASSERT("These seem to be specific to pre-A6XX");
```
wangra-google, line 2162: oh, i didn't know that we are allowed to use DIVE_ASSERT this way. did a quick search, this seems to be the only place that use it as warning message?

dive_core/command_hierarchy.cpp
```diff
@@ -2089,32 +2093,157 @@ void CommandHierarchyCreator::AppendPacketFieldNodes(const IMemoryManager &mem_m
...
+    {
+    case SS6_DIRECT: ext_src_addr = va_addr + sizeof(PM4_CP_LOAD_STATE6); break;
+    case SS6_BINDLESS: DIVE_ASSERT("Not supported yet"); break;
+    case SS6_INDIRECT:
+        ext_src_addr = packet.u32All1 & 0xfffffffc;
+        ext_src_addr |= ((uint64_t)packet.u32All2) << 32;
+        break;
+    }
+
+#define APPEND_SHARPS(Type)                                                  \
```
wangra-google, line 2180: is there a specific reason why you prefer macro here? sometimes i found it is a bit more difficult to debug macros. maybe better with lambda expression for a short func like this?

dive_core/command_hierarchy.cpp
```diff
@@ -2089,32 +2093,157 @@ void CommandHierarchyCreator::AppendPacketFieldNodes(const IMemoryManager &mem_m
...
+                AddChild(CommandHierarchy::kAllEventTopology, packet_node_index, const_node_index);
+                AddChild(CommandHierarchy::kRgpTopology, packet_node_index, const_node_index);
+            }
+        }
+        else if (packet.bitfields0.STATE_TYPE == ST6_UBO)
+            APPEND_SHARPS("A6XX_UBO")
+        else if (packet.bitfields0.STATE_TYPE == ST6_IBO)
+        {
+            DIVE_ASSERT(packet.bitfields0.STATE_BLOCK == SB6_CS_SHADER);
+            APPEND_SHARPS("A6XX_TEX_CONST")
```
wangra-google, line 2247: why do we append A6XX_TEX_CONST when the state type is IBO? by the way, what is A6XX_TEX_CONST? texture constant?

shanminchao:
> > Screenshots, 1st showing constant data, the 2nd one showing UBOs: image image
> 
> in the 2nd screenshot, each elem in the array is a ubo (with address and size)? where there are some ones with size of 0x0?

My (limited) understanding is that it's tied in with how the descriptor sets are setup. So if a descriptor set has a valid V# on the 0th, 2nd, and 3rd entry but none in the 1st entry, then the one in the 1st entry is usually not set. However, CP_LOAD_STATE6_* sets an array of them, which means if they want to setup 3 non-consecutive V#s, then they need to pad some of the in-between V#s they may not care about. This is more of a semi-educated guess.

wangra-google:
> This includes showing not only the packet contents, but also any constant data and V#/T#/S# data Missing support for bindless (no test case!) and showing shader disassembly (probably shouldn't be part of the packet disassembly)

more general question, does adreno keeps a heap that has all descriptors (V#/T#/S#) and use `CP_LOAD_STATE6_` to indicate which descriptor is used by current draw/dispatch?

dive_core/command_hierarchy.cpp
shanminchao: Lambda tends to be difficult to debug and slow due to Windows-STL issues. So I tend to avoid lambdas for short code snippets. But you're right, there's enough code here that it might make sense to so we can step into the code if needed.

dive_core/command_hierarchy.cpp
shanminchao: A6XX_TEX_CONST is a T#. IBO probably stands for Image Buffer Object?
I'm following cffdec.c's logic, and my reading of the code shows that IBO is only used when state-block is set to SB6_CS_SHADER. Here's the relevant code snippets:

`
      // SB4_CS_SHADER:
      [0xd][0] = {MESA_SHADER_COMPUTE, SHADER_PROG},
      [0xd][1] = {MESA_SHADER_COMPUTE, SHADER_CONST},
      [0xd][2] = {MESA_SHADER_COMPUTE, UBO},
      [0xd][3] = {MESA_SHADER_COMPUTE, SSBO_0}, /* a6xx location */
  ...
   case SSBO_0: {
      uint32_t *ssboconst = (uint32_t *)contents;

      for (i = 0; i < num_unit; i++) {
         int sz = 4;
         if (options->info->chip == 4) {
            dump_domain(ssboconst, 4, level + 2, "A4XX_SSBO_0");
         } else if (options->info->chip == 5) {
            dump_domain(ssboconst, 4, level + 2, "A5XX_SSBO_0");
         } else if ((6 <= options->info->chip) && (options->info->chip < 8)) {
            sz = 16;
            dump_domain(ssboconst, 16, level + 2, "A6XX_TEX_CONST");
         }
         dump_hex(ssboconst, sz, level + 1);
         ssboconst += sz;
      }
      break;
   }
`
The idea is that there's a lookup table, where [0xd][3] is where SB6_CS_SHADER=0xd and ST6_IBO=0x3. The case statements shows how it's used, and notice for 6xx and 7xx the SSBOs are in the form of T#s.

dive_core/command_hierarchy.cpp
shanminchao: Done!

dive_core/command_hierarchy.cpp
shanminchao: What do you mean by "this way"? I'm adding an assert there because I don't ever expect a command buffer with these parameters. But I still want to catch them just-in-case. Do you prefer I use something else instead?

dive_core/command_hierarchy.cpp
shanminchao: kEngineTopology is used if we have different queues. It's a queue-centric view. But yes, at the very very minimum, we shouldn't call it an "engine" the way AMD driver does. But I'm leaning towards removing it altogether.

I'm definitely using submit topology. When you select the "Submit" mode, that's what you're seeing.

dive_core/command_hierarchy.cpp
wangra-google: so should be `DIVE_ASSERT(false, "These seem to be specific to pre-A6XX");`?

dive_core/command_hierarchy.cpp
wangra-google: what is the difference between `kSubmitTopology` and `kAllEventTopology`?

dive_core/command_hierarchy.cpp
wangra-google: good to know, thanks!

dive_core/command_hierarchy.cpp
wangra-google: sorry maybe i am not seeing the right revision, but it seems that the macro is still there

dive_core/command_hierarchy.cpp
shanminchao: Submit topology lets you see a list of IBs. All-Event lets you see a list of all the GPU events (draws, dmas, etc).

dive_core/command_hierarchy.cpp
shanminchao: Oh oops, you're right!

dive_core/command_hierarchy.cpp
shanminchao: I must've messed up. Will upload again tonight.

dive_core/command_hierarchy.cpp
wangra-google: ah, i see, thanks!

PR #52, opened 2023-11-17
Add UBWC tracking

wangra-google:
Note that this only track UBWC states on render targets and depth stencil, it does not track input resources like textures. The ubwc flag for input resources are part of descriptors, we may need to figure out how to get all descriptors that are used by current draw/dispatch, and show all resources in the UI 
image

Also move gpu variants from RegInfo to part of the key of g_sRegInfo. This is required for cases like RB_DEPTH_BUFFER_INFO, A6XX and A7XX+ have different bitfields but share the same register
image
image

### DIFF
--- a/capture_service/android_application.cc
+++ b/capture_service/android_application.cc
@@ -18,8 +18,10 @@ limitations under the License.
 
 #include "absl/status/status.h"
 #include "absl/strings/match.h"
+#include "absl/strings/str_cat.h"
 #include "absl/strings/str_format.h"
 #include "absl/strings/str_split.h"
+#include "absl/strings/string_view.h"
 #include "constants.h"
 #include "device_mgr.h"
 #include "log.h"
@@ -135,7 +137,7 @@ absl::Status AndroidApplication::Start()
     RETURN_IF_ERROR(m_dev.Adb().Run("shell input keyevent KEYCODE_WAKEUP"));
     RETURN_IF_ERROR(
     m_dev.Adb().Run(absl::StrFormat("shell am start -S -W %s/%s", m_package, m_main_activity)));
-    m_started = true;
+    m_started = IsRunning();
     return absl::OkStatus();
 }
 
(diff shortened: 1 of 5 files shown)


dive_core/event_state.h
wangra-google, line 1: this file is generated from event_state.json

dive_core/event_state.cpp
wangra-google, line 1: this file is generated from event_state.json

dive_core/data_core.cpp
```diff
@@ -772,6 +772,42 @@ void CaptureMetadataCreator::FillHardwareSpecificStates(EventStateInfo::Iterator
...
+    DIVE_ASSERT(rb_mrt_buf_info_reg_start != kInvalidRegOffset);
+    while (rt_id < 8 && m_state_tracker.IsRegSet(rb_mrt_buf_info_reg_start + kElemCount * rt_id))
+    {
+        RB_MRT_BUF_INFO rb_mrt_buf_info;
+        const uint32_t  mrt_reg = rb_mrt_buf_info_reg_start + kElemCount * rt_id;
+        rb_mrt_buf_info.u32All = m_state_tracker.GetRegValue(mrt_reg);
+        event_state_it->SetUBWCEnabled(rt_id, rb_mrt_buf_info.bitfields.COLOR_TILE_MODE == TILE6_3);
+        if (GetGPUVariantType() == kA7XX)
+        {
+            event_state_it->SetUBWCLosslessEnabled(rt_id,
```
wangra-google, line 792: lossy compression is supported in UBWC3.0 which is added since A640, so in theory we could enable these for A6xx, but only enable for a7xx for now just to be consistent with the bitfield defined in turnip

dive_core/event_state.natvis
wangra-google, line 1: this file is generated from event_state.json

dive_core/event_state.cpp
```diff
@@ -380,6 +400,16 @@ template<> EventStateInfo::Iterator EventStateInfoT<EventStateInfo_CONFIG>::Add(
     new (ThreadSizePtr(Id(m_size))) a6xx_threadsize();
     new (EnableAllHelperLanesPtr(Id(m_size))) bool();
     new (EnablePartialHelperLanesPtr(Id(m_size))) bool();
+    for (uint32_t attachment = 0; attachment < 8; ++attachment)
```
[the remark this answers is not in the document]
wangra-google: 8 is the attachment count (seems to be always 8 on a7xx and a6xx) Changed 8 to kAttachmentCount to be clearer

PR #55, opened 2023-11-21
Enable trace by duration

RenfengLiu:
For application that does't have swapchain, there's no way to intercept the present call. Adding trace by duration instead of by frame boundary.

### DIFF
--- a/capture_service/android_application.cc
+++ b/capture_service/android_application.cc
@@ -117,7 +117,6 @@ AndroidApplication::AndroidApplication(AndroidDevice  &dev,
     m_type(type),
     m_started(false)
 {
-    ParsePackage().IgnoreError();
 }
 
 absl::Status AndroidApplication::ParsePackage()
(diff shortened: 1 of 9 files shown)


capture_service/android_trace_mgr.cc
```diff
@@ -48,6 +50,47 @@ void AndroidTraceManager::TriggerTrace()
...
+    SetCaptureName(path.c_str(), num.c_str());
+    {
+        absl::MutexLock lock(&m_state_lock);
+        m_state = TraceState::Triggered;
+    }
+    SetTraceFilePath(std::string(full_path));
+
+    {
+        absl::MutexLock lock(&m_state_lock);
+        SetCaptureState(1);
```
bjoeris, line 69: Do we know how safe it is to start and stop the trace while the application is potentially recording or submitting command buffers in another thread?

capture_service/trace_mgr.h
```diff
@@ -66,6 +66,8 @@ class AndroidTraceManager : public TraceManager
     virtual void WaitForTraceDone() override;
 
 private:
+    void               TraceByFrame();
+    void               TraceByDuration();
     bool               ShouldStartTrace();
```
wangra-google, line 71: not related to this change, but should "ShouldStartTrace()" and "ShouldStopTrace()" both be `const` since they don't change any member?

wangra-google (approved):

PR #57, opened 2023-11-27
UI fix: Don't setup the first device by default

RenfengLiu:


cli/fuzz_main.cpp
```diff
@@ -29,7 +30,9 @@ class membuf : public std::basic_streambuf<char>
 class memstream : public std::istream
 {
 public:
-    memstream(const uint8_t *p, size_t l) : std::istream(&_buffer), _buffer(p, l)
+    memstream(const uint8_t *p, size_t l) :
+        std::istream(&_buffer),
+        _buffer(p, l)
```
wangra-google, line 35: not from this change, but maybe better with the Dive naming convention of member name? (m_buffer)

wangra-google (approved):

PR #59, opened 2023-11-27
Minor fixes

RenfengLiu:


capture_service/android_application.cc
```diff
@@ -146,6 +148,22 @@ absl::Status AndroidApplication::Stop()
     return absl::OkStatus();
 }
 
+bool AndroidApplication::IsRunning() const
+{
+    return IsProcessRunning(m_package);
+}
+
+bool AndroidApplication::IsProcessRunning(absl::string_view process_name) const
```
wangra-google, line 156: do we need IsProcessRunning() or just IsRunning() that contains all code from IsProcessRunning should be fine? (Sorry i couldn't find anywhere else calling IsProcessRunning)

wangra-google: sorry still a bit confused, APK and CLI can have different impl of IsRunning()? the different args are cleas members anyways? (not really a bit deal, just want to make sure i understand it correctly :) )

wangra-google (approved):

PR #61, opened 2023-12-01
UI improvement 

RenfengLiu:
- Put download file in a separate thread so the UI won't frozen.
- Show progress bar
- Show refresh button


wangra-google:
> * Put download file in a separate thread so the UI won't frozen.
> * Show progress bar
> * Show refresh button

This is really cool! thanks for adding this :D

wangra-google (approved):

PR #62, opened 2023-12-02
Add support of SS6_BINDLESS with some bug fixes

wangra-google:
- add support of SS6_BINDLESS, this requires state tracking in CommandHierarchyCreator
- move state tracker to DataCore and allocate from heap since it is too large for stack
- fix a bug of wrong memory block got removed
- fix field offset: skip the intial PM4 and do not use packet_field.m_dword directly as offset
- fix field value with reg_field.m_shr
- use DIVE_VERIFY for all CopyMemory calls
- some code cleanup

### DIFF
--- a/dive_core/command_hierarchy.cpp
+++ b/dive_core/command_hierarchy.cpp
@@ -526,6 +526,13 @@ CommandHierarchy::AuxInfo CommandHierarchy::AuxInfo::SyncNode(SyncType type, Syn
 // =================================================================================================
 // CommandHierarchyCreator
 // =================================================================================================
+CommandHierarchyCreator::CommandHierarchyCreator(EmulateStateTracker &state_tracker) :
+    m_state_tracker(state_tracker)
+{
+    m_state_tracker.Reset();
+}
+
+//--------------------------------------------------------------------------------------------------
 bool CommandHierarchyCreator::CreateTrees(CommandHierarchy  *command_hierarchy_ptr,
                                           const CaptureData &capture_data,
                                           bool               flatten_chain_nodes,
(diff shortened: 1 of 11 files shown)


dive_core/command_hierarchy.cpp
```diff
@@ -1419,7 +1379,8 @@ uint64_t CommandHierarchyCreator::AddRegisterNode(uint32_t       reg,
     for (uint32_t field = 0; field < reg_info_ptr->m_fields.size(); ++field)
     {
         const RegField &reg_field = reg_info_ptr->m_fields[field];
-        uint64_t        field_value = (reg_value & reg_field.m_mask) >> reg_field.m_shift;
+        uint64_t        field_value = (reg_value & reg_field.m_mask) >> reg_field.m_shift
```
wangra-google, line 1382: @shanminchao can you confirm this is correct? i checked the cffdump code, and they use the shr to shift left

dive_core/command_hierarchy.cpp
```diff
@@ -1929,12 +1886,11 @@ void CommandHierarchyCreator::AppendPacketFieldNodes(const IMemoryManager &mem_m
...
             uint32_t dword_value = 0;
-            uint64_t dword_va_addr = va_addr + field_dword * sizeof(uint32_t);
-            bool     ret = mem_manager.CopyMemory(&dword_value,
-                                              submit_index,
-                                              dword_va_addr,
-                                              sizeof(uint32_t));
-            DIVE_VERIFY(ret);  // This should never fail!
+            // (field_dword - 1) since each field is always 1 32bit register, we don't have any
+            // 64bit field
+            uint64_t dword_va_addr = va_addr + (field_dword - 1) * sizeof(uint32_t);
```
wangra-google, line 1891: @shanminchao could you confirm if this is the right fix? m_dword seems to be the number of dwords that including current field whereas here we want to have the offset of current field.

shanminchao: I don't understand what is being fixed here. Before, va_addr pointed to the beginning of the packet, but this PR offsets it by the size of the 32-bit header, and then it's further offseted backwards here with a -1. Was there a specific issue being fixed here?

shanminchao: packet_field.m_dword is supposed to indicate which dword the field is in. I forget if the dword # starts at 0 or 1. But in any case, it was used to calculate an offset

wangra-google: currently, it starts from 1, so if i understand correctly, it is not offset, it is the dword count that includes current field, maybe this is not intended? 
https://github.com/google/dive/blob/e3b9bd69ade10131118f0a482c2988bbc704f843/dive_core/generateAdrenoHeader.py#L133

shanminchao: It's a count, used to calculate the offset.
So for the 1st dword field after the header, you'd calculate the offset by multiplying m_dword with sizeof(uint32_t). It is a count, but a count used to calculate an offset. Hence the original code:

uint32_t dword_va_addr = va_addr + field_dword * sizeof(uint32_t)

wangra-google: the reason for the fix is because `AppendPacketFieldNodes` is called in different places. the place where it is called in the lambda function doesn't have the PM4 packets, to be consistent, i made va_addr does not contain the PM4 packets. (this is why the previous change offset size of 32-bit header) And since va_addr doesn't contain the header 32bit, there is this (field_dword - 1) here

wangra-google: it is confusing, but i could put better comments to explain it better. or do you suggest to fix it in a different way?

shanminchao: Ooohhh I see. So the lambda function never worked before? I could've sworn I tested it! Or was it buggy only in the context of bindless?

So it's a result of a bug fix. I don't think the explanation is lacking; I just didn't understand what you were trying to do, having not noticed the other call site for this function.

wangra-google: I notice the bug in the bindless case when i was checking the content of the descriptors.

dive_core/command_hierarchy.cpp
```diff
@@ -1314,7 +1274,7 @@ uint64_t CommandHierarchyCreator::AddPacketNode(const IMemoryManager &mem_manage
             DIVE_ASSERT(packet_info_ptr != nullptr);
             AppendPacketFieldNodes(mem_manager,
                                    submit_index,
-                                   va_addr,
+                                   va_addr + sizeof(uint32_t),  // skip the type7 PM4
```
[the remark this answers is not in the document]
wangra-google: yeah, sizeof(Pm4Type7Header) is better, updated the code

dive_core/command_hierarchy.cpp
shanminchao: Got it, thanks for fixing!

shanminchao (approved):

PR #64, opened 2023-12-08
CI FIX

RenfengLiu:


wangra-google (approved):
thanks a lot for fixing this. i am confused why sometimes the ci is green, but some other times, it is red. is it because different build machines have different python version installed?

PR #68, opened 2023-12-12
Flatten PM4 hierarchy for event views

shanminchao:
Get rid of all the CP_INDIRECT_BUFFER packets in event views. Flatten the hierarchy. All packets that were children of CP_INDIRECT_BUFFER are added to the event node.

This fixes issue where a CP_INDIRECT_BUFFER that contains lots of draw calls results in a CP_INDIRECT_BUFFER packet shown for the 1st draw, and the rest of the draws contain no PM4 packets.

### DIFF
--- a/ui/command_buffer_model.cpp
+++ b/ui/command_buffer_model.cpp
@@ -55,11 +55,13 @@ void CommandBufferModel::Reset()
 }
 
 //--------------------------------------------------------------------------------------------------
-void CommandBufferModel::SetTopologyToView(const Dive::Topology *topology_ptr)
+void CommandBufferModel::SetTopologyToView(const Dive::Topology *topology_ptr,
+                                           bool                  show_level_column)
 {
     emit beginResetModel();
     m_topology_ptr = topology_ptr;
     m_selected_node_index = UINT64_MAX;
+    m_show_level_column = show_level_column;
     emit endResetModel();
 }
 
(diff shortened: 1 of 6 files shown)


ui/command_buffer_model.cpp
```diff
@@ -140,7 +142,11 @@ QVariant CommandBufferModel::headerData(int section, Qt::Orientation orientation
     if (orientation == Qt::Horizontal && role == Qt::DisplayRole)
     {
         if (section < CommandBufferModel::kColumnCount)
-            return QVariant(tr(CommandBufferColumnNames[section]));
+        {
+            // Don't print out ib-level column if m_show_level_column set to false
+            if (m_show_level_column || (section != kColumnIbLevel))
```
wangra-google, line 147: sorry a bit confused here: should here be (!m_show_level_column || (section != kColumnIbLevel))?

ui/command_buffer_model.h
```diff
@@ -70,4 +70,5 @@ public slots:
     std::map<uint64_t, QModelIndex> m_node_index_to_parent_map;
     const Dive::CommandHierarchy   &m_command_hierarchy;
     const Dive::Topology           *m_topology_ptr = nullptr;
+    bool                            m_show_level_column;
```
wangra-google, line 73: better to be initialized with a default value.

shanminchao: > better to be initialized with a default value.

Done

ui/command_buffer_model.cpp
shanminchao: This if-statement prints out the column name. And it only prints out column name if m_show_level_column is true. So I'm not sure why your if-statement has a "!m_show_level_column". 

The other thing it checks for is that it only prints column name if the column is not Ib-level column. Otherwise it has to check against m_show_level_column.

wangra-google: i see. i was confused because the comment says `Don't print out ib-level column if m_show_level_column set to false`, and I thought the early return here means `don't print out ib-level column` but with m_show_level_column equal to true.

wangra-google (dismissed):

ui/command_buffer_model.cpp
shanminchao: Fair point! Updated comment to be less confusing.

wangra-google (approved):

PR #71, opened 2023-12-15
Show CP_CONTEXT_REG_BUNCH register fields

shanminchao:


dive_core/command_hierarchy.cpp
```diff
@@ -1220,8 +1220,7 @@ uint64_t CommandHierarchyCreator::AddPacketNode(const IMemoryManager &mem_manage
                             packet_node_index);
         }
         else if (type7_header.opcode == Pal::Gfx9::IT_LOAD_UCONFIG_REG ||
-                type7_header.opcode == Pal::Gfx9::IT_LOAD_CONTEXT_REG ||
```
wangra-google, line 1223: is this part of this change? i guess i could do a full path to remove all these gfx9 commented code?

shanminchao: Oh crap did I remove that? Not intentional

shanminchao: Reverted this accidental change

dive_core/command_hierarchy.cpp
```diff
@@ -1764,14 +1773,59 @@ std::string CommandHierarchyCreator::GetEventString(const IMemoryManager &mem_ma
...
+    {
+        struct
+        {
+            uint32_t m_reg_offset;
+            uint32_t m_reg_value;
+        } reg_pair;
+        uint64_t pair_addr = va_addr + i * sizeof(uint32_t);
+        DIVE_VERIFY(mem_manager.CopyMemory(&reg_pair, submit_index, pair_addr, sizeof(reg_pair)));
+
+        const RegInfo *reg_info_ptr = GetRegInfo(reg_pair.m_reg_offset);
```
wangra-google, line 1796: maybe better with a `GetRegInfo` function since this code appears in the other `AppendRegNodes()` as well

wangra-google: Actually the 2 AppendRegNodes() looks very similar, is it possible to combine the 2 into 1 function?

shanminchao: They look similar but are different in fundamental ways. One expects a list of register values and has the offset passed in as a parameter, the other one expects (offset,value) pairs. There can be an attempt at extracting the common parts, but the 2 functions are short enough that I don't think the effort and additional indirection is worth it

wangra-google (approved):

PR #73, opened 2023-12-15
Temp fix of a crash caused by 0 shader address

wangra-google:


wangra-google:
@bjoeris not sure if this is the right fix since I am not sure if it is valid to have the addr to be 0 here. Could you verify if this is the correct fix? the problematic capture file is here:
https://drive.google.com/file/d/1-elAsbEDH2Ww4-zaPlnKLw88IQ2iyhu1/view

bjoeris (approved):

bjoeris:
> @bjoeris not sure if this is the right fix since I am not sure if it is valid to have the addr to be 0 here. Could you verify if this is the correct fix? the problematic capture file is here: https://drive.google.com/file/d/1-elAsbEDH2Ww4-zaPlnKLw88IQ2iyhu1/view

The traces I've tried all had UINT64_MAX for for missing shaders, but I think it's reasonably to also handle 0 as a missing shader.

PR #77, opened 2023-12-19
Add libarchive to third_party

RenfengLiu:
Also add prebuilt binaries on Windows

This is needed to load the compressed version of the trace file.


libarchive.cmake
```diff
@@ -0,0 +1,61 @@
...
+    set(CMAKE_PREFIX_PATH "${CMAKE_SOURCE_DIR}/prebuild/libarchive/")
+endif()
+
+find_package(LibArchive QUIET)
+
+if(NOT LibArchive_FOUND)
+# Run the build script for libarchive
+    if(WIN32)
+        add_custom_command(
+            COMMAND ${CMAKE_SOURCE_DIR}/scripts/build_libarchive.bat
```
bjoeris, line 29: If `build_libarchive.bat` can build libarchive, why do we need the prebuilt version?

bjoeris (commented):
Is there any reason to include the man files?

bjoeris (approved):

PR #82, opened 2024-01-17
Enable getting Perfetto trace

RenfengLiu:
- The feature is controlled by the cmake compilation flag `DIVE_ENABLE_PERFETTO`
- This will trigger a Perfetto trace along with the Dive trace, and two file will be generated/downloaded.
- Added a Perfetto event in the trace to mark the start and stop of Dive trace on the CPU side, which can be used to align the PM4 frame data with Perfetto frame data. 
- Missing some initial setup(`adb shell setprop debug.graphics.gpu.profiler.perfetto 1
`), will be added later.

### DIFF
--- a/capture_service/perfetto_trace.cc
+++ b/capture_service/perfetto_trace.cc
@@ -20,6 +20,7 @@ limitations under the License.
 #include <fcntl.h>
 #include <perfetto.h>
 #include <chrono>
+#include <cstdint>
 #include <string>
 #include <thread>
 
@@ -37,7 +38,9 @@ namespace Dive
 namespace
 {
 // TODO(renfeng): determine how long should be the trace, and pass in the value here.
-int32_t kTraceDurationMs = 5000;
+static constexpr int32_t kTraceDurationMs = 5000;
+static constexpr int32_t kFlushPeriodMs = 1000;
+static constexpr int32_t kBufferSize = 131072;  // 128k
 }  // namespace
 PerfettoTraceManager::PerfettoTraceManager()
 {
@@ -46,7 +49,7 @@ PerfettoTraceManager::PerfettoTraceManager()
 
 void PerfettoTraceManager::InitializePerfetto()
 {
-    // TODO: setprop debug.graphics.gpu.profiler.perfetto 1
+    // TODO(renfeng): setprop debug.graphics.gpu.profiler.perfetto 1
     LOGI("In InitializePerfetto \n");
     perfetto::TracingInitArgs args;
     args.backends = perfetto::kSystemBackend;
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -80,7 +80,11 @@ if(DIVE_NO_DISASSEMBLY)
     add_definitions(-DDIVE_NO_DISASSEMBLY=1)
 endif()
 
+option(DIVE_ENABLE_PERFETTO, "Enalbe functionality to get perfetto trace." ON)
 # Build Perfetto SDK
+if(DIVE_ENABLE_PERFETTO)
+message("DIVE_ENABLE_PERFETTO is enabled.")
+add_compile_definitions(DIVE_ENABLE_PERFETTO=1)
```
wangra-google, line 87: do we need to recompile the entire perfetto project when this is enabled?

capture_service/perfetto_trace.cc
```diff
@@ -0,0 +1,136 @@
...
+
+PERFETTO_TRACK_EVENT_STATIC_STORAGE();
+
+namespace Dive
+{
+
+namespace
+{
+// TODO(renfeng): determine how long should be the trace, and pass in the value here.
+int32_t kTraceDurationMs = 5000;
```
wangra-google, line 40: `static constexpr` ?

capture_service/perfetto_trace.cc
```diff
@@ -0,0 +1,136 @@
...
+
+void PerfettoTraceManager::StartNewSession(const std::string &trace_file_name)
+{
+    LOGI("Begin schedule trace session\n");
+    m_tracing_worker.Schedule([trace_file_name] {
+        LOGI("In StartTracing \n");
+        perfetto::TraceConfig cfg;
+        auto                 *buffers = cfg.add_buffers();
+        cfg.set_flush_period_ms(1000);
+        buffers->set_size_kb(131072);
```
wangra-google, line 69: out of curiosity, where do you get this 131072 (128k)? is it something we might increase in the future?

wangra-google (approved):

capture_service/perfetto_trace.cc
wangra-google: would there be a crash or memory corruption when the size is not big enough?

wangra-google (approved):

capture_service/perfetto_trace.cc
wangra-google: so corruption could potentially happen if the size is not big enough? my worry is that if this happens silently it might be hard for us to find out, but i guess if AGI is using this value, it should be big enough.

PR #87, opened 2024-01-26
Fix the slow capture problem for Dive

RenfengLiu:
Currently Dive captures all GPU buffer, some of which are hundreds of MB, which takes super long time to write out(especially when compressing). Those large buffers are not used anywhere when parsing the trace. Experiments shows that the largest buffer used for command stream is 128K, so we cap the dump size to that number. This will increase the dump speed order of magnitude(from minutes to seconds) for the fish_tornado sample.


wangra-google:
> Currently Dive captures all GPU buffer, some of which are hundreds of MB, which takes super long time to write out(especially when compressing). Those large buffers are not used anywhere when parsing the trace. Experiments shows that the largest buffer used for command stream is 128K, so we cap the dump size to that number. This will increase the dump speed order of magnitude(from minutes to seconds) for the fish_tornado sample.

do we know what are the other large buffers? do we possibly have large buffers for shader more than 128k?

wangra-google (approved):

PR #92, opened 2024-02-08
Load and parse Perfetto trace

RenfengLiu:
This parse the Perfetto trace and store the queried value in data structures that can be used later to show timing information.

### DIFF
--- a/perfetto_trace/gpu_slice_data.cpp
+++ b/perfetto_trace/gpu_slice_data.cpp
@@ -55,28 +55,26 @@ std::vector<int64_t> GpuSliceDataParser::ParseDiveTraceTimestamp()
 std::vector<int64_t> GpuSliceDataParser::ParseSubmissionIds(int64_t start_ts, int64_t end_ts)
 {
     std::vector<int64_t> submission_ids;
-    {
-        std::string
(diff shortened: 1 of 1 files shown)


perfetto_trace/gpu_slice_data.cpp
```diff
@@ -0,0 +1,264 @@
...
+{
+    std::vector<SurfaceData> surface_data;
+    std::vector<int64_t>     start_end_ts = ParseDiveTraceTimestamp();
+    assert(start_end_ts.size() == 2);
+
+    std::vector<int64_t> submission_ids = ParseSubmissionIds(start_end_ts[0], start_end_ts[1]);
+
+    for (auto submission_id : submission_ids)
+    {
+        surface_data.emplace_back(ParseSurfaceData(submission_id));
```
wangra-google, line 258: could surface_data be large? if this field is large, maybe better with passing a & into the function (instead return a copy) and update directly to avoid an extra copy?

wangra-google: oh, that is cool! good to know!

wangra-google (approved):

PR #95, opened 2024-02-22
Add support for tiled rendering pass

wangra-google:
- add proper support of CP_START_BIN&CP_END_BIN
- add proper support of CP_FIXED_STRIDE_DRAW_TABLE
- remove padding in generated pm4 structs to fix PM4_CP_FIXED_STRIDE_DRAW_TABLE
- fix the reg32 that doesn't have type and bitfield
- remove unused marker code

(This is the 1st pass so that Robin can work on correlation with the Perfetto data. I will make some changes to group tiles with some extra info in a different PR)
image

### DIFF
--- a/dive_core/capture_data.cpp
+++ b/dive_core/capture_data.cpp
@@ -1046,10 +1046,6 @@ CaptureData::LoadResult CaptureData::LoadAdrenoRdFile(FileReader &capture_file)
 
     struct BlockInfo
     {
-        // The first 2 dwords are set to 0xffffffff
-        uint32_t m_max_uint32_1;
-        uint32_t m_max_uint32_2;
-
         uint32_t m_block_type;
 
         // The number of bytes that follow this header.
@@ -1060,10 +1056,15 @@ CaptureData::LoadResult CaptureData::LoadAdrenoRdFile(FileReader &capture_file)
     uint64_t  cur_gpu_addr = UINT64_MAX;
     uint32_t  cur_size = UINT32_MAX;
     bool      is_new_submit = false;
+    bool      skip_commands = true;
     while (capture_file.read((char *)&block_info, sizeof(block_info)) > 0)
     {
-        if (block_info.m_max_uint32_1 != 0xffffffff || block_info.m_max_uint32_2 != 0xffffffff)
-            return LoadResult::kCorruptData;
+        // Read and discard any trailing 0xffffffff padding from previous block
+        while (block_info.m_block_type == 0xffffffff && block_info.m_data_size == 0xffffffff)
+        {
+            if (capture_file.read((char *)&block_info, sizeof(block_info)) <= 0)
+                return LoadResult::kCorruptData;
+        }
 
         switch (block_info.m_block_type)
         {
@@ -1076,7 +1077,10 @@ CaptureData::LoadResult CaptureData::LoadAdrenoRdFile(FileReader &capture_file)
             is_new_submit = true;
             break;
         case RD_CMDSTREAM_ADDR:
-            if (!LoadCmdStreamBlockAdreno(capture_file, block_info.m_data_size, is_new_submit))
+            if (!LoadCmdStreamBlockAdreno(capture_file,
+                                          block_info.m_data_size,
+                                          is_new_submit,
+                                          skip_commands))
                 return LoadResult::kFileIoError;
             is_new_submit = false;
             break;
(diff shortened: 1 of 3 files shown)


dive_core/common/emulate_pm4.cpp
```diff
@@ -929,6 +929,114 @@ bool EmulatePM4::AdvanceCb(const IMemoryManager &mem_manager,
...
+        //  |Common_block|
+        //  |CP_END_BIN|
+        // All the prefix blocks are in a separate IB
+        // Prefix_block contains setup for each tile, like viewport/scissors with fixed block size
+        // Common_block contains drawcalls and resolve
+
+        // only add the 1st prefix block here
+        // after this ib, we jump back to the Common_block
+        // once we advance to the CP_END_BIN, we add the rest ibs
+        // it is done this way since we don't know the size of the Common_block at this point
```
shanminchao, line 957: Do we not? The CP_START_BIN packet has a BODY_DWORDS field. Doesn't that field tell you how big the common block is?

dive_core/common/emulate_pm4.cpp
```diff
@@ -929,6 +929,114 @@ bool EmulatePM4::AdvanceCb(const IMemoryManager &mem_manager,
...
+                         bin * m_prefix_block_dword_size * sizeof(uint32_t),
+                         m_prefix_block_dword_size,
+                         false,
+                         IbType::kBinPrefix,
+                         emu_state_ptr))
+            {
+                return false;
+            }
+
+            if (!QueueIB(m_cp_start_bin_address,
```
shanminchao, line 994: The way it's written here, you have a prefix IB followed by a common IB, but only for all tiles except the first one.
The first one, which is covered by the CP_START_BIN, only covers a kBinPrefix but no kBinCommon. So that means there's no kBinCommon called for the first tile? Am I missing something?

dive_core/common/emulate_pm4.cpp
```diff
@@ -929,6 +929,114 @@ bool EmulatePM4::AdvanceCb(const IMemoryManager &mem_manager,
...
+        // CP_FIXED_STRIDE_DRAW_TABLE only availabe at a7xx+
+        // Executes an array of fixed-size command buffers where each buffer is assumed to have one
+        // draw call, skipping buffers with non - visible draw calls.
+        PM4_CP_FIXED_STRIDE_DRAW_TABLE packet;
+        DIVE_VERIFY(mem_manager.CopyMemory(&packet,
+                                           emu_state_ptr->m_submit_index,
+                                           emu_state_ptr->GetCurIb()->m_cur_va,
+                                           sizeof(PM4_CP_FIXED_STRIDE_DRAW_TABLE)));
+
+        for (uint32_t draw = 0; draw < packet.bitfields2.COUNT; ++draw)
```
shanminchao, line 1024: If the CP_START_BIN/CP_END_BIN's common block contains the draw call, then waht is this draw-table? Is this an a7xx-only way to do the draw call? And if so, what's in the common block when using a7xx?

shanminchao:
> * add proper support of CP_START_BIN&CP_END_BIN
> * add proper support of CP_FIXED_STRIDE_DRAW_TABLE
> * remove padding in generated pm4 structs to fix PM4_CP_FIXED_STRIDE_DRAW_TABLE
> * fix the reg32 that doesn't have type and bitfield
> * remove unused marker code
> 
> (This is the 1st pass so that Robin can work on correlation with the Perfetto data. I will make some changes to group tiles with some extra info in a different PR) image

In the picture you provided, there's CCU_RESOLVE calls inside the "Tile Rendering Pass". Is this supposed to be there?

dive_core/common/emulate_pm4.cpp
wangra-google: that filed is always 0 in both 2 test captures. :(

dive_core/common/emulate_pm4.cpp
wangra-google: since i don't know the size of the common block, i need to parse it. Note that all prefix blocks are in 1 IB, but I intentionally break it into multiple ones so that it is easier to parse. when i reach CP_START_BIN, i only add the 1st prefix block as IB, and start to parse the 1st prefix block, once it is done, the parser jump back to the common block since it is right after the IB of all prefix blocks, and now i am at the start of the common block. once i reach CP_END_BIN, i am reaching the end of common block, and now i have the start and size of the common block. at this point, i have already added all the packets for the 1st bin, that is why i skipped the 1st tile.

dive_core/common/emulate_pm4.cpp
wangra-google: PM4_CP_FIXED_STRIDE_DRAW_TABLE is part of the common block. it is available for a7xx and above gpus. for a6xx or older gpus, the drawcalls are just recorded normally without using this packet.

wangra-google:
> > * add proper support of CP_START_BIN&CP_END_BIN
> > * add proper support of CP_FIXED_STRIDE_DRAW_TABLE
> > * remove padding in generated pm4 structs to fix PM4_CP_FIXED_STRIDE_DRAW_TABLE
> > * fix the reg32 that doesn't have type and bitfield
> > * remove unused marker code
> > 
> > (This is the 1st pass so that Robin can work on correlation with the Perfetto data. I will make some changes to group tiles with some extra info in a different PR) image
> 
> In the picture you provided, there's CCU_RESOLVE calls inside the "Tile Rendering Pass". Is this supposed to be there?

yes, it is from CP_EVENT_WRITE7, with a value of `30`, for a5xx a6xx, it is interpreted as `BLIT`, but for a7xx, it is interpreted as `CCU_RESOLVE`.

dive_core/common/emulate_pm4.cpp
shanminchao: Gotcha. Would be good to add that to your comment block as well, for whoever reads it next

dive_core/common/emulate_pm4.cpp
shanminchao: Ahhhh ok. The common block is implicitly parsed as the next packets right after that 1st common block. Makes sense. I missed that part!

dive_core/common/emulate_pm4.cpp
wangra-google: updated the comment as suggested

dive_core/common/emulate_pm4.cpp
wangra-google: updated the comment as suggested

PR #96, opened 2024-02-23
Handle buffers with same address but different size being re-captured

shanminchao:
Previouly we only handled the recapture of same-sized buffer
This fixes a bug I'm finding from a Chromebook capture

### DIFF
--- a/dive_core/capture_data.cpp
+++ b/dive_core/capture_data.cpp
@@ -270,13 +270,26 @@ void MemoryManager::Finalize(bool same_submit_copy_only, bool duplicate_ib_captu
             }
             else
             {
-                if (memory_block.m_va_addr != prev_addr || memory_block.m_data_size != prev_size)
+                // Check if it's a completely new block
+                if (memory_block.m_va_addr >= (prev_addr + prev_size))
                     temp_memory_blocks.push_back(memory_block);
                 else
                 {
-                    // Replace previous memory block with the more updated current version
-                    delete[] temp_memory_blocks.back().m_data_ptr;
-                    temp_memory_blocks.back() = m_memory_blocks[i];
+                    // It's EITHER a new block or it has the same address as the previous block
+                    // (i.e. whole or partial overwrite)
+                    DIVE_ASSERT(memory_block.m_va_addr == prev_addr);
+
+                    // Use whichever one is bigger and get rid of the smaller one
+                    if (memory_block.m_data_size > temp_memory_blocks.back().m_data_size)
+                    {
+                        // Replace previous memory block with current one
+                        delete[] temp_memory_blocks.back().m_data_ptr;
+                        temp_memory_blocks.back() = m_memory_blocks[i];
+                    }
+                    else
+                    {
+                        delete[] m_memory_blocks[i].m_data_ptr;
+                    }
                 }
             }
             prev_addr = memory_block.m_va_addr;


dive_core/capture_data.cpp
```diff
@@ -270,11 +270,35 @@ void MemoryManager::Finalize(bool same_submit_copy_only, bool duplicate_ib_captu
...
+                                    temp_memory_blocks.back().m_data_size;
+                    uint8_t *new_data_block = new uint8_t[size];
+
+                    // Copy earlier block first
+                    memcpy(new_data_block,
+                           temp_memory_blocks.back().m_data_ptr,
+                           temp_memory_blocks.back().m_data_size);
+
+                    // Copy current block
+                    memcpy(new_data_block, memory_block.m_data_ptr, memory_block.m_data_size);
```
wangra-google, line 295: hmm, both memcpy() has the same dst address, so that the 2nd one will overwrite the data?

shanminchao: Yeah that's the idea. The 1st and 2nd blocks might have different sizes, but come in at different times. That's why we do a stable_sort at the top (as opposed to normal sort), and then we write to temp_memory_blocks in the order they appear. 
In the past, we just get rid of the earlier data blocks, since we've only seen allocations of the same addr/size pair. But now I have a capture where the latter memory block starts at the same addr but has a smaller size.

wangra-google: in that case should we always just copy the one with larger size? (assuming the overlapping part contains the same content)

wangra-google: oh, i see, the latter block has smaller size, that means the overlapping part contains different content?

shanminchao: Ohhhh you might be right! When do memory buffers get captured? Is it when they're allocated, or is it captured all at once right before command submission?

I was assuming they were captured during allocation, which means memory might've changed. But it doesn't make sense that it's captured before allocation, when there's no chance for writing to happen. In that case, if they're all captured at once, then simply keeping the bigger one will work, as you suggested.

Does this align with your thoughts?

wangra-google: do you know what is the block type of those overlapping blocks? RD_GPUADDR or RD_CMDSTREAM_ADDR or RD_BUFFER_CONTENTS? i think they are allocated differently.

shanminchao: RD_BUFFER_CONTENTS is where all the memory blocks are added

wangra-google: looks like those are the command streams. did a quick search, seems to be captured at queue submit. still don't quite understand why there could be 2 blocks with the same address but with different sizes. (especially shrink in size later)

shanminchao: These are all the buffers, including command streams. But also includes stuff like shaders, textures, etc. I think they're captured during submission.

But yeah, I don't know why there are more than 1 blocks with same address, regardless of size.

shanminchao: Anyways, if they're always called during submit, then we don't have to worry about overwriting and keeping old memory and all that. Just keep the block with largest size

wangra-google: k, so 1 memcpy should be enough?

shanminchao: Pushed new change. Don't even need memcpy. Just use the bigger one and get rid of the smaller one.

wangra-google (approved):

PR #100, opened 2024-02-23
Follow cffdump and ignore missing memory for load_state constants

shanminchao:
I'm not sure why this is the case. TODO: Investigate in Turnip driver how this works.

But chromebook capture definitely has captures where this memory block was never captured. And cffdump ignores it,
so we should too.

### DIFF
--- a/dive_core/command_hierarchy.cpp
+++ b/dive_core/command_hierarchy.cpp
@@ -2547,7 +2547,15 @@ void CommandHierarchyCreator::AddConstantsToPacketNode(const IMemoryManager &mem
             {
                 T        value;
                 uint64_t addr = ext_src_addr + ((i + j) * sizeof(T));
-                DIVE_VERIFY(mem_manager.CopyMemory(&value, submit_index, addr, sizeof(T)));
+
+                // For some reason, some captures refer to memory not backed by memory blocks
+                // Let's not treat it as an error, since cffdump handles this gracefully as well
+                if (!mem_manager.CopyMemory(&value, submit_index, addr, sizeof(T)))
+                {
+                    DIVE_LOG("Indirect constant buffer at 0x%p with no backing memory!",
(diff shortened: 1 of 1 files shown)


dive_core/command_hierarchy.cpp
```diff
@@ -2547,7 +2547,11 @@ void CommandHierarchyCreator::AddConstantsToPacketNode(const IMemoryManager &mem
             {
                 T        value;
                 uint64_t addr = ext_src_addr + ((i + j) * sizeof(T));
-                DIVE_VERIFY(mem_manager.CopyMemory(&value, submit_index, addr, sizeof(T)));
+
+                // For some reason, some captures refer to memory not backed by memory blocks
+                // Let's not treat it as an error, since cffdump handles this gracefully as well
+                if (!mem_manager.CopyMemory(&value, submit_index, addr, sizeof(T)))
```
wangra-google, line 2553: should we use `DIVE_LOG` (or maybe other func) to print warning messages in this case?

wangra-google:
> I'm not sure why this is the case. TODO: Investigate in Turnip driver how this works.
> 
> But chromebook capture definitely has captures where this memory block was never captured. And cffdump ignores it, so we should too.

probably because of this: 
image

dive_core/command_hierarchy.cpp
shanminchao: Added log as suggested

shanminchao:
> > I'm not sure why this is the case. TODO: Investigate in Turnip driver how this works.
> > But chromebook capture definitely has captures where this memory block was never captured. And cffdump ignores it, so we should too.
> 
> probably because of this (i think the freedreno shim layer captures everything by default): image

Thanks for explanation!

wangra-google (approved):

PR #101, opened 2024-02-24
Replace std::pair by auto to avoid copy

wangra-google:
- this change speeds up loading time 75-80% for both debug and release build


wangra-google:
Very good explanation here:
https://stackoverflow.com/questions/38881707/functor-for-find-if-and-map-avoid-copy-of-value

shanminchao:
> Very good explanation here: https://stackoverflow.com/questions/38881707/functor-for-find-if-and-map-avoid-copy-of-value

Oh wow, that's interesting. And kinda stupid, haha

shanminchao (approved):

PR #102, opened 2024-02-26
Optimization: use separate keys for different gpu variants

wangra-google:
Optimization: use separate keys for different gpu variants and use std::unordered_map::find() instead of std::find_if 
Before optim: Debug ~16s Release ~0.8s
After optim: Debug ~13s Release ~0.2s

### DIFF
--- a/dive_core/capture_data.cpp
+++ b/dive_core/capture_data.cpp
@@ -270,13 +270,26 @@ void MemoryManager::Finalize(bool same_submit_copy_only, bool duplicate_ib_captu
             }
             else
             {
-                if (memory_block.m_va_addr != prev_addr || memory_block.m_data_size != prev_size)
+                // Check if it's a completely new block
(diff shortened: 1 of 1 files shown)


shanminchao:
You have me as an assignee? Not a reviewer?
Also, can you give me a short snippet of the before and after of the pm4_info.cpp code so I can get an idea of the difference?

wangra-google:
> You have me as an assignee? Not a reviewer? Also, can you give me a short snippet of the before and after of the pm4_info.cpp code so I can get an idea of the difference?

oops, fixed
The actual difference is only in `const RegInfo *GetRegInfo(uint32_t reg)`, the other changes are just removing '|0x0` (since we are only parsing a6xx.xml) there is no reg that supports multiple variants other than a6xx and a7xx

image
image

shanminchao (approved):

PR #109, opened 2024-04-23
[mesa] Adding mutex to make sure only 1 thread can dump bos

wangra-google:
Need this mutex to make sure only 1 thread can dump bos.
This is required since we assume RD_GPUADDR is always followed by RD_BUFFER_CONTENTS


shanminchao:
Quick question: How does cffdump deal with these race-conditioned captures? Does it fail too?

shanminchao:
> > Quick question: How does cffdump deal with these race-conditioned captures? Does it fail too?
> 
> it skips loading that buffer and later gives the error message of the "could not find: " if anything is referencing that buffer without crashing. I think it is a valid fix even for mesa, will try to submit this back to mesa later.

Ok sounds good! I just wanted to make sure that this is valid even for mesa

shanminchao (approved):

PR #112, opened 2024-04-23
Fix context reg bunch

shanminchao:
64-bit values are realized in this packet by 2 consecutive pairs of dwords, where the 1st pair has the lower 32-bits and the 2nd pair has the higher 32-bits

### DIFF
--- a/dive_core/command_hierarchy.cpp
+++ b/dive_core/command_hierarchy.cpp
@@ -1782,16 +1782,18 @@ void CommandHierarchyCreator::AppendRegNodes(const IMemoryManager &mem_manager,
 {
     // This version of AppendRegNodes takes in a raw buffer consisting of register offset + value
     // pairs
-    DIVE_ASSERT(dword_count % 2 == 0);
-    for (uint32_t i = 0; i < dword_count; i += 2)
+    uint32_t dword = 0;
+    while (dword < dword_count)
     {
-        struct
+        struct RegPair
         {
             uint32_t m_reg_offset;
             uint32_t m_reg_value;
-        } reg_pair;
-        uint64_t pair_addr = va_addr + i * sizeof(uint32_t);
+        };
+        RegPair  reg_pair;
+        uint64_t pair_addr = va_addr + dword * sizeof(uint32_t);
         DIVE_VERIFY(mem_manager.CopyMemory(&reg_pair, submit_index, pair_addr, sizeof(reg_pair)));
+        dword += 2;
 
         const RegInfo *reg_info_ptr = GetRegInfo(reg_pair.m_reg_offset);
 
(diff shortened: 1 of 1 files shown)


dive_core/capture_data.cpp
```diff
@@ -1143,7 +1143,8 @@ CaptureData::LoadResult CaptureData::LoadAdrenoRdFile(FileReader &capture_file)
         case RD_CHIP_ID:
         {
             // If it wasn't set already by a RD_GPU_ID
-            if (GetGPUID() == 0)
+            // Or if it was an invalid gpu_id, which leads to a kGPUVariantNone
```
wangra-google, line 1146: this is in another pr already?

shanminchao: Yes I realized as I created this PR that I accidentally included the commit from the other PR. By the time I managed to fix it (2 minutes later!), you already made this comment. You're too fast!

dive_core/command_hierarchy.cpp
```diff
@@ -1801,13 +1803,28 @@ void CommandHierarchyCreator::AppendRegNodes(const IMemoryManager &mem_manager,
         if (reg_info_ptr == nullptr)
             reg_info_ptr = &temp;
 
-        DIVE_ASSERT(!reg_info_ptr->m_is_64_bit);
+        uint64_t reg_value = reg_pair.m_reg_value;
```
wangra-google, line 1806: Sorry i am a bit confused here, could you add some comments about the layout for both 32bit and 64bit?

Is this understanding correct?
32bit: 
|--offset-|--value--|
|--32bit--|--32bit--|

reg_value = value

64bit:
|-offset0-|-value0--|-offset1-|-value1-|
|--32bit--|--32bit--|--32bit--|--32bit--|
if(offset1 == offset0 + 1)
{
    reg_value = (value1<<32) | value0
    dword +=4 
}
else 
{
    reg_value = value0
    dword +=2
}

dive_core/command_hierarchy.cpp
```diff
@@ -1801,13 +1803,28 @@ void CommandHierarchyCreator::AppendRegNodes(const IMemoryManager &mem_manager,
...
+            DIVE_VERIFY(mem_manager.CopyMemory(&new_reg_pair,
+                                               submit_index,
+                                               new_pair_addr,
+                                               sizeof(new_reg_pair)));
+
+            // Sometimes the upper 64-bits are not set
+            // Probably because they're 0s and there's no need to set it
+            if (new_reg_pair.m_reg_offset == reg_pair.m_reg_offset + 1)
+            {
+                dword += 2;
```
wangra-google, line 1820: shouldn't this `dword +=2` be outside of the `if` so that for 64bit register, even the upper 32bit is not used, we should still advance dword here?

dive_core/command_hierarchy.cpp
```diff
@@ -1801,13 +1803,28 @@ void CommandHierarchyCreator::AppendRegNodes(const IMemoryManager &mem_manager,
...
+        if (reg_info_ptr->m_is_64_bit)
+        {
+            RegPair  new_reg_pair;
+            uint64_t new_pair_addr = va_addr + dword * sizeof(uint32_t);
+            DIVE_VERIFY(mem_manager.CopyMemory(&new_reg_pair,
+                                               submit_index,
+                                               new_pair_addr,
+                                               sizeof(new_reg_pair)));
+
+            // Sometimes the upper 64-bits are not set
```
wangra-google, line 1816: should this comment be upper 32-bits of the 64bit register?

dive_core/command_hierarchy.cpp
shanminchao: Yes that's the right understanding

dive_core/command_hierarchy.cpp
shanminchao: Good catch. I meant to say "upper part of 64-bits", but changing it 32-bits makes more sense

dive_core/command_hierarchy.cpp
shanminchao: There's already a dword+=2 earlier. If the offsets are consecutive, then it does the dword+=2 here again, which ends up doing the equivalent of dword+=4 for this iteration of the loop.
However, if the offsets are not consecutive, then we only process the first 2 32-bit dwords during this iteration of the loop. Then we go back to the beginning of the loop and process the next 2 32-bit dwords as though it was a whole separate register (which it is).

wangra-google: oh, so even the reg_info_ptr->m_is_64_bit flag is true, if `offset1 != offset0 + 1`, the register will still only take 2 dwords instead of 4, right?

shanminchao: Yup

shanminchao: Note: cffdump also spits out a warning in this particular case:

`
      if (regbase != r->regbase + 1) {
         printf("WARNING: 64b discontinuity (%x, expected %x)\n", regbase, r->regbase + 1);
         r->has_dword_lo = false;
         return true;
      }
`

And we indeed get this warning in the cffdump of the capture I found the problem in

wangra-google (approved):

dive_core/command_hierarchy.cpp
wangra-google: ah, ok. make sense. thanks!

PR #113, opened 2024-04-24
Some minor fixes for gles application with OpenXR

wangra-google:
* Fix the case where 2 memory blocks have the same gpu address and size but different content
* Hack fix the openxr layer with a dummy xrInitializeLoaderKHR
* Disable the assert of submissions get from perfetto based on vkQueueSubmit (for gl applications)
* Increase the max pending ib to 150
* Fix a warning of uninitialized member var


shanminchao:
So this doesn't actually modify history to rename the commit. Rather it just creates a new one with the correct title:

image

I don't think this is necessary...

wangra-google:
> So this doesn't actually modify history to rename the commit. Rather it just creates a new one with the correct title:
> 
> image
> 
> I don't think this is necessary...

oh, ok, i will close this one then and be careful next time.

PR #118, opened 2024-05-30
LRZ Validator

wangra-google:
- command line tool to output a text file that shows LRZ status at drawcall level

### DIFF
--- a/lrz_validator/main.cpp
+++ b/lrz_validator/main.cpp
@@ -14,48 +14,36 @@
  limitations under the License.
 */
 #include <iostream>
+#include <optional>
 #include <vector>
 
 #include "dive_core/data_core.h"
 #include "pm4_info.h"
 
-int main(int argc, char **argv)
+bool ValidataLRZ(const Dive::CaptureMetadata &meta_data, const std::string &output_file_name)
 {
-    Pm4InfoInit();
-    if ((argc != 2) && (argc != 3))
-    {
-        std::cout
-        << "You need to call: lrz_validator <input_file_name.rd> <output_file_name.txt>(optional)";
-        return 0;
-    }
-    char *input_file_name = argv[1];
-
-    std::string output_file_name = "lrz_validation_result.txt";
-    if (argc == 3)
-    {
-        output_file_name = argv[2];
-    }
-
-    Dive::LogCompound               log_compound;
-    std::unique_ptr<Dive::DataCore> data_core = std::make_unique<Dive::DataCore>(&log_compound);
-    Dive::CaptureData::LoadResult   load_res = data_core->LoadCaptureData(input_file_name);
-    if (load_res != Dive::CaptureData::LoadResult::kSuccess)
-    {
-        std::cout << "Loading capture " << argv[1] << " failed!";
-        return 0;
-    }
-
-    if (!data_core->CreateMetaData())
-    {
-        std::cout << "Failed to create meta data!";
-        return 0;
-    }
-
-    std::cout << "Output validation result to " << output_file_name << std::endl;
-    std::ofstream                output_file(output_file_name);
-    const Dive::CaptureMetadata &meta_data = data_core->GetCaptureMetadata();
     size_t                       event_count = meta_data.m_event_info.size();
     const Dive::EventStateInfo  &event_state = meta_data.m_event_state;
+    bool                         lrz_test_passed = true;
+    std::optional<std::ofstream> output_file = std::nullopt;
+    if (!output_file_name.empty())
+    {
+        std::cout << "Output detailed validation result to \"" << output_file_name << "\""
+                  << std::endl;
+        output_file.emplace(std::ofstream(output_file_name));
+    }
+
+    std::function<void(const std::string &)> OutputDetails;
+
+    if (output_file.has_value())
+    {
+        OutputDetails = [&output_file](const std::string &str) { *output_file << str; };
+    }
+    else
+    {
+        OutputDetails = [](const std::string &) {};
+    }
+
     for (size_t i = 0; i < event_count; ++i)
     {
         const Dive::EventInfo &info = meta_data.m_event_info[i];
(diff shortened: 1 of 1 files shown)


dive_core/command_hierarchy.cpp
```diff
@@ -1622,35 +1568,27 @@ std::string CommandHierarchyCreator::GetEventString(const IMemoryManager &mem_ma
     {
         PM4_CP_DRAW_INDX packet;
         DIVE_VERIFY(mem_manager.CopyMemory(&packet, submit_index, va_addr, sizeof(packet)))
-        string_stream << "DrawIndexOffset(NumIndices:" << packet.bitfields2.NUM_INDICES << ","
-                      << "IndexBase:" << packet.bitfields3.INDX_BASE << ","
-                      << "IndexSize:" << packet.bitfields4.INDX_SIZE << ","
-                      << "VizQuery:" << packet.bitfields0.VIZ_QUERY << ")";
+        string_stream << "DrawIndexOffset(NumIndices:" << packet.bitfields2.NUM_INDICES << ")";
```
wangra-google, line 1571: I removed the primtype, index base, index size and viz query from both ui and the lrz cli, found them not quite useful, we can add them back if needed

dive_core/common/emulate_pm4.h
```diff
@@ -214,7 +214,7 @@ class EmulatePM4
         kTotalIbLevels
     };
 
-    static const uint32_t kMaxPendingIbs = 150;
+    static const uint32_t kMaxPendingIbs = 600;
```
wangra-google, line 217: firsthands need something like 500, put 600 to be safe.

dive_core/common/emulate_pm4.cpp
```diff
@@ -822,7 +822,7 @@ bool EmulatePM4::AdvanceCb(const IMemoryManager &mem_manager,
     // When it shows up in production driver captures, the IB addr and size values are all
     // garbage. Not sure what it's used for at this point.
     if (type7_header.opcode == CP_COND_INDIRECT_BUFFER_PFE)
-        DIVE_LOG("Packet ignored: CP_COND_INDIRECT_BUFFER_PFE\n");
+        DIVE_DEBUG_LOG("Packet ignored: CP_COND_INDIRECT_BUFFER_PFE\n");
```
wangra-google, line 825: use debug log so that the message is disabled in release build

dive_core/command_hierarchy.cpp
shanminchao: Sounds good! The user can get them from the PM4 packet anyways.

dive_core/common/emulate_pm4.cpp
shanminchao: Sounds good!

dive_core/common/emulate_pm4.h
shanminchao: I keep increasing this. Might have to make it dynamically allocated. I'm trying to avoid std::vector in core parsing code, but maybe we'll have to use something like that at some point. A problem to worry about later...

dive_core/data_core.h
```diff
@@ -173,4 +176,42 @@ class CaptureMetadataCreator : public IEmulateCallbacks
 #endif
 };
 
+//--------------------------------------------------------------------------------------------------
+// validate LRZ from the capture
+class LRZValidator : public IEmulateCallbacks
```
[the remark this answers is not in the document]
wangra-google: hmm, maybe i don't fully understand your idea. I put some extra logic in the OnPacket() function which are not part of the statetracker. (trying to separate the validation output code from the state tracker)

dive_core/common/emulate_pm4.h
wangra-google: yeah, agree.

dive_core/data_core.h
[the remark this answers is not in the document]
shanminchao: The idea is that you rely on the existing CaptureMetadataCreator. Have it fill out all the EventStateInfo, as it usually does for Dive. Then in your main.cpp you can simply loop through the EventStateInfo and do your depth_test and lrz state logic. 

The idea is that one-off CLI commands can rely all the heavy lifting code-wise that we already wrote. They can just look at the metadata we determined rather than re-write callbacks and handle parsing of PM4 packets and duplicate some logic. If we're missing anything in CaptureMetadataCreator that blocks your attempt to do this, then maybe that just means we're missing stuff in CaptureMetadataCreator?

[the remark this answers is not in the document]
shanminchao: In other words, I would've imgained your lrz_validator to just use dive_core as-is without modifications, rather than require actually making changes in dive_core. Any changes that we add to dive_core should be minimal and indicative of information we didn't parse or cache properly to begin with. In principle anyways.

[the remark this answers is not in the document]
wangra-google: I see your point. My worries is that CaptureMetadataCreator is doing lots of other things that are not requried by the lrz_validator and will slowdown the lrz_validator (it is currently already kinda slow, for the firsthand scene with ~100 drawcalls, it is taking ~30s) The `EventStateInfo` contains the states for all drawcalls (binning + multiple tiles) whereas the lrz_validator only wants to show the drawcalls in the binning pass (we could add an extra field in EventStateInfo to indicate if the drawcall is in the binning pass, a bit weird if it is only used for lrz_validator)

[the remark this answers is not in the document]
shanminchao: I would be surprised if there's a real slowdown in CaptureMetadataCreator, because it's something we run everytime we load a capture. So far that never takes ~30s, and if it does, then it means we need to optimize it. I'd be curious what's taking so long for lrz_validator.

Yes, EventStateInfo will always contain more info than you could possibly need, but that's why it's so broadly useful. We can create lots of one-off CLI tools without worrying about parsing, because CaptureMetadataCreator should meet most needs. I can imagine other clients of dive_core (including Dive itself!) might want to differentiate binning vs tile passes, so it seems reasonable to add that info to CaptureMetadataCreator.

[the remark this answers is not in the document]
wangra-google: currently, ~90% of the time is spent in the loading of the texture, and the loading lrz function is taking ~7.5%  which is ~2.3s. (half of it are spent in Dive::EmulatePM4::AdvanceCb(Dive::MemoryManager::GetMaxContiguousSize) and the other half are spent in Dive::MemoryManager::CopyMemory)

Another thing i am worried is that one of the optimizations that i am thinking of is to merge the `create trees` with the `metatdata creator` so that we don't need to loop through the entire packets 2 times (we could avoid those heavy memory operations), but that also means we pay the cost of creating tress for lrz cli as well. 

in the latest commit, to avoid polluting dive_core, i moved lrz to main.cpp, so we do all the validation logic in the cli project. What do you think?

lrz_validator/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "dive_core/data_core.h"
+#include "pm4_info.h"
+
+int main(int argc, char **argv)
+{
+    Pm4InfoInit();
+    if ((argc != 2) && (argc != 3))
+    {
+        std::cout
+        << "You need to call: lrz_validator <input_file_name.rd> <output_file_name.txt>(optional)";
```
shanminchao, line 28: Suggestion: We print to stdout whether LRZ test passed or not. And if the user wants to find out why it doesn't pass, then they can pass in an output_file_name.txt. In that case, maybe make the output_file truly optional, in that we don't output any files unless the user specifies it. Give them a bit more control, since in some cases their automated tests might just want to know whether the LRZ stuff passes or not, and don't care about the details?

shanminchao: Their automated tests will probably parse stdout to see if the test passed or failed. Writing that into a verbose text file will force them to have to parse our complicated detailed output file, which could be a pain. That's why I suggest printout out a PASS/FAIL type of string in stdout only, with no details. So it's a predictable printout they can parse.

wangra-google: Thanks for the suggestion, make sense. I made the change as suggested in the latest commit.

dive_core/data_core.h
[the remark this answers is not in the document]
wangra-google: Updated the code as we discussed. The 30+sec is an old capture with a size of 1GB (~18GB after decompression), this is without robin's fix. With [robin's fix](https://github.com/google/dive/pull/87), the size of the capture from the same scene is about ~160MB (compressed), and loading takes about 6s.

shanminchao (approved):

PR #123, opened 2024-06-07
Events/IBs now show all submit PM4 packets, but with highlighting and auto-scroll

shanminchao:
Right now Dive has an issue with displaying draw calls that appear in different IB levels.
For example, Draw 1 can appear in IB2 and Draw 2 in IB1 - but the per-event command buffer view can't just start displaying deep in the command buffer tree. Dive tries to treat everything as IB1, and will have trouble visualizing Draw 1 correctly.

To fix this, when clicking on an event, now the Command Buffer view shows *all* PM4 packets for the submit. When clicking on the event, it will highlight the PM4 packets belonging to the selected draw call, and also auto-scroll to the end of the event PM4 packets:

image

This highlight and auto-scroll also works for IBs in the Submit view. 
Missing is the display of PM4 packets for the pass nodes (e.g. Resolve Pass). This will come later, after a critical bug for the passes are fixed first.

### DIFF
--- a/ui/command_buffer_model.cpp
+++ b/ui/command_buffer_model.cpp
@@ -55,13 +55,11 @@ void CommandBufferModel::Reset()
 }
 
 //--------------------------------------------------------------------------------------------------
-void CommandBufferModel::SetTopologyToView(const Dive::Topology *topology_ptr,
-                                           bool                  show_level_column)
+void CommandBufferModel::SetTopologyToView(const Dive::Topology *topology_ptr)
 {
     emit beginResetModel();
     m_topology_ptr = topology_ptr;
     m_selected_node_index = UINT64_MAX;
-    m_show_level_column = show_level_column;
     emit endResetModel();
 }
 
@@ -69,6 +67,12 @@ void CommandBufferModel::SetTopologyToView(const Dive::Topology *topology_ptr,
 int CommandBufferModel::columnCount(const QModelIndex &parent) const
 {
     return kColumnCount;
+}
+
+//--------------------------------------------------------------------------------------------------
+QModelIndex CommandBufferModel::scrollToIndex() const
+{
+    return m_scroll_to_index;
 }
 
 //--------------------------------------------------------------------------------------------------
@@ -85,10 +89,13 @@ QVariant CommandBufferModel::data(const QModelIndex &index, int role) const
             return int(Qt::AlignLeft | Qt::AlignVCenter);
     }
 
+    uint64_t node_index = (uint64_t)(index.internalPointer());
+    if (role == Qt::ForegroundRole && IsCurEvent(node_index))
+        return QColor(255, 128, 128);
+
     if (role != Qt::DisplayRole)
         return QVariant();
 
-    uint64_t       node_index = (uint64_t)(index.internalPointer());
     Dive::NodeType node_type = m_command_hierarchy.GetNodeType(node_index);
 
     // Column 2: Address (will be swapped via moveSection() to be visually the 0th column)
@@ -109,7 +116,7 @@ QVariant CommandBufferModel::data(const QModelIndex &index, int role) const
     }
     else if (index.column() == CommandBufferModel::kColumnIbLevel)
     {
-        if (node_type == Dive::NodeType::kPacketNode && m_show_level_column)
+        if (node_type == Dive::NodeType::kPacketNode)
         {
             uint64_t           ib_level = m_command_hierarchy.GetPacketNodeIbLevel(node_index);
             std::ostringstream ib_level_string_stream;
@@ -123,7 +130,15 @@ QVariant CommandBufferModel::data(const QModelIndex &index, int role) const
     }
     else  // if (index.column() == CommandBufferModel::kColumnPm4)
     {
+#ifndef NDEBUG  // DEBUG
+        // Print out node_index, which helps a lot in debugging command hierarchy issues
+        std::ostringstream addr_string_stream;
+        addr_string_stream << m_command_hierarchy.GetNodeDesc(node_index) << " (" << node_index
+                           << ")";
+        return QString::fromStdString(addr_string_stream.str());
+#else
         return QString(m_command_hierarchy.GetNodeDesc(node_index));
+#endif
     }
 }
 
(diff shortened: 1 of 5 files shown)


dive_core/command_hierarchy.cpp
```diff
@@ -75,6 +75,9 @@ uint64_t Topology::GetNumNodes() const
     DIVE_ASSERT(m_node_children.size() == m_node_shared_children.size());
     DIVE_ASSERT(m_node_children.size() == m_node_parent.size());
     DIVE_ASSERT(m_node_children.size() == m_node_child_index.size());
+    DIVE_ASSERT(m_node_children.size() == m_start_shared_child.size());
```
wangra-google, line 78: why is m_node_children.size() == m_start_shared_child.size() == m_end_shared_child.size() == m_root_node_index.size()?

dive_core/command_hierarchy.cpp
```diff
@@ -105,6 +108,25 @@ uint64_t Topology::GetChildNodeIndex(uint64_t node_index, uint64_t child_index)
...
+{
+    uint64_t num_children = GetNumChildren(node_index);
+    if (num_children > 0)
+        return GetChildNodeIndex(node_index, 0);
+    while (true)
+    {
+        if (node_index == kRootNodeIndex)
+            return UINT64_MAX;
+        uint64_t parent_node_index = GetParentNodeIndex(node_index);
+        uint64_t sibling_index = GetChildIndex(node_index) + 1;
```
wangra-google, line 123: a bit confused with the logic here, if the code reaches here, that means GetNumChildren(node_index) <= 0, so the current node (with node_index) should not have any child?

dive_core/command_hierarchy.h
```diff
@@ -83,9 +84,22 @@ class Topology
...
+    uint64_t GetNextNodeIndex(uint64_t node_index) const;
+
+    // Used to get list of shared children for a top level root node
+    // All children of this top level node share this alternate set of children
+    // Each non-top-level node has begin/end indices towards this set of shared children
     uint64_t GetNumSharedChildren(uint64_t node_index) const;
     uint64_t GetSharedChildNodeIndex(uint64_t node_index, uint64_t child_index) const;
-    uint64_t GetNextNodeIndex(uint64_t node_index) const;
+
+    // Indicate the being & end index of the shared children in the top-level node that belong to
```
wangra-google, line 95: just to confirm, the shared children are the pm4 packets in the command view tab, and the nodes are the drawcalls/dispatch/blit in the tree view, right? and the Start and End here indicates the start and end index of the pm4 packets of a selected drawcall/dispatch/blit, correct?

dive_core/command_hierarchy.cpp
wangra-google: ah, i see, it is because each non-root node will have 1 start, 1 end, and 1 root, right?

dive_core/command_hierarchy.cpp
```diff
@@ -2275,35 +2241,30 @@ void CommandHierarchyCreator::AddSharedChild(CommandHierarchy::TopologyType type
...
+{
+    DIVE_ASSERT(node_index < m_node_end_shared_child[type].size());
+    m_node_end_shared_child[type][node_index] = shared_child_node_index;
+}
 
-    vec.erase(vec.begin() + vec.size() - children_node_indices.size(), vec.end());
+//--------------------------------------------------------------------------------------------------
+void CommandHierarchyCreator::SetSharedChildRootNodeIndex(CommandHierarchy::TopologyType type,
+                                                          uint64_t                       node_index,
+                                                          uint64_t root_node_index)
```
wangra-google, line 2264: is the root node always the submit node? could it be the marker node as well?

dive_core/command_hierarchy.cpp
```diff
@@ -798,6 +824,18 @@ bool CommandHierarchyCreator::OnIbEnd(uint32_t                  submit_index,
     m_state_tracker.PopEnableMask();
     DIVE_ASSERT(!m_ib_stack.empty());
 
+    // Setup root & range of shared children that this IB encompasses
```
wangra-google, line 827: this is basically to mark the range of pm4s of this IB, right?

dive_core/command_hierarchy.cpp
wangra-google: not necessary in this change, but almost feel that it would be cleaner to have AoS instead of a SoA here.

dive_core/command_hierarchy.h
shanminchao: Pretty much. Shared nodes in principle are any nodes that can have multiple parents. In this case, PM4 packets can have as parents IB nodes, event nodes, etc. In our command hierarchy, *only* PM4 packets (not their fields) are shared nodes.

The non-shared nodes are everything else, including the children of shared nodes (e.g. the "fields" for each of the PM4 packets)

> and the Start and End here indicates the start and end index of the pm4 packets of a selected drawcall/dispatch/blit, correct?

Yes, the start and end index of the shared children that are part of selected drawcalls/dispatches/blits/ibs.

dive_core/command_hierarchy.cpp
shanminchao: Every child has its own set of children, start/end child, parent, etc. So they all must be same size.
Yes, in some ways AoS is cleaner than SoA, especially when we have so many parallel arrays. But as with all things AoS, it's easy to get unwieldy and bolt on abstractions when you do AoS, and more importantly (as you know), it's likely less cache friendly.

dive_core/command_hierarchy.cpp
shanminchao: I didn't modify this function at all for this PR, and in fact this GetNextNodeIndex() function isn't being used at all. The diff is only making this part show up because it's being weird.

I think the logic here is that if a node doesn't have children, then it's trying to search for "next node index" in sibling branches. We should probably remove this in a future PR since this isn't being used anymore.

dive_core/command_hierarchy.cpp
shanminchao: Yup

dive_core/command_hierarchy.cpp
shanminchao: It's explicitly set as the submit node in all calls to this function. If there's a topology where the submit doesn't show up, then it can be set to something else as well. But at the moment that's the only use case

wangra-google (approved):

PR #125, opened 2024-07-22
CP_START_BIN's prefix/common all come as separate ibs now, all before CP_END_BIN

shanminchao:
This is the first step towards appropriate UI visualization of prefix and common IBs in the command hierarchy

### DIFF
--- a/dive_core/common/emulate_pm4.h
+++ b/dive_core/common/emulate_pm4.h
@@ -50,8 +50,13 @@ struct IndirectBufferInfo
 };
 
 // clang-format off
-union Pm4Type4Header
+union Pm4Header
 {
+    struct
+    {
+        uint32_t offset_parity : 28;
+        uint32_t type : 4;
+    };
     struct
     {
         uint32_t count          : 7;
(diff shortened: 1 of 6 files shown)


dive_core/common/emulate_pm4.h
```diff
@@ -50,34 +50,34 @@ struct IndirectBufferInfo
 };
 
 // clang-format off
-union Pm4Type4Header
+union Pm4Header
 {
+    struct
+    {
+        uint32_t offset_parity : 28;
+        uint32_t type : 4;
```
wangra-google, line 58: i think it is still better with enum?

shanminchao: Well, it's a bitfield, so I kept it as that.
I realized the difference between the following was minimal anyways:

if (header.type == Dive::kPm4Type4)

vs

if (header.type == 4)

It seems kinda redundant to add a Dive::kPm4Type*.

Another point is that this bitfield is a mapping directly to fields obtained from a PM4 packet. So having an enum here doesn't quite work unless we create an enum to map even the other non-existing types (type1, type2, type3, type4, type5, etc)

wangra-google (approved):

PR #127, opened 2024-08-08
Enable run application with additional args

RenfengLiu:
This allows to capture bench mark apps that has additional arguments.

### DIFF
--- a/ui/trace_window.cpp
+++ b/ui/trace_window.cpp
@@ -107,6 +107,12 @@ TraceDialog::TraceDialog(QWidget *parent)
     m_cmd_layout->addWidget(m_file_label);
     m_cmd_layout->addWidget(m_cmd_input_box);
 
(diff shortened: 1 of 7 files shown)


ui/trace_window.cpp
```diff
@@ -283,12 +298,7 @@ bool TraceDialog::StartPackage(Dive::AndroidDevice *device, const std::string &a
             ShowErrorMessage(err_msg);
             return false;
         }
-        std::vector<std::string> v = absl::StrSplit(full_command, " ");
-        m_executable = v[0];
-
-        v.erase(v.begin());
-        m_command_args = absl::StrJoin(v, " ");
-
+        m_executable = full_command;
```
wangra-google, line 301: so now the full_command doesn't contain any args?

wangra-google (approved):

PR #135, opened 2024-09-18
Adds App filter

GrantComm:
Adds app filter that allows users to view all apps, debuggable apps, or non-debuggable apps when starting a capture.

### DIFF
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -19,6 +19,7 @@ limitations under the License.
 #include <filesystem>
 #include <memory>
 
+#include "../dive_core/common/common.h"
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
 #include "absl/strings/str_cat.h"
@@ -131,7 +132,7 @@ absl::StatusOr<std::vector<std::string>> AndroidDevice::ListPackage(PackageListO
                     package_list.push_back(package);
                 }
                 break;
-            default: assert(false); break;  // Unknown Package List Option.
+            default: DIVE_ASSERT(false); break;  // Unknown Package List Option.
             }
         }
     }


capture_service/device_mgr.cc
```diff
@@ -113,12 +119,26 @@ absl::StatusOr<std::vector<std::string>> AndroidDevice::ListPackage(PackageListO
...
+                if (!result.ok())
+                {
+                    return result.status();
+                }
+                output = *result;
                 if (!absl::StrContains(output, "DEBUGGABLE"))
                 {
-                    continue;
+                    package_list.push_back(package);
+                    ;
```
wangra-google, line 139: this line can be removed

capture_service/device_mgr.h
```diff
@@ -70,12 +70,14 @@ class AndroidDevice
 
     struct PackageListOptions
```
wangra-google, line 71: would it be better to use enum here? all 3 options seems to be mutually exclusive?

wangra-google (approved):

capture_service/device_mgr.cc
```diff
@@ -96,29 +96,42 @@ absl::StatusOr<std::vector<std::string>> AndroidDevice::ListPackage(PackageListO
...
+                {
+                    return result.status();
+                }
+                output = *result;
                 if (!absl::StrContains(output, "DEBUGGABLE"))
                 {
-                    continue;
+                    package_list.push_back(package);
                 }
+                break;
```
wangra-google, line 133: also better to add a `default` case and assert in the default case so that in the future if people add a new enum and forget to handle the new one here, the assert would be triggered

capture_service/device_mgr.cc
```diff
@@ -96,29 +96,43 @@ absl::StatusOr<std::vector<std::string>> AndroidDevice::ListPackage(PackageListO
...
+            default: assert("Error, unknown package list option.");
```
wangra-google, line 134: with `break` as well :)

GrantComm: Added the break and made changed the assert so that it would actually be triggered in the default case.

capture_service/device_mgr.cc
```diff
@@ -96,29 +96,43 @@ absl::StatusOr<std::vector<std::string>> AndroidDevice::ListPackage(PackageListO
...
+                    return result.status();
+                }
+                output = *result;
                 if (!absl::StrContains(output, "DEBUGGABLE"))
                 {
-                    continue;
+                    package_list.push_back(package);
                 }
+                break;
+            default: assert(false); break;  // Unknown Package List Option.
```
wangra-google, line 134: better with `DIVE_ASSERT`

wangra-google (approved):

PR #136, opened 2024-09-26
Enables shader disassembly on Windows/MSVC

GrantComm:
Shader disassembly relies on mesa generated code (ir3-isa.c) that MSVC refused to compile because it had zero size arrays.  This change checks ensures there are no zero sized arrays and adds compile checks for msvc

### DIFF
--- a/third_party/mesa/src/util/google_dive_ext.h
+++ b/third_party/mesa/src/util/google_dive_ext.h
@@ -14,6 +14,8 @@
  limitations under the License.
 */
 
+#ifdef _MSC_VER
+
 #ifndef _GOOGLE_DIVE_EXT_H
 #define _GOOGLE_DIVE_EXT_H
 
@@ -42,3 +44,5 @@ int futex_wake(uint32_t *addr, int count);
 #endif
 
 #endif /* _GOOGLE_DIVE_EXT_H*/
+
+#endif /* _MSC_VER*/
--- a/third_party/mesa/src/util/google_dive_ext.c
+++ b/third_party/mesa/src/util/google_dive_ext.c
@@ -14,6 +14,7 @@
  limitations under the License.
 */
 
+#ifdef _MSC_VER
 #include "google_dive_ext.h"
 
 char *strndup(const char *s, size_t n)
@@ -60,3 +61,5 @@ int futex_wake(uint32_t *addr, int count)
 {
     return SetEvent(addr);
 }
+
+#endif /* _MSC_VER*/


wangra-google:
a general comment here: we need to add `// GOOGLE: <reason for the change>` when we make any change on any file in the mesa folder. The reason is that we might need to integrate the new versions from mesa in the future, it would be very difficult for integration if we do not put comments on our local changes.

third_party/mesa/src/compiler/isaspec/isaspec_decode_impl.c
```diff
@@ -733,7 +776,13 @@ display(struct decode_scope *scope)
 				e++;
 			}
 
-			char *field_name = strndup(p, e-p);
+#ifdef _MSC_VER
+            char *field_name = (char *)malloc(e - p + 1); 
+            strncpy(field_name, p, e - p);
+            field_name[e - p] = '\0'; 
+#else
+            char *field_name = strndup(p, e - p);
```
wangra-google, line 784: out of curiosity, `strndup` is not working on windows?

wangra-google: would it be cleaner if we encode our own implementation into a header/cpp and wrap the header with #ifdef _MSC_VER

something like:
```
google_dive_ext.h
namespace Dive
{
...
char *strndup(const char *s, size_t n)
...
}

google_dive_ext.cpp
char *strndup(const char *s, size_t n) {
    size_t len = strnlen(s, n); 
    char *new_str = (char *)malloc(len + 1);
    if (new_str == NULL) {
        return NULL;
    }
    memcpy(new_str, s, len);
    new_str[len] = '\0'; 
    return new_str;
}
```
in this case, we could isolate our changes from the other part of the mesa code, would be much easier for future integration

third_party/mesa/src/compiler/isaspec/isaspec_decode_impl.c
```diff
@@ -184,7 +184,30 @@ isa_print(struct isa_print_state *state, const char *fmt, ...)
 	int ret;
 
 	va_start(args, fmt);
-	ret = vasprintf(&buffer, fmt, args);
+// GOOGLE: vasprintf is not directly available in Windows.
```
wangra-google, line 187: similar to `strndup`, `vasprintf` could probably moved to the dive specific .h/.c file

third_party/mesa/src/util/google_dive_ext.h
```diff
@@ -0,0 +1,34 @@
...
+#define _GOOGLE_DIVE_EXT_H
+
+#include <stdlib.h>
+#include <stdarg.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+char *strndup(const char *s, size_t n);
```
wangra-google, line 27: can you add some comments on why we need to reimplement those functions? thanks :)

third_party/mesa/src/util/simple_mtx.h
```diff
@@ -28,6 +28,10 @@
 #include "util/macros.h"
 #include "util/u_call_once.h"
 #include "u_atomic.h"
+// GOOGLE: Include windows header when the OS is Windows.
+#ifdef _WIN32
```
wangra-google, line 32: out of curiosity, is there a reason you use _WIN32 here instead of _MSC_VER?

third_party/mesa/src/util/simple_mtx.h
```diff
@@ -110,10 +114,18 @@ simple_mtx_lock(simple_mtx_t *mtx)
...
       if (c != 2)
          c = p_atomic_xchg(&mtx->val, 2);
-      while (c != 0) {
-         futex_wait(&mtx->val, 2, NULL);
-         c = p_atomic_xchg(&mtx->val, 2);
-      }
+// GOOGLE: Windows does not use futexes.
+#ifdef _WIN32
+        while (c != 0) {
+            WaitForSingleObject(mtx->val, 2);
```
wangra-google, line 120: can we define `futex_wait` in google_dive_ext.h/c as well? (with the actual implementation with `WaitForSingleObject`)

third_party/mesa/src/util/simple_mtx.h
```diff
@@ -132,7 +144,12 @@ simple_mtx_unlock(simple_mtx_t *mtx)
 
    if (__builtin_expect(c != 1, 0)) {
       mtx->val = 0;
-      futex_wake(&mtx->val, 1);
+// GOOGLE: Windows uses events instead of futexes.
+#ifdef _WIN32
+        SetEvent(mtx->val);
+#else
+        futex_wake(&mtx->val, 1);
```
wangra-google, line 151: same as above

third_party/mesa/src/util/simple_mtx.h
GrantComm: It should've been _MSC_VER to be more exact but works in a similar way. Removed since it is no longer needed.

third_party/mesa/src/freedreno/ir3/instr-a3xx.h
```diff
@@ -24,16 +24,33 @@
 #ifndef INSTR_A3XX_H_
 #define INSTR_A3XX_H_
 
+// GOOGLE: A pragma pack directive is needed to pack a structure in Windows.
+#if defined(__GNUC__)
 #define PACKED __attribute__((__packed__))
+#elif defined(_MSC_VER)
+#define PACKED
```
wangra-google, line 31: hmm, i may not get this. 
this means we define `PACKED` as empty, and then add 2 `pragma`
#pragma pack(push, 1)
#pragma pack(pop)

But if we want the pragma to affect a struct, we need to wrap the struct declaration with the pragma. Do i miss anything here?

(whereas with `__attribute__((__packed__))`, we make sure the struct is packed.)

third_party/mesa/src/compiler/isaspec/isaspec_decode_impl.c
```diff
@@ -912,8 +916,7 @@ decode_bitset(void *out, struct decode_scope *scope)
 			while (*e != '}') {
 				e++;
 			}
-
-			char *field_name = strndup(p, e-p);
+			char *field_name = strndup(p, e - p);
```
wangra-google, line 919: we can probably keep the original one here

third_party/mesa/src/freedreno/ir3/instr-a3xx.h
GrantComm: Correct, this is in the wrong spot since the nothing is packed here. I moved the PACKED msvc definition to the other instr-a3xx file and moved the pragma pack(pop) to be after the last packed struct

third_party/freedreno/includes/instr-a3xx.h
```diff
@@ -748,6 +757,10 @@ typedef union PACKED {
 	};
 } instr_t;
 
+#if defined(_MSC_VER)
```
wangra-google, line 760: could you also add `// GOOGLE: <reason for modification>` here? so that it would be easier for future integration. thanks :)

wangra-google (approved):

wangra-google:
strange that i keep having following building errors after applying this PR:
image

dive_core/shader_disassembly.cpp
```diff
@@ -19,12 +19,14 @@
 #include "dive_core/common/common.h"
 #include "pm4_info.h"
 
-#ifndef _MSC_VER
+#ifdef _MSC_VER
```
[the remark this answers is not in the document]
GrantComm: I was following the compile guards which previously specified _MSC_VER since that is what's being used to compile everything versus _WIN32 specifying the platform.

third_party/mesa/src/util/google_dive_ext.h
```diff
@@ -0,0 +1,44 @@
...
+#include <stdlib.h>
+#include <stdarg.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/**
```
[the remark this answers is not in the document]
GrantComm: I added the _MSC_VER compile guards since msvc is what's being used to compile

wangra-google (approved):

wangra-google:
> strange that i keep having following building errors after applying this PR (having this error in both Debug and Release): image

i tried another pc, no such error. must be something strange locally on that pc.

third_party/mesa/src/util/google_dive_ext.h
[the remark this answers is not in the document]
GrantComm: Oh that's weird. It is in the latest commit: https://github.com/google/dive/pull/136/commits/10f32606f0f030f92566d2a9dbeff9c28521352d

PR #137, opened 2024-10-09
Shows appropriate text for device selection box

GrantComm:
Adds items to device selection dropdown to signify if there are devices available.

Screenshot 2024-10-08 at 6 20 49 PM
Screenshot 2024-10-09 at 9 43 05 AM
Screenshot 2024-10-09 at 9 43 15 AM

### DIFF
--- a/ui/trace_window.cpp
+++ b/ui/trace_window.cpp
@@ -88,28 +88,7 @@ TraceDialog::TraceDialog(QWidget *parent)
     m_main_layout = new QVBoxLayout();
 
     m_devices = Dive::GetDeviceManager().ListDevice();
-    if (m_devices.size() == 0)
-    {
-        QStandardItem *item = new QStandardItem("No devices found");
-        item->setFlags(item->flags() & ~Qt::ItemIsSelectable);
-        m_dev_model->appendRow(item);
-        m_dev_box->setCurrentIndex(0);
-    }
-    else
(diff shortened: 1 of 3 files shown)


ui/trace_window.cpp
```diff
@@ -88,18 +88,35 @@ TraceDialog::TraceDialog(QWidget *parent)
     m_main_layout = new QVBoxLayout();
 
     m_devices = Dive::GetDeviceManager().ListDevice();
-    for (size_t i = 0; i < m_devices.size(); i++)
+    if (m_devices.size() == 0)
```
wangra-google, line 91: you can use `empty()`

ui/trace_window.cpp
```diff
@@ -228,26 +245,44 @@ void TraceDialog::UpdateDeviceList()
     m_devices = cur_list;
     m_dev_model->clear();
 
-    for (size_t i = 0; i < m_devices.size(); i++)
+    if (m_devices.size() == 0)
```
wangra-google, line 248: same as above

wangra-google: actually this code block looks like a copy of the above, can we put it in a function instead of duplicating the code. (a bit error prone this way, people need to modify code in both places if any modification is required. )

wangra-google (approved):

PR #139, opened 2024-11-07
Optimize draw selection by getting rid of the auto-expand

shanminchao:


ui/command_tab_view.cpp
```diff
@@ -98,9 +98,11 @@ void CommandTabView::OnSelectionChanged(const QModelIndex &index)
...
     m_command_buffer_model->OnSelectionChanged(index);
 
-    // After m_command_buffer_view is filled out in CommandBufferModel::OnSelectionChanged(), expand
-    // the tree
-    m_command_buffer_view->expandAll();
+    // After m_command_buffer_view is filled out in CommandBufferModel::OnSelectionChanged(), do NOT
+    // expandAll. For huge trees (e.g. 20+ million nodes), it needs to traverse all expanded nodes
+    // to figure out how to setup scrollbar. So the more collapsed it is, the less it has to
+    // traverse
+    // m_command_buffer_view->expandAll();
```
wangra-google, line 105: shall we remove this block completely?

ui/command_buffer_model.cpp
```diff
@@ -419,7 +420,7 @@ bool CommandBufferModel::CreateNodeToParentMap(uint64_t parent_row,
         // one of the children. Will no longer be part of current event/pass in the future If
         // is_selected==false, and return value is true, then that means event/pass started in one
         // of the children. Will continue be part of the current event/pass going forward
-        is_selected = CreateNodeToParentMap(child, child_node_index, is_selected);
+        is_selected = CreateNodeToParentMap(num_children + child, child_node_index, is_selected);
```
wangra-google, line 423: Does it mean that we will have more nodes now? :p

wangra-google (approved):

ui/command_tab_view.cpp
shanminchao: I kept it there as a warning in case we're ever tempted again to expandAll for UX reasons.
But if you prefer removing that, I can

ui/command_buffer_model.cpp
shanminchao: This is a bug fix that took me forever to track down.
Basically the order of entries in each level of the tree is the normal children followed by the shared children. 

For example:
image

Here we see ADDR_LO, ADDR_HI, and SIZE as field nodes, before a whole bunch of shared nodes appear. So the first shared node, CP_SET_DRAW_STATE, is actually in the 4th row. Without the "num_children+" part, CP_SET_DRAW_STATE is set to the in the 0th row, which causes all sorts of seemingly random issues with QT.

wangra-google: good to know, thanks!

PR #142, opened 2024-11-12
bump this the error limit from 5 to 50 to show the full shaders

wangra-google:
This is to increase the limit of the max_error count so that the full shader disassembly can be printed even with errors:
https://github.com/google/dive/blob/main/third_party/mesa/src/compiler/isaspec/isaspec_decode_impl.c#L767

(Some of the ISA code may not be interpreted correctly since our version of mesa is out dated. I will update the mesa repo when i have time.)

### DIFF
--- a/layer/openxr_layer.cc
+++ b/layer/openxr_layer.cc
@@ -50,18 +50,18 @@ struct XrInstanceData
 {
     XrInstance               instance;
     XrGeneratedDispatchTable dispatch_table;
-    ServerRunner            &server;
-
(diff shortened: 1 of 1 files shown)


shanminchao:
Did you mean to assign us as reviewers?

wangra-google:
haha, yeah, sorry, updated you as reviewers

wangra-google:
> Just curious, if error happens, would we expect the shader correctly disassembled?

it will show something like this (only for problematic instructions):
image

PR #146, opened 2024-12-02
Adds Gfxr capture through dive cli

GrantComm:
Adds the ability to perform a gfxr capture from the dive cli.

To perform a gfxr capture the gfxr_capture command must be used. Values for the device architecture, frame or frame range, and name of the directory containing the file that is saved on the device are set with flags. The following is an example:

./dive_client_cli --device localhost:40695 --command gfxr_capture --package com.google.android.filament.hellotriangle --type vulkan --device_architecture arm64-v8a --gfxr_capture_file_dir gfxr_triangle_captures --frame_range 1-3

### DIFF
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -422,15 +422,10 @@ absl::Status DeviceManager::Cleanup(const std::string &serial, const std::string
 }
 
 absl::Status AndroidDevice::RetrieveTrace(const std::string &trace_path,
-                                          const std::string &save_path,
-                                          const bool         isGfxrCapture)
+                                          const std::string &save_path)
 {
     RETURN_IF_ERROR(Adb().Run(absl::StrFormat("pull %s %s", trace_path, save_path)));
-    if (isGfxrCapture)
-    {
-        return Adb().Run(absl::StrFormat("shell rm -rf %s", trace_path));
-    }
-    return Adb().Run(absl::StrFormat("shell rm %s", trace_path));
+    return Adb().Run(absl::StrFormat("shell rm -rf %s", trace_path));
 }
 
 void AndroidDevice::EnableGfxr(bool enable_gfxr)
--- a/capture_service/device_mgr.h
+++ b/capture_service/device_mgr.h
@@ -96,9 +96,7 @@ public:
     AdbSession       &Adb() { return m_adb; }
 
     AndroidApplication *GetCurrentApplication() { return m_app.get(); }
-    absl::Status        RetrieveTrace(const std::string &trace_file_path,
-                                      const std::string &save_path,
-                                      const bool         is_gfxr_capture);
+    absl::Status RetrieveTrace(const std::string &trace_file_path, const std::string &save_path);
 
 private:
     const std::string                   m_serial;
--- a/capture_service/dive_client_cli.cc
+++ b/capture_service/dive_client_cli.cc
@@ -314,8 +314,7 @@ bool trigger_capture(Dive::DeviceManager& mgr)
     }
     target_download_path /= p.filename();
     auto ret = mgr.GetDevice()->RetrieveTrace(*trace_file_path,
-                                              target_download_path.generic_string(),
-                                              false);
+                                              target_download_path.generic_string());
     if (ret.ok())
         std::cout << "Capture saved at " << target_download_path << std::endl;
     else
@@ -339,8 +338,7 @@ bool retrieve_gfxr_capture(Dive::DeviceManager& mgr, const std::string& gfxr_cap
     }
     std::string capture_directory = Dive::kGfxrCaptureDirectory + gfxr_capture_directory;
     auto        ret = mgr.GetDevice()->RetrieveTrace(capture_directory,
-                                              target_download_path.generic_string(),
-                                              true);
+                                              target_download_path.generic_string());
     if (ret.ok())
         std::cout << "GFXR capture directory saved at " << target_download_path << std::endl;
     else
--- a/ui/trace_window.cpp
+++ b/ui/trace_window.cpp
@@ -563,7 +563,7 @@ void TraceWorker::run()
 
     std::chrono::steady_clock::time_point begin = std::chrono::steady_clock::now();
     qDebug() << "Begin to download the trace file to " << target.generic_string().c_str();
-    auto r = device->RetrieveTrace(*trace_file_path, target.generic_string(), false);
+    auto r = device->RetrieveTrace(*trace_file_path, target.generic_string());
     progress_bar_worker->terminate();
     m_progress_bar->setValue(100);
     if (r.ok())


capture_service/android_application.h
```diff
@@ -51,6 +51,15 @@ class AndroidApplication
     bool                 IsDebuggable() const { return m_is_debuggable; }
     bool                 IsStarted() const { return m_started; }
     virtual bool         IsRunning() const;
+    void                 gfxrEnabled(bool enable);
```
wangra-google, line 54: it would be better to be consistent with existing naming conventions for function names (start with an upper case letter)

capture_service/android_application.h
```diff
@@ -61,8 +70,13 @@ class AndroidApplication
     ApplicationType m_type;
     std::string     m_main_activity;
     std::string     m_command_args;
+    std::string     m_device_architecture;
```
wangra-google, line 73: could you put some comments on what are the available device architecture options?

capture_service/device_mgr.h
```diff
@@ -92,15 +96,17 @@ class AndroidDevice
...
+                                      const std::string &save_path,
+                                      const bool         isGfxrCapture);
 
 private:
     const std::string                   m_serial;
     DeviceInfo                          m_dev_info;
     AdbSession                          m_adb;
     DeviceState                         m_original_state;
     std::unique_ptr<AndroidApplication> app;
+    bool                                kGfxrEnabled;
```
wangra-google, line 109: `k` is normally used for `constant`, for members, you can use m_ as prefix

capture_service/device_mgr.h
```diff
@@ -92,15 +96,17 @@ class AndroidDevice
...
+    absl::Status        RetrieveTrace(const std::string &trace_file_path,
+                                      const std::string &save_path,
+                                      const bool         isGfxrCapture);
 
 private:
     const std::string                   m_serial;
     DeviceInfo                          m_dev_info;
     AdbSession                          m_adb;
     DeviceState                         m_original_state;
     std::unique_ptr<AndroidApplication> app;
```
wangra-google, line 108: not really your code, but could you update this member to have prefix `m_` to be consistent with other members? thanks :)

capture_service/dive_client_cli.cc
```diff
@@ -113,6 +121,27 @@ download_path,
 ".",
 "specify the full path to download the capture on the host, default to current directory.");
 
+ABSL_FLAG(
+std::string,
+device_architecture,
+"x86",
+"specify the device architecture to capture with gfxr. If not specified, the default is x86.");
```
wangra-google, line 128: could you add the other valid options in here? (so that the user don't have the check code to find all the valid options :) )

capture_service/dive_client_cli.cc
```diff
@@ -113,6 +121,27 @@ download_path,
...
+          "file name is gfxr_capture.");
+ABSL_FLAG(int,
+          frame,
+          -1,
+          "specify which frame to capture with gfxr. If not specified, the default is -1.");
+ABSL_FLAG(
+std::string,
+frame_range,
+"",
+"specify the range of frames to capture with gfxr. If not specified, the default is frame "
```
wangra-google, line 142: a bit confused here, if the user does not specify frame_range, what is the default value?

capture_service/dive_client_cli.cc
```diff
@@ -180,7 +209,11 @@ bool run_package(Dive::DeviceManager& mgr,
                  const std::string&   app_type,
                  const std::string&   package,
                  const std::string&   command,
-                 const std::string&   command_args)
+                 const std::string&   command_args,
+                 const std::string&   device_architecture,
+                 const std::string&   gfxr_capture_directory,
+                 const std::string&   gfxr_capture_frames,
+                 bool                 isGfxrCapture)
```
wangra-google, line 216: better to be consistent with other args on naming

capture_service/device_mgr.cc
```diff
@@ -140,15 +140,24 @@ absl::StatusOr<std::vector<std::string>> AndroidDevice::ListPackage(PackageListO
...
 {
     LOGD("cwd: %s\n", std::filesystem::current_path().c_str());
     std::vector<std::filesystem::path> search_paths{ std::filesystem::path{ "./install" },
                                                      std::filesystem::path{
                                                      "../../build_android/Release/bin" },
                                                      std::filesystem::path{ "../../install" },
                                                      std::filesystem::path{ "./" } };
-    std::filesystem::path              lib_path{ name };
+
+    if (device_architecture != "")
```
wangra-google, line 153: i guess device_architecture should never be ""? the capture would fail if it is ""?

capture_service/device_mgr.h
```diff
@@ -67,6 +67,7 @@ class AndroidDevice
     absl::Status Init();
     absl::Status SetupDevice();
     absl::Status CleanupDevice();
+    void         enableGfxr(bool enableGfxr);
```
wangra-google, line 70: same naming convention as above functions

capture_service/device_mgr.h
```diff
@@ -92,15 +96,17 @@ class AndroidDevice
     AdbSession       &Adb() { return m_adb; }
 
     AndroidApplication *GetCurrentApplication() { return app.get(); }
-    absl::Status        RetrieveTraceFile(const std::string &trace_file_path,
-                                          const std::string &save_path);
+    absl::Status        RetrieveTrace(const std::string &trace_file_path,
+                                      const std::string &save_path,
+                                      const bool         isGfxrCapture);
```
wangra-google, line 101: same naming convention as other args

wangra-google (dismissed):

capture_service/android_application.h
```diff
@@ -61,8 +70,14 @@ class AndroidApplication
...
-    bool            m_is_debuggable;
-    bool            m_started;
+    // Available architectures are arm64-v8, armeabi-v7a, x86, and x86_64.
+    std::string m_device_architecture;
+    std::string m_gfxr_capture_file_directory;
+    std::string m_gfxr_capture_frames;
+    bool        m_is_debuggable;
+    bool        m_started;
+
+    bool kGfxrEnabled;
```
wangra-google, line 80: better to be named as m_is_gfxr_enabled

capture_service/android_application.h
```diff
@@ -51,6 +51,15 @@ class AndroidApplication
...
     bool                 IsStarted() const { return m_started; }
     virtual bool         IsRunning() const;
+    void                 SetGfxrEnabled(bool enable);
+    void SetArchitecture(const std::string &architecture) { m_device_architecture = architecture; };
+    void SetFrames(const std::string &frames) { m_gfxr_capture_frames = frames; };
+    void SetGfxrCaptureFileDirectory(const std::string &capture_file_directory)
+    {
+        m_gfxr_capture_file_directory = capture_file_directory;
+    };
+    absl::Status createGfxrDirectory(const std::string command_args);
```
wangra-google, line 61: better to be consistent on naming convention with the other functions

capture_service/dive_client_cli.cc
```diff
@@ -113,6 +121,27 @@ download_path,
...
+"specify the device architecture to capture with gfxr. If not specified, the default is x86.");
+ABSL_FLAG(std::string,
+          gfxr_capture_file_dir,
+          "gfxr_capture",
+          "specify the name of the directory for the gfxr capture. If not specified, the default "
+          "file name is gfxr_capture.");
+ABSL_FLAG(int,
+          frame,
+          -1,
+          "specify which frame to capture with gfxr. If not specified, the default is -1.");
```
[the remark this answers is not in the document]
GrantComm: It is just a default value. Was thinking of changing it to 0 and then adjusting the check in the main() function.

capture_service/dive_client_cli.cc
GrantComm: Default will just be an empty string which is checked in the main() function

capture_service/device_mgr.cc
GrantComm: If it is an empty string then the gfxr layer will not be found. The device architecture is empty when running anything that is not gfxr capture.

wangra-google: in our case, should we add some warning if this field is missing? (i am a bit worried about silent failures. )

capture_service/dive_client_cli.cc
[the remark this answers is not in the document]
wangra-google: what is the behavior when this is set to -1? (it would be easier for the user to understand the behavior if we could add this in the comment. :) )

capture_service/dive_client_cli.cc
wangra-google: i think we could add some descriptions of the behavior of the case when the default value is used.

capture_service/dive_client_cli.cc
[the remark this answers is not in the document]
GrantComm: There is a check in the main function on for the gfxr case (line 461). If the default -1 is used, then the frame range value is checked. If both are the default values (-1 and "") the user is prompted with a message and the capture process is terminated.

capture_service/dive_client_cli.cc
wangra-google: the goal here is to provide more info about the default behavior so that the user knows what to expect :)

GrantComm: Ok, added more of a description about the default values for each of the capture frame flags.

capture_service/dive_client_cli.cc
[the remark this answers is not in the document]
wangra-google: thanks for the explanation and the updated comments :)

capture_service/dive_client_cli.cc
wangra-google: thanks

capture_service/device_mgr.cc
wangra-google: oh, i just see the default will be x86

wangra-google (approved):

capture_service/device_mgr.cc
```diff
@@ -381,11 +421,21 @@ absl::Status DeviceManager::Cleanup(const std::string &serial, const std::string
...
-absl::Status AndroidDevice::RetrieveTraceFile(const std::string &trace_file_path,
-                                              const std::string &save_path)
+absl::Status AndroidDevice::RetrieveTrace(const std::string &trace_path,
+                                          const std::string &save_path,
+                                          const bool         isGfxrCapture)
+{
+    RETURN_IF_ERROR(Adb().Run(absl::StrFormat("pull %s %s", trace_path, save_path)));
+    if (isGfxrCapture)
+    {
+        return Adb().Run(absl::StrFormat("shell rm -rf %s", trace_path));
```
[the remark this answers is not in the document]
GrantComm: For the gfxr capture path, the tracefile is saved in a specific directory that should be deleted after retrieving it. 

The regular capture process retrieves a file and then deletes it.

capture_service/device_mgr.cc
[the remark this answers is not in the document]
GrantComm: Ok that makes sense. Will have a separate PR for refactoring changes.

PR #155, opened 2025-01-03
Stl replacement

shanminchao:
This is a STL vector replacement to speed up Windows DEBUG + some command hierarchy optimizations that reduces vector usage overhead (e.g. using std::move, pre-reserving, etc)

On my Windows machine in DEBUG mode, overall load time for a particularly bad capture went from 1562s down to 466.14s with this PR.

In RELEASE mode for the same capture, it went from ~85s down to about ~77s. The RELEASE mode savings mostly results from the optimizations done in this PR, and less so from the vector replacement.

### DIFF
--- a/dive_core/command_hierarchy.cpp
+++ b/dive_core/command_hierarchy.cpp
@@ -2242,6 +2242,9 @@ uint64_t CommandHierarchyCreator::GetChildCount(CommandHierarchy::TopologyType t
 //--------------------------------------------------------------------------------------------------
 void CommandHierarchyCreator::CreateTopologies()
 {
+    uint64_t total_num_children[CommandHierarchy::kTopologyTypeCount] = {};
+    uint64_t total_num_shared_children[CommandHierarchy::kTopologyTypeCount] = {};
+
     // A kVulkanCallTopology is a kAllEventTopology without the following:
     //  kDrawDispatchDmaNode, kSyncNode, kPostambleStateNode, kMarkerNode-kBarrier
     auto FilterOut = [&](size_t node_index) {
@@ -2275,6 +2278,11 @@ void CommandHierarchyCreator::CreateTopologies()
 
         // Go through primary children of a particular node, and only add non-ignored nodes
         const DiveVector<uint64_t> &children = m_node_children[src_topology][0][node_index];
+
+        // Optionally pre-reserve the maximum size for performance reasons
+        // This may result in slightly more memory being used
+        m_node_children[dst_topology][0][node_index].reserve(children.size());
+
         for (size_t child = 0; child < children.size(); ++child)
         {
             if (!FilterOut(children[child]))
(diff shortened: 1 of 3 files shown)


dive_core/command_hierarchy.cpp
```diff
@@ -444,21 +444,21 @@ size_t CommandHierarchy::GetEventIndex(uint64_t node_index) const
...
+                                          char         *metadata_ptr,
+                                          uint32_t      metadata_size)
 {
     DIVE_ASSERT(m_node_type.size() == m_description.size());
     DIVE_ASSERT(m_node_type.size() == m_aux_info.size());
     DIVE_ASSERT(m_node_type.size() == m_metadata.size());
 
     m_node_type.push_back(type);
-    m_description.push_back(desc);
+    m_description.push_back(std::move(desc));
```
wangra-google, line 458: should we treat "aux_info" the same way to avoid the copy?

dive_core/command_hierarchy.cpp
```diff
@@ -444,21 +444,21 @@ size_t CommandHierarchy::GetEventIndex(uint64_t node_index) const
...
     m_aux_info.push_back(aux_info);
 
-    std::vector<uint8_t> temp(metadata_size);
+    DiveVector<uint8_t> temp((uint64_t)metadata_size);
```
wangra-google, line 461: why do we need to cast to uint64_t?

dive_core/stl_replacement.h
```diff
@@ -0,0 +1,75 @@
+/*
+ Copyright 2024 Google LLC
```
wangra-google, line 2: nit: 2025 :)

dive_core/stl_replacement.hpp
```diff
@@ -0,0 +1,267 @@
+/*
+ Copyright 2024 Google LLC
```
wangra-google, line 2: nit: 2025 :)

dive_core/stl_replacement.hpp
```diff
@@ -0,0 +1,267 @@
...
+}
+
+//--------------------------------------------------------------------------------------------------
+template<class Type> void Vector<Type>::resize(uint64_t size, const Type &a)
+{
+    reserve(size);
+
+    // Call the copy constructor for all the new elements
+    for (uint64_t i = m_size; i < size; ++i)
+        new (&m_buffer[i]) Type(a);
```
wangra-google, line 203: not sure how much better it is, but you may skip calling the constructor if the type is a simple type by checking `std::is_trivially_constructible`

dive_core/stl_replacement.hpp
```diff
@@ -0,0 +1,267 @@
...
+    // Do not call resize() directly, since it invokes default constructor
+    // And not all classes have default constructors
+    reserve(a.m_size);
+    m_size = a.m_size;
+    std::copy(a.m_buffer, a.m_buffer + a.m_size, m_buffer);
+    return *this;
+}
+
+//--------------------------------------------------------------------------------------------------
+template<class Type> Vector<Type> &Vector<Type>::operator=(Vector<Type> &&a)
```
wangra-google, line 104: maybe better with self assignment check?

dive_core/stl_replacement.hpp
```diff
@@ -0,0 +1,267 @@
...
+//--------------------------------------------------------------------------------------------------
+template<class Type> Type &Vector<Type>::operator[](uint64_t i) const
+{
+    DIVE_ASSERT(i < m_size);
+    return m_buffer[i];
+}
+
+//--------------------------------------------------------------------------------------------------
+template<class Type> Vector<Type> &Vector<Type>::operator=(const Vector<Type> &a)
+{
```
wangra-google, line 94: maybe better with self assignment check?

dive_core/command_hierarchy.cpp
shanminchao: aux_info doesn't have a move operator/constructor, and it's also a 64-bit struct, so doing a std::move shouldn't make any difference whatsoever

dive_core/stl_replacement.h
shanminchao: I technically wrote all this in 2024! But I changed it :)

dive_core/stl_replacement.hpp
shanminchao: Done

dive_core/stl_replacement.hpp
shanminchao: Done

wangra-google (approved):

PR #156, opened 2025-01-08
Additional Optimizations

shanminchao:
This PR covers the following optimizations:

* Deferred shader loading, which for a particularly bad capture saves about 15-20s load time (from a previous high of 72s to 56s, for example)
* Pre-reserve some vectors heavily used during command hierarchy construction. This saves about a couple of seconds from the load time, often lost in the noise

This PR also includes an assert fix when selecting the submit node.

### DIFF
--- a/ui/command_buffer_model.cpp
+++ b/ui/command_buffer_model.cpp
@@ -264,34 +264,38 @@ void CommandBufferModel::OnSelectionChanged(const QModelIndex &index)
     m_node_is_selected_bit_list.resize(bit_list_size);
     CreateNodeToParentMap(UINT64_MAX, root_node_index, false);
 
-    // Determine the scroll-to position
-    m_scroll_to_index = QModelIndex();
-    uint64_t end_node_index = m_topology_ptr->GetEndSharedChildNodeIndex(m_selected_node_index);
-    uint64_t parent_node_index = (uint64_t)(m_node_parent_list[end_node_index].internalPointer());
-    uint64_t num_children = m_topology_ptr->GetNumSharedChildren(parent_node_index);
-    for (uint64_t child = 0; child < num_children; ++child)
+    // Determine the scroll-to position if not at a root node
+    if (m_selected_node_index != root_node_index)
     {
-        uint64_t child_node_index = m_topology_ptr->GetSharedChildNodeIndex(parent_node_index,
-                                                                            child);
-        // Cache the row index for the specific end node
-        // Recall a parent has normal children + shared children (e.g. normal fields + packets)
-        // The row has to account for both
-        if (child_node_index == end_node_index)
+        m_scroll_to_index = QModelIndex();
+        uint64_t end_node_index = m_topology_ptr->GetEndSharedChildNodeIndex(m_selected_node_index);
+        uint64_t parent_node_index = (uint64_t)(m_node_parent_list[end_node_index]
+                                                .internalPointer());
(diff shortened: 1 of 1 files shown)


dive_core/shader_disassembly.h
```diff
@@ -54,25 +57,56 @@ class ShaderInstruction
...
+        return m_gpr_count;
+    }
 
 private:
-    std::string m_listing;
-    // std::vector<ShaderInstruction> m_instructions;
-    std::vector<std::string> m_instructions_text;
-    std::vector<uint64_t>    m_instructions_raw;
-    uint32_t                 m_gpr_count;
+    void Disassemble() const;
```
wangra-google, line 98: is there a reason you prefer this func to be const? personally i found it is a bit confusing since when i see const, i assume this func would not modify any member, then i realize the mutable keywords.

dive_core/command_hierarchy.h
```diff
@@ -368,9 +368,12 @@ class CommandHierarchyCreator : public IEmulateCallbacks
     // If flatten_chain_nodes set to true, then chain nodes are children of the top-most
     // root ib or call ib node, and never a child of another chain node. This prevents a
     // deep tree of chain nodes when a capture chains together tons of IBs.
+    // Optional: Passing a non-UINT64_MAX value to num_pm4_packets will allow the creator
+    // to pre-reserve the memory needed and potentially speed up the creation
     bool CreateTrees(CommandHierarchy  *command_hierarchy_ptr,
                      const CaptureData &capture_data,
                      bool               flatten_chain_nodes,
+                     uint64_t           num_pm4_packets,
```
wangra-google, line 376: maybe better with std::optional<uint64_t> reserve_size, and check "if (reserve_size.has_value())"

dive_core/shader_disassembly.h
shanminchao: It's "const" because the accessor functions (e.g. GetNumInstructions()) are const as well. And I want to call Disassemble() the first time those functions are called, and not before. This is to achieve the "defer" nature of the disassembly. 

So yeah, this is ugly. If you have any suggestions to make it cleaner, let me know!

dive_core/command_hierarchy.h
shanminchao: Done

dive_core/shader_disassembly.h
wangra-google: it is tricky, here is my idea (still not ideal):

1. if we decide to use `mutable`, maybe better to group all those mutable members into a struct so that we only need to put `mutable` keywords once
2. we can do a const_cast will calling Disassemble(). i feel this one is more explicit than using mutable so that when anyone sees the function implementation, it is explicit that the Disassemble() will modify members

```
struct DisassembledData
{
    bool                     m_disassembled;
    std::string              m_listing;
    std::vector<std::string> m_instructions_text;
    std::vector<uint64_t>    m_instructions_raw;
    uint32_t                 m_gpr_count;
};

public:
    uint32_t GetGPRCount() const { return getData().m_gpr_count; }

private:
    const DisassembledData& getData() const
    {
        if (!m_disassembledData)
        {
            const_cast<Disassembly*>(this)->Disassemble();
        }
        return *m_disassembledData;
    }

    std::optional<DisassembledData> m_disassembledData = std::nullopt;

```

what do you think?

shanminchao: I like it! Implemented as suggested.

wangra-google (approved):

PR #158, opened 2025-01-09
Run gRPC service independently

elviscapiaq:


elviscapiaq:
I put this PR as a draft to validate the implementation. Please let me know if I am moving in the right track.
Also, I did the testing with several apps, including the Unity app with two-instance creation. For all tests, the gRPC service runs successfully.

capture_service/service.cc
```diff
@@ -132,6 +132,8 @@ grpc::Status DiveServiceImpl::DownloadFile(grpc::ServerContext             *cont
     return grpc::Status::OK;
 }
 
+std::unique_ptr<grpc::Server> server;
```
wangra-google, line 135: i would avoid using global var. maybe we could do something like this?
```
std::unique_ptr<grpc::Server> GetServer() {
  static std::unique_ptr<grpc::Server> server; // No initialization here
  return server; 
}

void InitializeServer() {
  auto& server = GetServer(); // Get reference to the unique_ptr
  server = grpc::ServerBuilder().BuildAndStart(); 
}

void StopServer() {
  auto& server = GetServer(); // Get reference to the unique_ptr
  if (server) {
    server->Shutdown(); 
  }
} 
```

elviscapiaq: Thanks, I applied the suggested changes.

layer/openxr_layer.cc
```diff
@@ -56,13 +58,21 @@ struct XrSessionData
 {
     XrSession                session;
     XrGeneratedDispatchTable dispatch_table;
-    ServerRunner            &server;
+};
 
-    XrSessionData() :
-        server(GetServerRunner())
+struct MyServer
```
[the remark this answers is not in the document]
elviscapiaq: Done.

elviscapiaq:
> you'll also need to update the Vulkan side as well, otherwise the server will be started twice for Vulkan application. please also test with Vulkan application, OpenXR application, and also the shared Unity application that we know has problem.
I added also a global object to initialize the server for an Vulkan app, however, it is not working. When I move the server initialization to `DiveInterceptCreateInstance` at `vk_layer_base.cc` it works.

layer/vk_layer_base.cc
```diff
@@ -515,6 +523,10 @@ extern "C"
             return (PFN_vkVoidFunction)&DiveInterceptEnumerateInstanceExtensionProperties;
         if (0 == strcmp(func, "vkCreateInstance"))
             return (PFN_vkVoidFunction)&DiveInterceptCreateInstance;
+        if (0 == strcmp(func, "vkDestroyInstance"))
+        {
+            LOGI("vkDestroyInstance");
```
wangra-google, line 528: why do we need this func? only for debug log?

elviscapiaq:
I have updated this PR which now works for all tests I have.
The problem was that the loading/unloading behavior of the Vulkan/OpenXR Dive layer library varies significantly across Android applications.For example for Vulkan apps it loads - unloads - reloads the library during execution. For an Unity-based OpenXR
app that creates two instances, the behavior is the same.
    
To addres this now the gRPC Server starts when the library loads for the first time. Then, a function prevents the library from being unloaded. This keeps both the library and server loaded exactly once, even if apps try to reload or close them. In this way, the Server lives for the entire lifecycle of the app.

layer/vk_layer_base.cc
wangra-google: do we actually need any interception function here? or the `if` check is just for extra debug print?

elviscapiaq: Not really. I will delete it, thanks.
Also, I am looking into WIndows to see why it is crashing (On Linux everthing looks good).

elviscapiaq:
I updated the PR to support Windows as well.

elviscapiaq:
I tested on Windows, and everything work fine.

wangra-google (approved):

PR #159, opened 2025-01-10
Fixes for building gfxr on Windows targeting Android

angela28chen:
Some changes to get Dive Android libraries building on Windows after the GFXR addition.
+ Triggering gfxr gradle builds and extracting/copying files is now only done inside CMakeLists.txt for the Linux host build, this has been implemented in `scripts/build_android.bat` for Windows.
+ The gfxr builds for different versions (Debug/Release) are separated so that it's easier to only build one by modifying `BUILD_TYPE` in `scripts/build_android.bat`
+ Different from the Linux host build, the replay folder is copied to `build_android/Debug/third_party/gfxreconstruct/android/tools/replay/` rather than `build_android/Debug/third_party/gfxreconstruct/android/tools/replay/build/outputs/apk/debug`
+ Updated documentation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,7 +80,7 @@ if(DIVE_NO_DISASSEMBLY)
     add_definitions(-DDIVE_NO_DISASSEMBLY=1)
 endif()
 
-option(DIVE_ENABLE_PERFETTO, "Enalbe functionality to get perfetto trace." OFF)
+option(DIVE_ENABLE_PERFETTO, "Enable functionality to get perfetto trace." OFF)
 include(perfetto.cmake)
 
 add_subdirectory(capture_service)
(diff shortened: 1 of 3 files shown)


wangra-google (approved):

GrantComm (approved):

angela28chen:
I agree it's better to be consistent between the two platforms, but I tried again and I am still unable to get the cmake calling gradle to function properly on Windows so I think modifying the batch file is still needed. I plan to make a followup CL for Linux with the following:
1. Disentangle Debug/Release builds
2. Change replay folder location to match the one specified in this PR on Windows

I verified that on Linux, triggering the gradle build from within cmake file works well. I'm not sure it's worth the effort to change that to the sh file just for consistency, but I could look into that if it's desirable. That will be remaining difference.

PR #160, opened 2025-01-10
Ask user to grant INTERNET permission + grant all-files access

elviscapiaq:
This PR is an starting point of the implementation. 
Currently, there's no way to determine if the target Android app has the `android.permission.INTERNET` granted.
Since `android.permission.INTERNET` is a normal permission, it must be declared at install time (https://developer.android.com/training/permissions/declaring).  Therefore, we cannot modify it at runtime.

The current solution checks if `android.permission.INTERNET` is granted. If not, it returns an error message asking the user to add this permission in the AndroidManifest.xml file.

Remember we need `android.permission.INTERNET` granted to use the gRPC service. On the other hand, since Android 11 we don't need to grant the android.permission.WRITE_EXTERNAL_STORAGE, It is depreciated. For future needs we can use `MANAGE_EXTERNAL_STORAGE. to access specific files.

### DIFF
--- a/capture_service/android_application.cc
+++ b/capture_service/android_application.cc
@@ -182,6 +182,8 @@ VulkanApplication::~VulkanApplication()
 absl::Status VulkanApplication::Setup()
 {
     LOGD("Setup Vulkan application: %s\n", m_package.c_str());
+    RETURN_IF_ERROR(HasInternetPermission());
+    RETURN_IF_ERROR(GrantAllFilesAccess());
     RETURN_IF_ERROR(m_dev.Adb().Run("root"));
     RETURN_IF_ERROR(m_dev.Adb().Run("wait-for-device"));
     Stop().IgnoreError();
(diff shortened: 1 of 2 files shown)


capture_service/dive_client_cli.cc
```diff
@@ -249,6 +265,16 @@ bool run_package(Dive::DeviceManager& mgr,
         return false;
     }
 
+    std::string command_query = absl::
+    StrFormat("adb shell dumpsys package %s | grep \"android.permission.INTERNET: granted=true\"",
```
wangra-google, line 269: curious, what is the difference between this adb cmd and editing the manifest file?

capture_service/dive_client_cli.cc
```diff
@@ -216,6 +216,22 @@ bool list_package(Dive::DeviceManager& mgr, const std::string& device_serial)
     return true;
 }
 
+std::string GetCommandOutput(const std::string& command)
```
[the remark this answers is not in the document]
elviscapiaq: Updated!

elviscapiaq:
I updated this PR including the granting process for  ` MANAGE_EXTERNAL_STORAGE`.

capture_service/dive_client_cli.cc
elviscapiaq: This adb cmd only checks if the `android.permission.INTERNET` was granted at installation time (i.e. in the Manifest file). On the other hand, the manifest file is where we should grant this permission.

PR #165, opened 2025-01-27
Additional load time optimization

shanminchao:
The big wins here are from converting many of the critical path functions from using maps to DiveVectors in pm4_info.cpp.
Additionally, some additional pre-reservations are made in commandhierarchy.

On my Windows machine, for a particularly heavy capture, the RELEASE load time decreased by about 3-7 seconds, from about 56s total to about 50s. (There's a lot of variance in profiling in Windows). DEBUG load time is currently at about ~330s, about a 40s improvement from the previous 370s.

On my beefier work Linux machine, load times in RELEASE went down from 23s to about 21.5s.

### DIFF
--- a/dive_core/command_hierarchy.cpp
+++ b/dive_core/command_hierarchy.cpp
@@ -458,10 +458,14 @@ uint64_t CommandHierarchy::Nodes::AddNode(NodeType      type,
     m_description.push_back(std::move(desc));
     m_aux_info.push_back(aux_info);
 
-    DiveVector<uint8_t> temp(metadata_size);
     if (metadata_ptr != nullptr)
+    {
+        DiveVector<uint8_t> temp(metadata_size);
         memcpy(&temp[0], metadata_ptr, metadata_size);
-    m_metadata.push_back(std::move(temp));
+        m_metadata.push_back(std::move(temp));
+    }
+    else
+        m_metadata.resize(m_metadata.size() + 1);
 
     return m_node_type.size() - 1;
 }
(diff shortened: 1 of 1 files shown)


wangra-google:
> The big wins here are from converting many of the critical path functions from using maps to DiveVectors in pm4_info.cpp. Additionally, some additional pre-reservations are made in commandhierarchy.
> 
> On my Windows machine, for a particularly heavy capture, the RELEASE load time decreased by about 3-7 seconds, from about 56s total to about 50s. (There's a lot of variance in profiling in Windows). DEBUG load time is currently at about ~330s, about a 40s improvement from the previous 370s.
> 
> On my beefier work Linux machine, load times in RELEASE went down from 23s to about 21.5s.

That is great! :D

dive_core/generatePm4Info_adreno.py
```diff
@@ -219,6 +221,13 @@ def outputPm4InfoInitFunc(pm4_info_file, registers_et_root, opcode_dict):
 valid_opcodes = {}
 # ---------------------------------------------------------------------------------------
 def outputOpcodes(pm4_info_file, opcode_dict):
+  # Find max opcode first
```
wangra-google, line 224: would "max_opcode = max(opcode_dict)" work?

dive_core/generatePm4Info_adreno.py
```diff
@@ -391,6 +400,14 @@ def outputRegisterInfo(pm4_info_file, registers_et_root, enum_index_dict):
     if is_reg_32 or is_reg_64:
       regs.append(element)
 
+  # Determine highest register offset
+  max_offset = 0
+  for reg in regs:
+    offset = int(reg.attrib['offset'],0)
+    if offset > max_offset:
+      max_offset = offset
```
wangra-google, line 408: better with `max_offset = max(max_offset, offset)`?

dive_core/stl_replacement.hpp
```diff
@@ -73,6 +76,7 @@ Vector<Type>::Vector(std::initializer_list<Type> a) :
     m_size(0)
```
wangra-google, line 76: can directly assign the value here?

dive_core/stl_replacement.hpp
```diff
@@ -85,7 +89,7 @@ template<class Type> Vector<Type>::~Vector()
 //--------------------------------------------------------------------------------------------------
 template<class Type> Type &Vector<Type>::operator[](uint64_t i) const
 {
-    DIVE_ASSERT(i < m_size);
```
wangra-google, line 88: is there a reason not using DIVE_ASSERT?

dive_core/stl_replacement.hpp
shanminchao: reserve() function references m_size, so have to set it to 0 first

dive_core/stl_replacement.hpp
shanminchao: Good point. pm4_info.cpp didn't have access to DIVE_ASSERT, but that doesn't mean we can't include the common.h header here to get access to DIVE_ASSERT here. Modified it back.

dive_core/generatePm4Info_adreno.py
shanminchao: Good point. Changed as suggested.

dive_core/generatePm4Info_adreno.py
shanminchao: Good point. Changed as suggested.

wangra-google (approved):

PR #166, opened 2025-02-11
Use Unix (local) domain socket in abstract namespace for server

elviscapiaq:
- Currently it works for Linux. No INTERNET permission is needed for apps (Vulkan and OpenXR).

### DIFF
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -192,7 +192,8 @@ absl::Status AndroidDevice::SetupDevice()
         Adb().Run(absl::StrFormat("push %s %s",
                                   ResolveAndroidLibPath(kXrLayerLibName, "").generic_string(),
                                   kTargetPath)));
-        RETURN_IF_ERROR(Adb().Run(absl::StrFormat("forward tcp:%d tcp:%d", kPort, kPort)));
+        RETURN_IF_ERROR(
+        Adb().Run(absl::StrFormat("forward tcp:%d localabstract:dive_%d", kPort, kPort)));
     }
 
 #if defined(DIVE_ENABLE_PERFETTO)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -107,30 +107,30 @@ if(NOT ANDROID)
         add_subdirectory(ui)
     endif()
 
-    add_subdirectory(third_party/gfxreconstruct)
+    # add_subdirectory(third_party/gfxreconstruct)
```
wangra-google, line 110: why do you want to disable gfxr?

capture_service/device_mgr.cc
```diff
@@ -192,7 +192,7 @@ absl::Status AndroidDevice::SetupDevice()
         Adb().Run(absl::StrFormat("push %s %s",
                                   ResolveAndroidLibPath(kXrLayerLibName, "").generic_string(),
                                   kTargetPath)));
-        RETURN_IF_ERROR(Adb().Run(absl::StrFormat("forward tcp:%d tcp:%d", kPort, kPort)));
+        RETURN_IF_ERROR(Adb().Run(absl::StrFormat("forward tcp:%d localabstract:my_abstract_socket", kPort)));
```
wangra-google, line 195: better remove `my_`

CMakeLists.txt
elviscapiaq: I just wanted to test it, no really removed it.

capture_service/device_mgr.cc
elviscapiaq: Update!

elviscapiaq:
I updated the PR for review now. Dive capturing works properly for Vulkan and OpenXR apps without INTERNET permission.
Using Unix (local) domain socket in abstract namespace for server avoids the need to grant INTERNET permission to the target application. Also, no file-based permissions apply since it is in an abstract namespace.

wangra-google (approved):

PR #167, opened 2025-02-18
Enable dynamic port allocation for the client.

elviscapiaq:
- If the target port is already occupied by another application on the client, the client should utilize a different, dynamically selected port.
- Because the server employs an abstract socket, port availability is not a concern on the server side.

### DIFF
--- a/capture_service/constants.h
+++ b/capture_service/constants.h
@@ -32,6 +32,10 @@ inline constexpr char kGfxrRuntimeCapture[] = "gfxr_runtime_capture";
 inline constexpr char kManifestFileName[] = "XrApiLayer_dive.json";
 inline constexpr char kManifestFilePath[] = "/system/etc/openxr/1/api_layers/implicit.d/";
 inline constexpr char kVulkanGlobalPath[] = "/data/local/debug/vulkan";
-inline constexpr int  kPort = 19999;
 inline constexpr int  kDownLoadFileChunkSize = 4096;
+inline constexpr char kUnixAbstractPath[] = "dive_abstract";
+inline constexpr int
+kFirstPort = 49391;  // A port number within the dynamic port range (49152 to 65535)
+inline constexpr int
+kPortRange = 7;  // A small range of ports should be enough to find an available one
 }  // namespace Dive
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -46,7 +46,8 @@ std::string DeviceInfo::GetDisplayName() const
 
 AndroidDevice::AndroidDevice(const std::string &serial) :
     m_serial(serial),
-    m_adb(serial)
+    m_adb(serial),
+    m_port(kFirstPort)
 {
 }
 
@@ -173,6 +174,24 @@ std::filesystem::path ResolveAndroidLibPath(const std::string &name,
     return lib_path;
 }
 
+absl::Status AndroidDevice::ForwardFirstAvailablePort()
+{
+    for (int p = kFirstPort; p <= kFirstPort + kPortRange; p++)
+    {
+        auto res = Adb().RunAndGetResult(
+        absl::StrFormat("forward tcp:%d localabstract:%s", p, kUnixAbstractPath));
+        if (res.ok())
+        {
+            m_port = p;
+            return absl::OkStatus();
+        }
+    }
+    return absl::Status(absl::StatusCode::kUnknown,
+                        absl::StrFormat("Failed to forward available port between %d and %d",
+                                        kFirstPort,
+                                        kFirstPort + kPortRange));
+}
+
 absl::Status AndroidDevice::SetupDevice()
 {
     RETURN_IF_ERROR(Adb().Run("shell setenforce 0"));
@@ -192,8 +211,7 @@ absl::Status AndroidDevice::SetupDevice()
         Adb().Run(absl::StrFormat("push %s %s",
                                   ResolveAndroidLibPath(kXrLayerLibName, "").generic_string(),
                                   kTargetPath)));
-        RETURN_IF_ERROR(
-        Adb().Run(absl::StrFormat("forward tcp:%d localabstract:dive_%d", kPort, kPort)));
+        RETURN_IF_ERROR(ForwardFirstAvailablePort());
     }
 
 #if defined(DIVE_ENABLE_PERFETTO)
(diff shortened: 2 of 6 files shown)


capture_service/constants.h
```diff
@@ -32,6 +32,8 @@ inline constexpr char kGfxrRuntimeCapture[] = "gfxr_runtime_capture";
 inline constexpr char kManifestFileName[] = "XrApiLayer_dive.json";
 inline constexpr char kManifestFilePath[] = "/system/etc/openxr/1/api_layers/implicit.d/";
 inline constexpr char kVulkanGlobalPath[] = "/data/local/debug/vulkan";
-inline constexpr int  kPort = 19999;
 inline constexpr int  kDownLoadFileChunkSize = 4096;
+inline constexpr char kUnixAbstractPath[] = "dive_abstract";
+inline constexpr int  kFirstPort = 39999;
+inline constexpr int  kLastPort = kFirstPort + 7;
```
wangra-google, line 38: why do we choose 7? and why is the first port 39999?

capture_service/device_mgr.cc
```diff
@@ -39,6 +39,12 @@ DeviceManager &GetDeviceManager()
     return mgr;
 }
 
+int &Port()
+{
+    static int port = 0;
```
wangra-google, line 44: just curious, why do you prefer a static var instead of a member of AndroidDevice?

capture_service/constants.h
elviscapiaq: It was a typo, I wanted to put the first port equal to 49999 (in the range of a dynamic port 49152–65535).
Since the chances to have an process using a port between this range is low, 7 should be more than enough.
Do you have any recommndation for these ports?

wangra-google: could you add comments on why 49999 & 7 are chosen so that in the future if people want to make any change, they know how to adjust those values. Thanks :)

elviscapiaq: Done.

capture_service/device_mgr.cc
elviscapiaq: Thanks for the question. I updated to use as a member of AndroidDevice.

capture_service/device_mgr.h
```diff
@@ -94,6 +95,7 @@ class AndroidDevice
     absl::Status      StopApp();
     const AdbSession &Adb() const { return m_adb; }
     AdbSession       &Adb() { return m_adb; }
+    int               Port() { return m_port; }
```
wangra-google, line 98: const

capture_service/device_mgr.h
```diff
@@ -105,6 +107,7 @@ class AndroidDevice
     DeviceState                         m_original_state;
     std::unique_ptr<AndroidApplication> m_app;
     bool                                m_gfxr_enabled;
+    int                                 m_port;
```
wangra-google, line 110: maybe better to initialize m_port in the constructor?

capture_service/device_mgr.h
elviscapiaq: Done.

capture_service/device_mgr.h
elviscapiaq: Agreed!

capture_service/device_mgr.cc
```diff
@@ -44,9 +44,10 @@ std::string DeviceInfo::GetDisplayName() const
     return absl::StrCat(m_manufacturer, " ", m_model, " (", m_serial, ")");
 }
 
-AndroidDevice::AndroidDevice(const std::string &serial) :
```
wangra-google, line 47: i feel letting user choose port might be problematic since we explicitly defines kLastPort. Also the user might choose something outside the valid range (49152 to 65535). maybe initialize port to  kFirstPort should be good enough?

capture_service/constants.h
```diff
@@ -32,6 +32,11 @@ inline constexpr char kGfxrRuntimeCapture[] = "gfxr_runtime_capture";
 inline constexpr char kManifestFileName[] = "XrApiLayer_dive.json";
 inline constexpr char kManifestFilePath[] = "/system/etc/openxr/1/api_layers/implicit.d/";
 inline constexpr char kVulkanGlobalPath[] = "/data/local/debug/vulkan";
-inline constexpr int  kPort = 19999;
 inline constexpr int  kDownLoadFileChunkSize = 4096;
+inline constexpr char kUnixAbstractPath[] = "dive_abstract";
+inline constexpr int
+kFirstPort = 49391;  // A port number within the dynamic port range (49152 to 65535)
+inline constexpr int
+kLastPort = kFirstPort + 7;  // A small range of ports should be enough to find an available one
```
wangra-google, line 40: maybe it is better to define `kPortRange = 7` instead of defining  kLastPort in case we want to let user to define custom port in the future.

capture_service/device_mgr.cc
elviscapiaq: Updated.

capture_service/constants.h
elviscapiaq: Updated.

wangra-google (approved):

capture_service/dive_client_cli.cc
```diff
@@ -293,7 +292,7 @@ bool run_package(Dive::DeviceManager& mgr,
 
 bool trigger_capture(Dive::DeviceManager& mgr)
 {
-    std::string target_str = absl::GetFlag(FLAGS_target);
+    std::string target_str = absl::StrFormat("localhost:%d", mgr.GetDevice()->Port());
```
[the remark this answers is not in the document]
elviscapiaq: Thanks for poiting it out. I updated the code and now it also works on Dive UI.

capture_service/constants.h
```diff
@@ -32,6 +32,10 @@ inline constexpr char kGfxrRuntimeCapture[] = "gfxr_runtime_capture";
 inline constexpr char kManifestFileName[] = "XrApiLayer_dive.json";
 inline constexpr char kManifestFilePath[] = "/system/etc/openxr/1/api_layers/implicit.d/";
 inline constexpr char kVulkanGlobalPath[] = "/data/local/debug/vulkan";
-inline constexpr int  kPort = 19999;
 inline constexpr int  kDownLoadFileChunkSize = 4096;
+inline constexpr char kUnixAbstractPath[] = "dive_abstract";
```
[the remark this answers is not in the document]
elviscapiaq: Because it is an abstract socket, this is not a path in the file system, it is just like a local variable that lives while its process is running (namespace of the kernel's memory). Therefore, we don't need to worry about cleaning it up.

wangra-google (approved):

PR #169, opened 2025-02-18
Add support for CP_CONTEXT_REG_BUNCH in state tracker

shanminchao:
The state tracker wasn't parsing CP_CONTEXT_REG_BUNCH, and as a result all state set with that PM4 packet are not reflected in the render state view.


wangra-google (approved):

wangra-google:
> The state tracker wasn't parsing CP_CONTEXT_REG_BUNCH, and as a result all state set with that PM4 packet are not reflected in the render state view.

nice. curious how come it works for some of the captures? i guess CP_CONTEXT_REG_BUNCH is only used for gles apps?

shanminchao:
> > The state tracker wasn't parsing CP_CONTEXT_REG_BUNCH, and as a result all state set with that PM4 packet are not reflected in the render state view.
> 
> nice. curious how come it works for some of the captures? i guess CP_CONTEXT_REG_BUNCH is only used for gles apps?

I guess the other captures (i.e. Vulkan) rely more on type4 packets to set state?

wangra-google:
> > > The state tracker wasn't parsing CP_CONTEXT_REG_BUNCH, and as a result all state set with that PM4 packet are not reflected in the render state view.
> > 
> > 
> > nice. curious how come it works for some of the captures? i guess CP_CONTEXT_REG_BUNCH is only used for gles apps?
> 
> I guess the other captures (i.e. Vulkan) rely more on type4 packets to set state?

With this change, we can see the renderstates being updated correctly? (so not the big buffer issue?)

wangra-google (approved):

wangra-google:
> > > > > The state tracker wasn't parsing CP_CONTEXT_REG_BUNCH, and as a result all state set with that PM4 packet are not reflected in the render state view.
> > > > 
> > > > 
> > > > nice. curious how come it works for some of the captures? i guess CP_CONTEXT_REG_BUNCH is only used for gles apps?
> > > 
> > > 
> > > I guess the other captures (i.e. Vulkan) rely more on type4 packets to set state?
> > 
> > 
> > With this change, we can see the renderstates being updated correctly? (so not the big buffer issue?)
> 
> With the trace you had in mind, with this change, the render-mode setting is now correct. However, many of the other states still look wrong to me.

good to know, thanks.

PR #172, opened 2025-02-26
Fix an issue in gRPC initialization

elviscapiaq:
A Static Initialization Order Fiasco occurred. The static ServerRunner variable depends on the static server object. However, the initialization order is not guaranteed in this scenario, meaning ServerRunner could be created before server is initialized.

### DIFF
--- a/capture_service/service.cc
+++ b/capture_service/service.cc
@@ -130,23 +130,6 @@ grpc::Status DiveServiceImpl::DownloadFile(grpc::ServerContext             *cont
     return grpc::Status::OK;
 }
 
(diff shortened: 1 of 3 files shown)


wangra-google (approved):

elviscapiaq:
I also updated the PR to prevent a "Static Initialization Order Fiasco" between the static `init_server` and `server` (from `GetServer()`) While this doesn't currently affect normal gRPC functionality, it could impact order of execution of destructors in the future, which is a potential issue. I think any potential side effects are now highly unlikely.

wangra-google (approved):

PR #179, opened 2025-03-05
Fixes capture stuck in the downloading progress bar

elviscapiaq:
- It was used the `DownloadFile` function from `DiveClient` to pass a callback function that keeps track the current downloaded size.

### DIFF
--- a/ui/trace_window.h
+++ b/ui/trace_window.h
@@ -49,6 +49,7 @@ public:
     }
 signals:
     void TraceAvailable(const QString &);
+    void DownloadedSize(uint64_t size);
 
 private:
     QProgressDialog *m_progress_bar;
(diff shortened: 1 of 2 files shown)


ui/trace_window.h
```diff
@@ -87,15 +88,22 @@ class ProgressBarWorker : public QThread
         m_progress_bar(pd),
         m_capture_name(path),
         m_capture_size(size),
-        m_gfxr_capture(is_gfxr_capture)
+        m_gfxr_capture(is_gfxr_capture),
+        m_downloaded_size(0)
     {
     }
 
+    int64_t GetDownloadedSize() { return m_downloaded_size; }
```
wangra-google, line 96: const

wangra-google (approved):

PR #180, opened 2025-03-07
Build the Dive OpenXR/Vulkan Layer only for Android

elviscapiaq:
- We have been building the Dive Layer for non-Android platforms like
  Windows, however, we only want this library for Android (device).

### DIFF
--- a/layer/CMakeLists.txt
+++ b/layer/CMakeLists.txt
@@ -16,6 +16,13 @@
 
 cmake_minimum_required(VERSION 3.5)
 
+if (ANDROID)
+    message("Build diveCaptureLayer for Android.")
+else()
+    message("Not Android platform.")
+    return()
+endif()
+
 project(diveCaptureLayer)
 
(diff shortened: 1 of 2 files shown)


wangra-google:
i have mixed feeling on this change. personally i still use vs2022 to build the windows version to make sure no compiling errors in the layer code. it is easier to detect errors in vs2022 in general. @RenfengLiu what do you think?

wangra-google:
> > i have mixed feeling on this change. personally i still use vs2022 to build the windows version to make sure no compiling errors in the layer code. it is easier to detect errors in vs2022 in general. @RenfengLiu what do you think?
> 
> I'm not sure about the errors, if this is the case, probably we'll need to adjust the cmake flags to show them. Personally, I think this is better as there's fewer maintenance needs to support windows build of the layer, as we dont' really need the layer to run,and the pm4 capture stuff won'd work on windows as well.

not only showing them, but also click the error and redirect quickly to the errors.

wangra-google (approved):

PR #181, opened 2025-03-11
Includes asset file with gfxr capture file

GrantComm:
Includes asset file with gfxr capture file. After a capture, both the asset file and the capture file are pulled onto the local machine.

### DIFF
--- a/dive_core/shader_disassembly.cpp
+++ b/dive_core/shader_disassembly.cpp
@@ -114,7 +114,16 @@ void Disassembly::Disassemble()
     {
         DisassembledData disassembled_data;
         uint64_t         max_size = m_mem_manager.GetMaxContiguousSize(m_submit_index, m_address);
-        uint8_t*         data_ptr = new uint8_t[max_size];
+
+        // The disassembler does not early-out when it encounters an "end" instruction (at least not
+        // in its "prepass"), so passing it a too-big max_size can make the disassembly very slow!
+        // Let's set an arbitrary limit for now. The "correct" fix would be for the disassembler to
+        // early-out.
+        uint64_t kMaxSizeLimit = 64 * 1024;
+        if (max_size > kMaxSizeLimit)
+            max_size = kMaxSizeLimit;
+
+        uint8_t* data_ptr = new uint8_t[max_size];
         DIVE_VERIFY(m_mem_manager.CopyMemory(data_ptr, m_submit_index, m_address, max_size));
 
         struct shader_stats stats;
--- a/layer/openxr_layer.cc
+++ b/layer/openxr_layer.cc
@@ -104,7 +104,6 @@ XRAPI_ATTR XrResult XRAPI_CALL ApiDiveLayerXrEndFrame(XrSession             sess
 {
     XrResult result = XrResult::XR_ERROR_HANDLE_INVALID;
 
-    LOGD("ApiDiveLayerXrEndFrame key %lu, sess %p \n", DataKey(session), session);
     auto sess_data = GetXrSessionLayerData(DataKey(session));
     if (sess_data)
     {
--- a/third_party/freedreno/wrap/dive-wrap.c
+++ b/third_party/freedreno/wrap/dive-wrap.c
@@ -18,6 +18,12 @@ limitations under the License.
 
 #include <pthread.h>
 #include <stdlib.h>
+#include <sys/system_properties.h>
+
+
+// TODO: (renfeng) update the capture interface so that user can set this property value.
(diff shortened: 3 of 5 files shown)


capture_service/dive_client_cli.cc
```diff
@@ -377,21 +457,25 @@ void trigger_gfxr_capture(Dive::DeviceManager& mgr,
...
                 if (!ret.ok())
                 {
                     std::cout << "There was an error starting the gfxr runtime capture."
                               << std::endl;
                     return;
                 }
                 is_capturing = true;
-                std::cout << "Capture started." << std::endl;
+                std::cout << "Capture started. Press g+enter to stop the capture." << std::endl;
             }
```
wangra-google, line 479: sorry i am a bit confused here: I thought we only capture 1 frame?

GrantComm: Yes, only a single frame is captured in each capture file. The second g+enter is used to set capture_android_trigger to false and retrieve the capture from the device. I can reword it so that it is "Press g+enter to retrieve the capture" instead

wangra-google: yeah, i think that would be better. also need to update the comments in line 429

third_party/gfxreconstruct/framework/encode/capture_manager.cpp
```diff
@@ -1202,6 +1202,8 @@ void CommonCaptureManager::ActivateTrimming(std::shared_lock<ApiCallMutexT>& cur
         GFXRECON_ASSERT(thread_data != nullptr);
         if (use_asset_file_)
         {
+            // GOOGLE: Clears the asset file name for each trimmed capture.
+            asset_file_name_.clear();
```
angela28chen, line 1206: Just out of curiosity, what happens if you don't clear the asset file name here?

capture_service/dive_client_cli.cc
```diff
@@ -344,13 +346,92 @@ absl::Status is_capture_directory_busy(Dive::DeviceManager& mgr,
...
+        if (!std::filesystem::create_directories(target_download_path, ec))
+        {
+            std::cout << "Error creating directory: " << ec << std::endl;
+            return false;
+        }
+    }
+    else
+    {
+        // If the target directory already exists on the local machine, append a number to it to
+        // differentiate.
```
angela28chen, line 371: Do we want this? I feel like modifying/overwriting the exact dir specified would be more intuitive.

capture_service/android_application.cc
```diff
@@ -278,6 +278,8 @@ absl::Status AndroidApplication::GfxrSetup()
 
     RETURN_IF_ERROR(m_dev.Adb().Run("shell setprop debug.gfxrecon.capture_trigger_frames 1"));
 
+    RETURN_IF_ERROR(m_dev.Adb().Run("shell setprop debug.gfxrecon.capture_use_asset_file true"));
```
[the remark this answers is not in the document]
GrantComm: Yes, I added the reset of those props on application cleanup

capture_service/dive_client_cli.cc
```diff
@@ -344,13 +346,91 @@ absl::Status is_capture_directory_busy(Dive::DeviceManager& mgr,
                              absl::InternalError("Capture file operation in progress.");
 }
 
+bool retrieve_gfxr_capture(Dive::DeviceManager& mgr, const std::string& gfxr_capture_directory)
```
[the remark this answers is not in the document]
GrantComm: Ok, yes a similar process happens in the ui but the logic is a little different since the ui uses the progress bar. I will try to modify so there is more shared between the ui and the cli when handling the retrieval of files.

third_party/gfxreconstruct/framework/encode/capture_manager.cpp
GrantComm: If the asset file name is not cleared, the same file is used for subsequent captures during the session. Clearing the name ensures a new asset file is created for each single frame capture.

capture_service/dive_client_cli.cc
GrantComm: I added this so that it would be easier to differentiate captures so that there would be no need to clear the existing directory and replace the files. This way you could do a single frame capture and then another single frame capture without having to restart the whole capture process.

angela28chen (approved):
Thanks Grant for following up offline with a brief discussion around folder vs file stuff! This change LGTM and I will put up a PR to make replay be consistent with this change (using flag for folder path rather than GFXR file path).

PR #192, opened 2025-04-02
Update README to use new app names

footballhead:
Since https://github.com/google/bigwheels/commit/df7b4cace0858b1edcde3bfdc7983a66838cdd09, BigWheels projects no longer have digit prefixes. This tripped me up when I was trying to run commands because I was getting cryptic errors like:

```
Error: No UID for com.google.bigwheels.project_04_cube_xr.debug in user 0“.
```

Hopefully, this will help others onboard smoother.


angela28chen (approved):
Thanks!

PR #193, opened 2025-04-02
Select the device to avoid a cleanup crash

footballhead:
I encountered a crash with `dive_client_cli --command cleanup`. DeviceManager::Cleanup calls GetDevice (m_device accessor) to clean up any ports that it was using. However, m_device was nullptr since DeviceManager::SelectDevice was never called. Call SelectDevice before calling Cleanup to avoid the crash.

With this fix, cleanup no longer crashes and the program completes with exit status 0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ Examples:
 
  - Install the dependencies on device and start the package
  ```
- ./dive_client_cli --device 9A221FFAZ004TL --command run --package com.google.bigwheels.project_04_cube_xr.debug --type openxr --download_path "/path/to/save/captures"
+ ./dive_client_cli --device 9A221FFAZ004TL --command run --package com.google.bigwheels.project_cube_xr.debug --type openxr --download_path "/path/to/save/captures"
  ```
 Then you can follow the hint output to trigger a capture by press key `t` and `enter` or exit by press key `enter` only.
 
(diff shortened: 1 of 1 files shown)


capture_service/dive_client_cli.cc
```diff
@@ -582,6 +582,12 @@ bool clean_up_app_and_device(Dive::DeviceManager& mgr, const std::string& packag
                      "clean up package specific settings.";
     }
 
+    if (absl::StatusOr<Dive::AndroidDevice*> device = mgr.SelectDevice(serial); !device.ok())
```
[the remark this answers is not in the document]
footballhead: Makes sense, good suggestion. Applied

capture_service/dive_client_cli.cc
```diff
@@ -582,6 +582,15 @@ bool clean_up_app_and_device(Dive::DeviceManager& mgr, const std::string& packag
...
+    if (mgr.GetDevice() == nullptr)
+    {
+        if (absl::StatusOr<Dive::AndroidDevice*> device = mgr.SelectDevice(serial); !device.ok())
+        {
+            std::cout << "Failed to select device: " << device.status() << std::endl;
+            return false;
+        }
+    }
+
     return mgr.Cleanup(serial, package).ok();
```
footballhead, line 594: Thinking about this some more in the context of #167 , what are the responsibilities of DeviceManager::Cleanup? Should it clean up all ports that could be chosen--not just the `kFirstPort`--in the range`[kFirstPort..kFirstPort+kPortRange]`  at the risk of trampling any other applications? Seems odd to just clean one port when it may not correspond to a previously uncleaned session using a different port.

elviscapiaq: Actually, it finds the first available port in range `[kFirstPort, kFirstPort + kPortRange]`. That means if some port is being used we don't touch it. Then, in the cleaning stage, we remove that available port. Note that we only used and clean one single port.

elviscapiaq (approved):

capture_service/dive_client_cli.cc
footballhead: Thanks for the clarification!

PR #194, opened 2025-04-02
Updates minimum CMake values

GrantComm:
Updates minimum CMake values in the dive CMake files. This change also includes the - DCMAKE_POLICY_VERSION_MINIMUM=3.5 flag for windows and android builds. This flag is needed since grpc has a c-ares submodule that uses cmake version 3.1.0 which is not supported. The latest release version of grpc does not currently have an updated version of c-ares.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -100,7 +100,7 @@ jobs:
           which clang-14
       - name: Build the UI with gcc-14
         run: |
-          mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER:FILEPATH=/usr/bin/gcc-14 -DCMAKE_CXX_COMPILER:FILEPATH=/usr/bin/g++-14 .. -GNinja && ninja && cd ..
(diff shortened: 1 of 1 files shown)


wangra-google (approved):

footballhead (approved):
My understanding is that minimum 3.10 and `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` are recommendations made by cmake output.

Do we need to do any testing outside of making sure it builds? I think we would be warned about policy differences in the first place, but I don't recall seeing anything.

GrantComm:
> My understanding is that minimum 3.10 and `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` are recommendations made by cmake output for the failing builds.
> 
> Do we need to do any testing outside of making sure it builds? I think we would be warned about policy differences in the first place, but I don't recall seeing anything.

I am not sure about other specific testing needed. I just built locally and made sure I was able to launch the UI and start the cli. I did see some warnings about some of the other third party libraries related to versions not being supported in the future. Maybe those should be updated in follow up prs as well.

PR #197, opened 2025-04-22
Updating the link for Qt5 offline installers at README

elviscapiaq:
- The current link for online installers is only now for Qt6.
- Note that since Qt 5.15 no more offline installers are provided.
- If you want to install Qt 5.15 or a newer version, you have to build it from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ git submodule update --init --recursive
 
 ### Prerequisite
 
- - The QT framework, can be installed from [QT online installer](https://download.qt.io/archive/online_installers/4.6/). We are currently using QT 5.11.2
+ - The QT framework, can be installed from [QT online installer](https://download.qt.io/archive/online_installers/4.6/). We are currently using QT 5.15.2. Note that to install QT 5.15.2 from the online installer, you have to enable (turn on) the `archived` versions and then click on `filter`.
  - gRPC [dependencies](https://github.com/grpc/grpc/blob/master/BUILDING.md#pre-requisites)
  - Android NDK (currently we are using 25.2.9519653). Set the `ANDROID_NDK_HOME` environment variable.
   ```


README.md
```diff
@@ -11,7 +11,7 @@ git submodule update --init --recursive
 
 ### Prerequisite
 
- - The QT framework, can be installed from [QT online installer](https://download.qt.io/archive/online_installers/4.6/). We are currently using QT 5.11.2
+ - The QT framework, can be installed from [QT offline installer](https://download.qt.io/new_archive/qt/). We are currently using QT 5.11.2. Note that offline installers are not available any more since Qt 5.15. If you want to install [Qt 5.15](https://download.qt.io/archive/qt/) or a newer version then you have to build it from source.
```
wangra-google, line 14: i am wondering if we could just upload 5.15.2 into dive proj so that people dont need to bother manually build. @RenfengLiu what do you think?

footballhead, line 14: Do we need the offline installer? The online installer has 5.15 if you turn on "archived" versions.

elviscapiaq: You are right. Didn't notice that enabling "archived" and then clicking on "filter" allows you to install Qt 5.15.
Also, note that only Qt 5.15 is available but not older versions like Qt5.11. Thank you so much Mike for this observation.

elviscapiaq: Based on MIke's observatiob, there is a way to install Qt 5.15 directly from the online installer.
So, I think we can use this version since it is more stable than older ones.
I updated the README to use the version 5.15 via the online installer leaving comments how to get this version.
Regarding Renderdoc, I took a quick look and it seems they include the libraries in its source, however, I think we can create a bug for this potential alternative and then have a different code change for it. Let me know what you thinlk.

wangra-google (approved):

PR #199, opened 2025-04-25
Minor fix for initialization

RenfengLiu:
The m_gfxr_enabled is not initialized, and when using UI the behavior is different sometimes.

### DIFF
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -56,7 +56,7 @@ AndroidDevice::~AndroidDevice()
 {
     if (!m_serial.empty())
     {
-        CleanupDevice().IgnoreError();
+        CleanupDevice();
     }
     LOGD("AndroidDevice destroyed.\n");
 }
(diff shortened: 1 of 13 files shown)


capture_service/device_mgr.cc
```diff
@@ -47,6 +47,7 @@ std::string DeviceInfo::GetDisplayName() const
 AndroidDevice::AndroidDevice(const std::string &serial) :
     m_serial(serial),
     m_adb(serial),
+    m_gfxr_enabled(false),
```
footballhead, line 50: Should we prefer setting this in the header instead? This and future constructors would benefit from it.

footballhead (approved):

GrantComm (approved):

capture_service/device_mgr.cc
footballhead: If we look at https://google.github.io/styleguide/cppguide.html I don't see any arguments for or against default member initializers in headers. The only main rule re what is in/out of headers is for functions.

footballhead (approved):

capture_service/device_mgr.cc
footballhead, line 50: nit: you technically don't need these anymore but the redundancy won't hurt anything. We can have the defaults then still be explicit for each particular constructor about the behavior we desire.

It might impact future refactoring if we want to change the default member initialization as well as the constructors but forget to update the constructors (because we assume that the default is used). But that should be easier to catch than undefined behavior.

PR #200, opened 2025-04-25
Prevent gRPC from crash

RenfengLiu:
This starts the gRPC server after the instances have been created. This is to make sure the libraries are loaded before running gRPC server. And it can be removed after we transit to our own socket implementation.

This is to fix b/413642671

the reason behind the crash is that the global static variable defined in gRPC (such as third_party/grpc/src/core/lib/iomgr/timer_manager.cc)  now has undefined initialization order compared to `        [[maybe_unused]] static struct InitServer init_server; ` in layer/layer_common.cc. 

This PR checks if instances have been created or not, because by that time, all global static variable must has been initialized, thus can prevent grpc from crash.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ git submodule update --init --recursive
 
 ### Prerequisite
 
- - The QT framework, can be installed from [QT online installer](https://download.qt.io/archive/online_installers/4.6/). We are currently using QT 5.11.2
+ - The QT framework, can be installed from [QT online installer](https://download.qt.io/archive/online_installers/4.6/). We are currently using QT 5.15.2. Note that to install QT 5.15.2 from the online installer, you have to enable (turn on) the `archived` versions and then click on `filter`.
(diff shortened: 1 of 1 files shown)


layer/layer_common.cc
```diff
@@ -80,7 +108,7 @@ void PreventLibraryUnload()
     Dl_info info;
     if (dladdr(reinterpret_cast<void *>(&PreventLibraryUnload), &info))
     {
-        dlopen(info.dli_fname, RTLD_NOW | RTLD_NOLOAD | RTLD_LOCAL | RTLD_NODELETE);
+        dlopen(info.dli_fname, RTLD_NOLOAD | RTLD_NODELETE);
```
wangra-google, line 111: just curious: why do you remove RTLD_NOW?

elviscapiaq: What about RTLD_LOCAL being removed as well? Looks like without those flags it is less restrictive and it doesn't matter if the symbols are resolved?

wangra-google: good to know, thanks

wangra-google (approved):

PR #201, opened 2025-04-26
Fixes the issue with threads in Dive UI for Windows

elviscapiaq:
- To avoid sending events to an object created in another thread, a signal-slot mechanism was used to indirectly update the progress bar value.


MichaelAnttila (changes requested):
I believe you have to use the signal/slot mechanism for show() as well.

footballhead (approved):

MichaelAnttila (approved):

PR #203, opened 2025-05-01
Add the option to the runtime layer to disable timestamp

wangra-google:


wangra-google:
> So this is do disable the applications timestamp query? in that case, will that cause the vkGetQueryPoolResults waiting forever in cases of VK_QUERY_RESULT_WAIT_BIT is enabled?

nice catch, added vkGetQueryPoolResults support

PR #204, opened 2025-05-07
Initialize GFXR capture trigger property

footballhead:
`debug.gfxrecon.capture_android_trigger` is not set on the first launch of Dive. This causes GFXR to not use triggered trim captures: attempts to set capture_android_trigger to trigger the capture are ignored.

The GFXR docs (USAGE_android.md) suggest that if you want trimmed capture mode then you need to set the property to something prior to startup.

Given the observed behavior, always set the property to `false` on setup. Additionally, clear the property on exit so that other apps which may want to use GFXR with a full capture don't accidentally inherit the setting.

### DIFF
--- a/capture_service/android_application.cc
+++ b/capture_service/android_application.cc
@@ -374,7 +374,7 @@ absl::Status OpenXRApplication::Cleanup()
         RETURN_IF_ERROR(
         m_dev.Adb().Run(absl::StrFormat("shell setprop wrap.%s \\\"\\\"", m_package)));
     }
-    LOGD("OpenXRApplication %s cleanup done.", m_package.c_str());
+    LOGD("OpenXRApplication %s cleanup done.\n", m_package.c_str());
     return absl::OkStatus();
 }
 


GrantComm:
debug.gfxrecon.capture_android_trigger is set to true during runtime when the user starts a capture and false on the retrieval of the capture file both in the cli and ui. Were you seeing non-trimmed captures when inspecting them? There should only be 1 frame captured when inspecting the retrieved capture file with gfxrecon-info

GrantComm (approved):

footballhead:
> debug.gfxrecon.capture_android_trigger is set to true during runtime when the user starts a capture and false on the retrieval of the capture file both in the cli and ui. Were you seeing non-trimmed captures when inspecting them? There should only be 1 frame captured when inspecting the retrieved capture file with gfxrecon-info

It was behaving like it was taking a non-trimmed capture. File name was `com.google.bigwheels.project_cube_xr.debug_20250506T182101.gfxr`. I wasn't seeing lines in the logcat acknowledging the change to capture_android_trigger. I pulled the capture but it claims to have 0 frames. I can't convert it because it errors with "Incomplete block at end of file". My theory is that, since the full capture takes everything between vkCreateInstance and vkDestroyInstance, and the app was still running, the capture was still going when I pulled it.

GrantComm:
> > debug.gfxrecon.capture_android_trigger is set to true during runtime when the user starts a capture and false on the retrieval of the capture file both in the cli and ui. Were you seeing non-trimmed captures when inspecting them? There should only be 1 frame captured when inspecting the retrieved capture file with gfxrecon-info
> 
> It was behaving like it was taking a non-trimmed capture. File name was `com.google.bigwheels.project_cube_xr.debug_20250506T182101.gfxr`. I wasn't seeing lines in the logcat acknowledging the change to capture_android_trigger. I pulled the capture but it claims to have 0 frames. I can't convert it because it errors with "Incomplete block at end of file". My theory is that, since the full capture takes everything between vkCreateInstance and vkDestroyInstance, and the app was still running, the capture was still going when I pulled it.

Ahh ok

PR #206, opened 2025-05-07
Add the support of measuring gpu time for command buffer to RT layer

wangra-google:
The runtime layer can be installed through
script/install_runtime_vklayer.bat 

Current implementation assumes that there is no overlapping of the works from different command buffers which is not valid if async compute is used. 

Also note that the result from this vk layer is slightly different from the value got from perfetto kgsl event. We are contacting qc to understand why there is a difference.

### DIFF
--- a/capture_service/android_application.cc
+++ b/capture_service/android_application.cc
@@ -220,6 +220,8 @@ absl::Status VulkanApplication::Cleanup()
         RETURN_IF_ERROR(
         m_dev.Adb().Run("shell setprop debug.gfxrecon.capture_use_asset_file false"));
         RETURN_IF_ERROR(
+        m_dev.Adb().Run("shell setprop debug.gfxrecon.capture_android_trigger \\\"\\\""));
+        RETURN_IF_ERROR(
         m_dev.Adb().Run(absl::StrFormat("shell run-as %s rm %s", m_package, kVkGfxrLayerLibName)));
     }
     else
@@ -287,6 +289,9 @@ absl::Status AndroidApplication::GfxrSetup()
     RETURN_IF_ERROR(m_dev.Adb().Run("shell setprop debug.gfxrecon.capture_trigger_frames 1"));
 
     RETURN_IF_ERROR(m_dev.Adb().Run("shell setprop debug.gfxrecon.capture_use_asset_file true"));
+
+    // capture_android_trigger must be set in order for GFXR to listen for triggers.
+    RETURN_IF_ERROR(m_dev.Adb().Run("shell setprop debug.gfxrecon.capture_android_trigger false"));
 
     LOGD("GFXR capture setup for %s done\n", m_package.c_str());
     return absl::OkStatus();
@@ -359,6 +364,8 @@ absl::Status OpenXRApplication::Cleanup()
         RETURN_IF_ERROR(
         m_dev.Adb().Run("shell setprop debug.gfxrecon.capture_use_asset_file false"));
         RETURN_IF_ERROR(
+        m_dev.Adb().Run("shell setprop debug.gfxrecon.capture_android_trigger \\\"\\\""));
+        RETURN_IF_ERROR(
         m_dev.Adb().Run(absl::StrFormat("shell run-as %s rm %s", m_package, kVkGfxrLayerLibName)));
     }
     else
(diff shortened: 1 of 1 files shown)


wangra-google:
> as we are intercept more and more calls, maybe it's a good idea to considering using the vkspecgen to generate those boilerplate code

definitely. for now it is ok to do this manually since it is still manageable, but definitely need to auto generate those code if we need to intercept more function.

runtime_layer/vk_rt_layer_impl.cc
```diff
@@ -25,71 +25,167 @@ limitations under the License.
...
+#include <cmath>
+#include <cstring>
 
 namespace DiveLayer
 {
 
 static bool sEnableDrawcallReport = false;
 static bool sEnableDrawcallLimit = false;
 static bool sEnableDrawcallFilter = false;
+// To use sEnableOpenXRGPUTiming, make sure to
```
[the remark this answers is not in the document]
wangra-google: I am waiting for Elvis change so that we could control those options from Dive UI.

runtime_layer/vk_rt_layer_impl.cc
```diff
@@ -25,71 +25,167 @@ limitations under the License.
...
+
+    return stats;
+}
+
+void DiveRuntimeLayer::FrameMetrics::PrintStats(const FrameMetrics::Stats& stats)
+{
+    LOGI("FrameMetrics: ");
+    LOGI("  Min: %f ms ", stats.min);
+    LOGI("  Max: %f ms ", stats.max);
+    LOGI("  Mean: %f ms ", stats.average);
```
shanminchao, line 92: Wouldn't median be useful too? In fact, I'd argue that median is more useful than mean for frame time analysis, no?

runtime_layer/vk_rt_layer_impl.cc
```diff
@@ -150,24 +246,406 @@ VkResult GetQueryPoolResults(PFN_vkGetQueryPoolResults pfn,
...
+{
+    pfn(commandBuffer, pLabelInfo);
+
+    if (strcmp("vr-marker,frame_end,type,application", pLabelInfo->pLabelName) == 0)
+    {
+        if (m_cmds.find(commandBuffer) == m_cmds.end())
+        {
+            LOGE("%p is not in the cmd cache!", commandBuffer);
+        }
+        m_cmds[commandBuffer].is_frameboundary = true;
```
shanminchao, line 647: This is very specific to a specific marker inserted for XR use cases, right? Don't we want to generalize this to support non-XR apps?

runtime_layer/vk_rt_layer_impl.h
```diff
@@ -18,54 +18,184 @@ limitations under the License.
...
+
+        Stats GetStatistics() const;
+
+        void PrintStats(const Stats& stats);
+
+    private:
+        double CalculateAverage() const;
+        double CalculateStdDev(double average) const;
+
+        // Store frame data as a vector of vectors of doubles
```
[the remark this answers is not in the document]
wangra-google: good catch. removed the comments

runtime_layer/vk_rt_layer_impl.cc
```diff
@@ -150,24 +246,406 @@ VkResult GetQueryPoolResults(PFN_vkGetQueryPoolResults pfn,
...
+VkResult DiveRuntimeLayer::ResetCommandBuffer(PFN_vkResetCommandBuffer  pfn,
+                                              VkCommandBuffer           commandBuffer,
+                                              VkCommandBufferResetFlags flags)
+{
+    if (m_cmds.find(commandBuffer) == m_cmds.end())
+    {
+        LOGE("%p is not in the cmd cache!", commandBuffer);
+    }
+    m_cmds[commandBuffer].timestamp_offset = CommandBufferInfo::kInvalidTimeStampOffset;
+    m_cmds[commandBuffer].is_frameboundary = false;
```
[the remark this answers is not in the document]
wangra-google: indeed, added reset for that flag

runtime_layer/vk_rt_layer_impl.cc
wangra-google: this is specific for xr for now, (that is why i named the bool to be `sEnableOpenXRGPUTiming`). It should be simple to make it more generic. Just want to make sure this one works properly (consistent with kgsl event, waiting for QC to explain why there is a difference.) before making is more generic.

runtime_layer/vk_rt_layer_impl.cc
wangra-google: good point, added it into the median support

PR #198, opened 2025-04-24
More descriptive logging and refactoring DeviceManager::Cleanup

angela28chen:
# Changes
* Cleanup
  * Removed redundancy between `DeviceManager::Cleanup`, `VulkanApplication::Cleanup` and `AndroidDevice::CleanupDevice`
  * Tweaked the cleanup commands so that all of them will return 0 during normal operation (There were errors being shown for unimportant things like deleting a non-existent file, etc.)
* Logging in general
  * Removed "result->m_ret is 0"
  * Log descriptive error if encountered (will contain the return value). Before, the error was suppressed and I think this has lead to some bad silent failures.
  * Set the default of `RunCommand` to quiet=false, so that in debug mode most commands will be shown. I think there's a lot of redundancy in our CLI code that might be flying under the radar because of the quiet commands. For example the cleanup code before this PR, and also you can see below that `ro.product.manufacturer` and `ro.product.model` are queried multiple times.
* Verified working on Windows and Linux host 

# DeviceManager::Cleanup Before
```
(dive) C:\src\git\dive>build\bin\Debug\dive_client_cli.exe --device 9A231FFAZ0092E  --command cleanup --package com.google.bigwheels.project_fishtornado.debug
result->m_ret is 0
Command: adb devices
 Output: List of devices attached
9A231FFAZ0092E  device
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
enforce: Enforcing
select: Google Pixel 4 (9A231FFAZ0092E)
AndroidDevice created.
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 1
result->m_ret is 1
result->m_ret is 1
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
result->m_ret is 0
Cleanup device 9A231FFAZ0092E
restore Enforcing to Enforcing
result->m_ret is 0
result->m_ret is 1
Cleanup device 9A231FFAZ0092E done
AndroidDevice destroyed.
```

# DeviceManager::Cleanup After
```
(dive) C:\src\git\dive>build\bin\Debug\dive_client_cli.exe --device 9A231FFAZ0092E  --command cleanup --package com.google.bigwheels.project_fishtornado.debug
> adb devices
List of devices attached
[...]

PR #207, opened 2025-05-09
Fix nodiscard warnings by ignoring returns

footballhead:
absl::Status is marked as nodiscard via ABSL_MUST_USE_RESULT. The compiler will warn if it's not consumed. This is the case for some functions in AndroidDevice after PR #198. An example:

```
/usr/local/google/home/hitchens/git/dive/capture_service/device_mgr.cc: In member function ‘absl::lts_20230802::Status Dive::AndroidDevice::CleanupDevice()’:
/usr/local/google/home/hitchens/git/dive/capture_service/device_mgr.cc:250:22: warning: ignoring returned value of type ‘absl::lts_20230802::Status’, declared with attribute ‘nodiscard’ [-Wunused-result]
  250 |             Adb().Run(absl::StrFormat("forward --remove tcp:%d", Port()));
      |             ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/local/google/home/hitchens/git/dive/capture_service/device_mgr.h:22,
                 from /usr/local/google/home/hitchens/git/dive/capture_service/device_mgr.cc:17:
/usr/local/google/home/hitchens/git/dive/capture_service/command_utils.h:85:25: note: in call to ‘absl::lts_20230802::Status Dive::AdbSession::Run(const std::string&) const’, declared here
   85 |     inline absl::Status Run(const std::string &command) const
      |                         ^~~
In file included from /usr/local/google/home/hitchens/git/dive/capture_service/device_mgr.h:19:
/usr/local/google/home/hitchens/git/dive/third_party/grpc/third_party/abseil-cpp/absl/status/status.h:424:7: note: ‘absl::lts_20230802::Status’ declared here
  424 | class Status final {
      |       ^~~~~~
```

The intent in #198 was that some commands that fail should not propagate and halt the program. In this case, we don't care about the return result and should ignore it. absl::Status provides the IgnoreError() method for this purpose.

### DIFF
--- a/runtime_layer/vk_rt_dispatch.cc
+++ b/runtime_layer/vk_rt_dispatch.cc
@@ -45,8 +45,20 @@ void InitDeviceDispatchTable(VkDevice device, PFN_vkGetDeviceProcAddr pa, Device
     dt->CmdResetQueryPool = (PFN_vkCmdResetQueryPool)pa(device, "vkCmdResetQueryPool");
     dt->CmdWriteTimestamp = (PFN_vkCmdWriteTimestamp)pa(device, "vkCmdWriteTimestamp");
     dt->GetQueryPoolResults = (PFN_vkGetQueryPoolResults)pa(device, "vkGetQueryPoolResults");
+    dt->DestroyCommandPool = (PFN_vkDestroyCommandPool)pa(device, "vkDestroyCommandPool");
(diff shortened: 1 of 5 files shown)


elviscapiaq (approved):
Thanks for catching it.

PR #208, opened 2025-05-09
Fix some PM4 bugs

wangra-google:
- Fix the case when some PM4s share the same value but with different variants (CP_THREAD_CONTROL (A7XX-) and IN_IB_PREFETCH_END (A2XX) both use 0x17)
- Include current package into scrollto range for markers (note the logic between markers and drawcalls are different: for markers, we want to show all the PM4s from the one that set current marker to the one before next marker, whereas for drawcalls, we want to show all PM4s from the one after previous drawcall/marker to the one that issues current drawcall)

### DIFF
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -56,7 +56,7 @@ AndroidDevice::~AndroidDevice()
 {
     if (!m_serial.empty())
     {
-        CleanupDevice();
+        CleanupDevice().IgnoreError();
     }
     LOGD("AndroidDevice destroyed.\n");
 }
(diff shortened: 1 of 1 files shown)


dive_core/generatePm4Info_adreno.py
```diff
@@ -823,6 +825,11 @@ def outputPacketInfo(pm4_info_file, registers_et_root, enum_index_dict, opcode_d
       # Keep track of instance #. Only the 1st instance belongs in the vector. The rest are in the multimap.
       if opcode not in packet_type_instances:
         packet_type_instances[opcode] = 1
+        if (opcode == 0):
+          pm4_info_file.write("    // For descriptors, we purposefully try to include them as \"packets\" for easier parsing.\n")
```
footballhead, line 829: PEP 8 encourages using single quotes here to avoid backslashes (https://peps.python.org/pep-0008/#string-quotes).

```suggestion
          pm4_info_file.write('    // For descriptors, we purposefully try to include them as "packets" for easier parsing.\n')
```

dive_core/generatePm4Info_adreno.py
```diff
@@ -823,6 +825,11 @@ def outputPacketInfo(pm4_info_file, registers_et_root, enum_index_dict, opcode_d
...
+          pm4_info_file.write("    // They're not technically PM4 packets, hence the 0x0.\n")
+          pm4_info_file.write("    // It's used: \"const PacketInfo *packet_info_ptr = GetPacketInfo(0, sharp_struct_name);\".\n")
```
footballhead, line 831: PEP 8 encourages using single quotes here to avoid backslashes (https://peps.python.org/pep-0008/#string-quotes). You've got some options:

Either--given that this includes code samples--you could use backticks instead (like Markdown).

```suggestion
          pm4_info_file.write("    // It's used: `const PacketInfo *packet_info_ptr = GetPacketInfo(0, sharp_struct_name);`.\n")
```

Or: keep the code in double quotes and replace the contractions (to avoid backslashes on the single quote):

```suggestion
          pm4_info_file.write('    // It is used: "const PacketInfo *packet_info_ptr = GetPacketInfo(0, sharp_struct_name);".\n')
```

dive_core/generatePm4Info_adreno.py
wangra-google: Thanks for the suggestion, fixed

dive_core/generatePm4Info_adreno.py
wangra-google: Thanks for the suggestion, fixed

dive_core/generatePm4Info_adreno.py
```diff
@@ -843,8 +850,20 @@ def outputPacketInfo(pm4_info_file, registers_et_root, enum_index_dict, opcode_d
...
+      if 'variants' in pm4_type_packet_value.attrib:
+        variants = pm4_type_packet_value.attrib['variants']
+        variants_bitfield = GetGPUVariantsBitField(variants)
+        if (variants_bitfield != 0):
+          # kGPUVariantsBits has 6 bits
+          # it seems that the variant is only used for the non-domain ones
+          for i in range(6):
+            cur_variant_bitfield = (1<<i)
+            if cur_variant_bitfield & variants_bitfield:
+              pm4_info_file.write("    g_sPacketInfoVariant[(0x%x << kGPUVariantsBits) | 0x%x] = { \"%s\", 0, UINT8_MAX, {" % (opcode, cur_variant_bitfield, packet_name) + " } };\n")
```
footballhead, line 863: PEP 8 encourages using single-quoted strings here to avoid backslashes (https://peps.python.org/pep-0008/#string-quotes).
```suggestion
              pm4_info_file.write('    g_sPacketInfoVariant[(0x%x << kGPUVariantsBits) | 0x%x] = { "%s", 0, UINT8_MAX, {' % (opcode, cur_variant_bitfield, packet_name) + " } };\n")
```

dive_core/generatePm4Info_adreno.py
```diff
@@ -843,8 +850,20 @@ def outputPacketInfo(pm4_info_file, registers_et_root, enum_index_dict, opcode_d
...
+      else:
+        pm4_info_file.write("    g_sPacketInfo[0x%x] = { \"%s\", 0, UINT8_MAX, {" % (opcode, packet_name) + " } };\n")
```
footballhead, line 865: PEP 8 encourages using single-quoted strings here to avoid backslashes (https://peps.python.org/pep-0008/#string-quotes).

```suggestion
        pm4_info_file.write('    g_sPacketInfo[0x%x] = { "%s", 0, UINT8_MAX, {' % (opcode, packet_name) + " } };\n")
```

wangra-google: thanks for pointing this. I will do a clean up in a different cl to make sure all formats are consistent (not just these ones in this cl)

dive_core/generatePm4Info_adreno.py
wangra-google: thanks for pointing this. I will do a clean up in a different cl to make sure all formats are consistent (not just these ones in this cl)

dive_core/generatePm4Info_adreno.py
```diff
@@ -843,8 +850,20 @@ def outputPacketInfo(pm4_info_file, registers_et_root, enum_index_dict, opcode_d
     domain = registers_et_root.find('{http://nouveau.freedesktop.org/}domain[@name="'+packet_name+'"]')
     if domain is None:
       opcode = int(pm4_type_packet_value.attrib['value'],0)
-      pm4_info_file.write("    g_sPacketInfo[0x%x] = { \"%s\", 0, UINT8_MAX, {" % (opcode, packet_name))
-      pm4_info_file.write(" } };\n")
+
+      if 'variants' in pm4_type_packet_value.attrib:
```
shanminchao, line 854: Could you add a comment indicating some example packets of where we have "variants" and therefore need to use g_sPacketInfoVariant?

dive_core/generatePm4Info_adreno.py
```diff
@@ -963,8 +982,19 @@ def outputFunctionsCpp(pm4_info_file):
 
 const PacketInfo *GetPacketInfo(uint32_t op_code)
 {
+    // check without variant as key
     if (g_sPacketInfo[op_code].m_name == nullptr)
-        return nullptr;
+    {
+        // check with variant as key
```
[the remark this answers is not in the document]
wangra-google: sorry i am a bit confused, isn't `m_name` the same as the `variant bit`? If the op_code has a variant, g_sPacketInfo[op_code].m_name would be nullptr? the lookup would be O(1) anyways? unless you prefer to have a bitfield for all op_code?

dive_core/generatePm4Info_adreno.py
wangra-google: done

shanminchao: Thanks!

dive_core/generatePm4Info_adreno.py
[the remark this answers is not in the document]
shanminchao: You are right. You made it behave like g_sRegInfo, where if the op_code has a variant, then it doesn't exist in g_sPacketInfo[op_code] at all, and therefore that m_name would be nullptr. My mistake!

shanminchao (approved):

PR #212, opened 2025-05-15
Add basic functionality for modifying a GFXR file to divecli

angela28chen:
Considerations to build `gfxrecon_decode` into `dive_core`:
* Windows
  * Renaming `CopyMemory` to `RetrieveMemoryData` to avoid name collision with Windows macros.
  * Linking library `psapi` to all targets using library `gfxrecon_decode`.
* Linux
  * If `gfxrecon_decode` is linked with keyword `PUBLIC` then the target `dive_ui` does not build because there are many new issues with the Qt build. So instead it is linked with `PRIVATE` and the `target_include_directories` are manually specified.

DiveBlockData:
* Stores metadata and modifications for a GFXR file, allowing a new one to be written
* Usage:
1. Create an empty DiveBlockData object
2. Register it with FileProcessor, then read a GFXR file
3. Lock DiveBlockData's original file info
4. [TODO] Apply modifications
5. Use DiveBlockData to write a new GFXR file

Manual test:
* Verified that `divecli` will produce an identical GFXR file on Windows and on Linux

### DIFF
--- a/third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
+++ b/third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
@@ -63,7 +63,9 @@ bool DiveBlockData::FinalizeOriginalBlocksMapSizes()
         original_blocks_map_[i].size = size;
     }
 
-    // Removing the last "block" which was not a real block, just the marker added at the end of the file
+    // The file processor calls AddOriginalBlock() even at the very end of the GFXR file, so this last block has a size
+    // of 0 and its offset is equal to the file size. The info was used in the calculation of the size of the
+    // penultimate block and now the last block needs to be trimmed.
     original_blocks_map_.pop_back();
 
     original_blocks_map_locked_ = true;
@@ -175,7 +177,8 @@ bool DiveBlockData::WriteGFXRFile(const std::string& original_file_path, const s
 
     if (max_block_size > kDiveBlockBufferSize)
     {
-        GFXRECON_LOG_WARNING("kDiveBlockBufferSize (%d) is too small to accommodate max block size (%d)",
+        GFXRECON_LOG_WARNING("kDiveBlockBufferSize (%d) is too small to accommodate max block size (%d) and will cause "
+                             "more copy operations in CopyBlockBetweenFiles",
                              kDiveBlockBufferSize,
                              max_block_size);
     }
--- a/dive_core/capture_event_info.cpp
+++ b/dive_core/capture_event_info.cpp
@@ -34,7 +34,7 @@ SyncType GetSyncType(const IMemoryManager &mem_manager,
     if (opcode == CP_EVENT_WRITE7)
     {
         PM4_CP_EVENT_WRITE packet;
-        DIVE_VERIFY(mem_manager.CopyMemory(&packet, submit_index, addr, sizeof(packet)));
+        DIVE_VERIFY(mem_manager.RetrieveMemoryData(&packet, submit_index, addr, sizeof(packet)));
         SyncType type = (SyncType)packet.bitfields0.EVENT;
         return type;
     }


angela28chen:
Points of discussion:

1. Is it a good idea to be renaming `CopyMemory` or should we have a header file full of Windows undefines for Dive? Personally I think renaming is OK.
2. Is there a reason that `psapi` cannot be built into the `gfxrecon_decode` library? I see that it's not done like that in the upstream repo, but conceptually I don't understand why.

third_party/gfxreconstruct/framework/decode/file_processor.cpp
```diff
@@ -190,6 +190,12 @@ bool FileProcessor::ContinueDecoding()
     }
     else
     {
+        if (run_without_decoders_)
```
GrantComm, line 193: Can you add " // GOOGLE:" here

third_party/gfxreconstruct/framework/decode/file_processor.cpp
```diff
@@ -481,6 +487,8 @@ bool FileProcessor::ReadBlockHeader(format::BlockHeader* block_header)
 {
     assert(block_header != nullptr);
 
+    StoreBlockInfo(); // Before any bytes are read
```
GrantComm, line 490: Can you add " // GOOGLE:" here as well

third_party/gfxreconstruct/framework/decode/file_processor.h
```diff
@@ -159,6 +159,8 @@ class FileProcessor
 
     void PrintBlockInfo() const;
 
+    virtual void StoreBlockInfo() {}
```
GrantComm, line 162: Can you add " // GOOGLE:" here as well

GrantComm (changes requested):

GrantComm:
> Thanks again to @GrantComm and @footballhead for the help with getting this to build!
> 
> Points of discussion:
> 
> 1. Is it a good idea to be renaming `CopyMemory` or should we have a header file full of Windows undefines for Dive? Personally I think renaming is OK.
> 2. Is there a reason that `psapi` cannot be built into the `gfxrecon_decode` library? I see that it's not done like that in the upstream repo, but conceptually I don't understand why.
> 3. What conflict does building Qt on Linux have with linking `gfxrecon_decode` publicly?
> 
> If you have any opinion on these please discuss.

I also think the renaming makes sense. As far as building Qt on Linux and linking gfxrecon_decode, I think there is some collision in a defines somewhere, which is why the changes in the cmake files are necessary

dive_core/capture_data.cpp
```diff
@@ -348,10 +349,10 @@ const MemoryAllocationInfo &MemoryManager::GetMemoryAllocationInfo() const
 }
 
 //--------------------------------------------------------------------------------------------------
-bool MemoryManager::CopyMemory(void    *buffer_ptr,
-                               uint32_t submit_index,
-                               uint64_t va_addr,
-                               uint64_t size) const
+bool MemoryManager::CopyMem(void    *buffer_ptr,
```
wangra-google, line 352: I feel that this func copies data from managed memory blocks to a provided buffer, with checks for submit indices and virtual addresses. Maybe `RetrieveMemoryData` is better? It suggests that data is being fetched from a managed source. (When i see CopyMemory, my first impression is to look for source and destination which are not in the args)

lrz_validator/CMakeLists.txt
```diff
@@ -43,11 +43,13 @@ list(FILTER LIB_SRC_FILES EXCLUDE REGEX "main.cpp")
 
 add_executable(${PROJECT_NAME} ${HDR_FILES} "main.cpp")
 target_link_libraries(${PROJECT_NAME} PRIVATE dive_core)
+target_include_directories(${PROJECT_NAME} PRIVATE ${THIRDPARTY_DIRECTORY}/gfxreconstruct/framework)
 
 if (MSVC)
   # 4100: unreferenced formal parameter
   # 4201: prevent nameless struct/union
   target_compile_options(${PROJECT_NAME} PRIVATE /W4 /WX /wd4100 /wd4201)
+  target_link_libraries(${PROJECT_NAME} PRIVATE psapi)
```
wangra-google, line 52: just curious, why do you need `psapi`?

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+    {
+        GFXRECON_LOG_ERROR("Original block map is empty");
+        return false;
+    }
+
+    // Calculating block sizes
+    original_header_size_bytes_ = original_blocks_map_[0].offset;
+    for (int i = 0; i < original_blocks_map_.size() - 1; i++)
+    {
+        uint64_t size                = original_blocks_map_.at(i + 1).offset - original_blocks_map_.at(i).offset;
```
wangra-google, line 63: is there any specific reason that you want to use `.at`? from what i understand, the only benefit of using it is to have extra boundary check with a cost of performance. in this case, the `i` will never be out of bound?

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+            "Unexpected block id mismatch with index: %d, expected index: %d", index, original_blocks_map_.size());
+        return false;
+    }
+
+    original_blocks_map_.push_back(BlockBytesLocation{ offset, 0 });
+
+    return true;
+}
+
+bool DiveBlockData::LockOriginalBlocksMap()
```
wangra-google, line 45: The function is doing more than the name indicates which might be confusing. 
- lock blocks map
- update blocks map size
- removing the last block

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+    // Calculating block sizes
+    original_header_size_bytes_ = original_blocks_map_[0].offset;
+    for (int i = 0; i < original_blocks_map_.size() - 1; i++)
+    {
+        uint64_t size                = original_blocks_map_.at(i + 1).offset - original_blocks_map_.at(i).offset;
+        original_blocks_map_[i].size = size;
+    }
+
+    // Removing the last "block" which is the end of the file
+    original_blocks_map_.pop_back();
```
wangra-google, line 68: why do we need to remove the last block?

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    original_blocks_map_locked_ = true;
+    return true;
+}
+
+bool DiveBlockData::GetNewBlocksOrder()
```
wangra-google, line 74: i feel this func is UpdateNewBlockOrder (purely based on the name, i would assume GetNewBlocksOrder() wont modify any member and returns the `new_blocks_order_`)

third_party/gfxreconstruct/framework/decode/dive_block_data.h
```diff
@@ -0,0 +1,75 @@
...
+
+#ifndef GFXRECON_DECODE_DIVE_BLOCK_DATA_H
+#define GFXRECON_DECODE_DIVE_BLOCK_DATA_H
+
+#include "util/defines.h"
+
+#include <string>
+#include <vector>
+
+static const unsigned int BUFFER_SIZE = 4096;
```
wangra-google, line 28: i could be wrong, but i did a quick search, it seems that gfxr prefer using `static constexpr ` for consts with `k` as prefix. Also `buffer size` seems to be a bit too generic especially in a header. maybe something like kDiveBlockBufferSize

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+#include "decode/dive_block_data.h"
+#include "util/logging.h"
+
+#include <fstream>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+bool DiveBlockData::AddOriginalBlock(uint32_t index, uint64_t offset)
```
wangra-google, line 25: maybe better to change `index` to type `size_t` since it is used to compare against vecter::size()

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+    original_blocks_map_.push_back(BlockBytesLocation{ offset, 0 });
+
+    return true;
+}
+
+bool DiveBlockData::LockOriginalBlocksMap()
+{
+    if (original_blocks_map_locked_)
+    {
+        GFXRECON_LOG_INFO("Original block map already locked");
```
wangra-google, line 49: i feel this log seems to be a bit redundant unless you want to give some warning here that something might goes wrong in this case?

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+        return true;
+    }
+
+    if (original_blocks_map_.size() == 0)
```
wangra-google, line 53: better with `.empty()`?

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+    {
+        GFXRECON_LOG_ERROR("Original block map is empty");
+        return false;
+    }
+
+    // Calculating block sizes
+    original_header_size_bytes_ = original_blocks_map_[0].offset;
+    for (int i = 0; i < original_blocks_map_.size() - 1; i++)
```
wangra-google, line 61: nit: use `size_t` for the loop

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+    FILE* new_fd;
+    result = util::platform::FileOpen(&new_fd, new_file_path.c_str(), "wb");
+    if (result || new_fd == nullptr)
+    {
+        GFXRECON_LOG_ERROR("Failed to open file %s", new_file_path.c_str());
+        return false;
+    }
+
+    // Copy the original header
+    util::platform::FileRead(block_buffer_, original_header_size_bytes_, original_fd);
```
wangra-google, line 122: is there a possibility that `BUFFER_SIZE` is smaller than `original_header_size_bytes_`? maybe better with an assert?

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    bool res = GetNewBlocksOrder();
+    if (!res)
+    {
+        GFXRECON_LOG_ERROR("Cannot determine new blocks order");
+        return false;
+    }
+
+    FILE* original_fd;
+    int   result = util::platform::FileOpen(&original_fd, original_file_path.c_str(), "rb");
```
wangra-google, line 106: do we need to FileClose `original_fd` and `new_fd`?

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
wangra-google: you can keep the bool, but maybe better with some comments with `TODO()` on why it is bool

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,160 @@
...
+        NewBlockMetadata block;
+        block.is_original = true;
+        block.id          = primary_index;
+        new_blocks_order_.push_back(block);
+    }
+
+    return true;
+}
+
+bool DiveBlockData::WriteGFXRFile(const std::string& original_file_path, const std::string& new_file_path)
```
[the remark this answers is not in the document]
wangra-google: thanks

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
wangra-google: Sorry for being a bit picky, but I feel `FinalizeOriginalBlocMapSizes` might be better. the idea here is to allow the person who doesn't know the code at all quickly guessing what the func does without looking at the definition. :)

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
wangra-google: sorry i don't see the new change on this. is it possible that the latest version is not committed correct?

wangra-google: thanks

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
angela28chen: The file processor calls "AddOriginalBlock" even at the very end of the file before deciding to exit the loop, so the very last "block" ends up with an offset that is equal to the size of the file in bytes. This line drops that "block". It's still useful to know the offset to calculate the size of the block n-1

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
angela28chen: No worries, renamed to `FinalizeOriginalBlocksMapSizes()`

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
angela28chen: OK, I added to the TODO on line 77

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+            if (!CopyBlockBetweenFiles(current_block_location.size, original_fd, new_fd))
+            {
+                GFXRECON_LOG_ERROR("Could not copy original block id: %d", current_block_metadata.id);
+                return false;
+            }
+        }
+    }
+
+    if (max_block_size > kDiveBlockBufferSize)
```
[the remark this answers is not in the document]
wangra-google: make sense, can you add this info in the warning message? explaining that this will cause more copy operations in `CopyBlockBetweenFiles`

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
wangra-google: good to know, thanks. can you add this in the comment? thanks :)

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
wangra-google: thanks

lrz_validator/CMakeLists.txt
GrantComm: It is used with Windows builds in [driver_info.cpp](https://github.com/LunarG/gfxreconstruct/blob/406aa68e9881e75f14cc88da1b972999bda4c412/framework/util/driver_info.cpp#L37) and [file_path.cpp](https://github.com/LunarG/gfxreconstruct/blob/406aa68e9881e75f14cc88da1b972999bda4c412/framework/util/file_path.cpp#L34)

GrantComm: > It is used with Windows builds in [driver_info.cpp](https://github.com/LunarG/gfxreconstruct/blob/406aa68e9881e75f14cc88da1b972999bda4c412/framework/util/driver_info.cpp#L37) and [file_path.cpp](https://github.com/LunarG/gfxreconstruct/blob/406aa68e9881e75f14cc88da1b972999bda4c412/framework/util/file_path.cpp#L34)

I believe I linked it to get the dive windows build to work with gfxr

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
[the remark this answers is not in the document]
angela28chen: Done

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
angela28chen: Done

wangra-google (approved):

GrantComm (approved):

PR #213, opened 2025-05-15
Cleanup unused topologies and old vulkan parsing code

shanminchao:


ui/main_window.cpp
```diff
@@ -59,14 +59,10 @@
...
-static const char *kViewModeStrings[kViewModeStringCount] = { "Engine", "Submit", "Events" };
-static const char *kEventViewModeStrings[kEventViewModeStringCount] = {
-    "Vulkan Events",
-    "All Vulkan Calls",
-    "All Vulkan Calls + GPU Events"
-};
+static const int   kViewModeStringCount = 2;
+static const int   kEventViewModeStringCount = 1;
+static const char *kViewModeStrings[kViewModeStringCount] = { "Submit", "Events" };
+static const char *kEventViewModeStrings[kEventViewModeStringCount] = { "GPU Events" };
```
wangra-google, line 65: do we still need this to be an array if the count is 1? or do you think there might be new mode in the future?

wangra-google (approved):

GrantComm (approved):

ui/main_window.cpp
shanminchao: I suspect Grant may want to add his Vulkan topologies in there, so I didn't remove that array

PR #224, opened 2025-05-30
Adding a modification framework to DiveBlockData

angela28chen:
Refactoring DiveBlockData and adding functionality to store and apply block-level modifications.
- The location of a modification is described with `primary_index` and `secondary_index`, describing which original block it's anchored to and how it is positioned in relation to the anchor block (negative for immediately before, 0 for overwriting, and positive for immediately afterwards)
- Original block metadata and modification block metadata are stored separately so that modifications can be manipulated easily without affecting the original data.
- `BlockVisitor` is defined using the Visitor design pattern, with `WriterBlockVisitor` made to hold handles to open files (original and new gfxr files) and write the new file block-by-block in the traversal order

### DIFF
--- a/gfxr_ext/decode/dive_block_data.cpp
+++ b/gfxr_ext/decode/dive_block_data.cpp
@@ -242,7 +242,6 @@ bool DiveBlockData::TraverseBlocks(BlockVisitor& visitor) const
     // Go through block-by-block in order of primary_id
     for (uint32_t primary_id = 0; primary_id < original_blocks_map_.size(); primary_id++)
     {
-        GFXRECON_LOG_INFO("UNICORN primary (%d)", primary_id);
         std::map<int32_t, std::shared_ptr<IDiveBlock>> blocks_to_write = {};
         if (modifications_map_.count(primary_id) > 0)
         {
@@ -262,7 +261,6 @@ bool DiveBlockData::TraverseBlocks(BlockVisitor& visitor) const
         {
             int32_t                     secondary_id = it->first;
             std::shared_ptr<IDiveBlock> current_modification_ptr = it->second;
-            GFXRECON_LOG_INFO("UNICORN secondary (%d, %d)", primary_id, secondary_id);
 
             if (current_modification_ptr == nullptr)
             {


third_party/gfxreconstruct/framework/decode/dive_block_data.h
```diff
@@ -52,24 +108,33 @@ class DiveBlockData
...
+                                      original_blocks_map_        = {}; // Starting block index of 0
+    DiveSingleGfxrBlockInOriginalFile original_header_block_      = {};
+    bool                              original_blocks_map_locked_ = false;
+
+    // Info for modifications
+    //
+    // The first key (uint32_t) is the primary_id which is the original_id. Valid values:
+    // [0...original_blocks_map_.size()-1]
+    //
+    // The second key (int32_t) is the secondary_id which represents the position of
```
wangra-google, line 132: sorry, could you remind me if there is a diff between -1 and -100 for secondary_id? is it used for sorting if there are multiple modifications next to the same primary_id?

third_party/gfxreconstruct/framework/decode/dive_block_data.h
```diff
@@ -30,20 +32,74 @@ static constexpr size_t kDiveBlockBufferSize = 4096;
...
+    bool Visit(DiveSingleGfxrBlockInBuffer& block) override;
 
+  private:
+    FILE* original_file_ptr_                 = nullptr;
+    FILE* new_file_ptr_                      = nullptr;
+    char  copy_buffer_[kDiveBlockBufferSize] = {};
+};
+
+// Abstract class representing a binary block encoded in .gfxr format
+class DiveSingleBlock
```
wangra-google, line 64: just curious, it is named as `Single`Block because there will be cases for `Multi`Block in the future?

third_party/gfxreconstruct/framework/decode/dive_block_data.h
```diff
@@ -30,20 +32,74 @@ static constexpr size_t kDiveBlockBufferSize = 4096;
...
+    ~DiveSingleGfxrBlockInOriginalFile() {}
+    bool Accept(BlockVisitor& visitor) override { return visitor.Visit(*this); }
+
+    uint64_t offset_ = 0;
+    uint64_t size_   = 0;
+};
+
+// Representing a gfxr-encoded block in a buffer
+// If blob_ptr is undefined, that is interpreted as an empty block
+class DiveSingleGfxrBlockInBuffer : public DiveSingleBlock
```
wangra-google, line 87: just to make sure i understand correctly, all modifications will be in DiveSingleGfxrBlockInBuffer? if that is the case, would it be better to call this `DiveGfxModificationBlock`? (and the other one as `DiveGfxOriginalBlock`)?

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
```diff
@@ -141,48 +203,47 @@ bool DiveBlockData::WriteGFXRFile(const std::string& original_file_path, const s
...
-            BlockBytesLocation current_block_location = original_blocks_map_.at(current_block_metadata.id);
+            blocks_to_write[0] = original_blocks_map_[primary_id];
+        }
 
-            if (current_block_location.size > max_block_size)
-            {
-                max_block_size = current_block_location.size;
-            }
+        // For a given primary_id, go through block-by-block in order of secondary_id
+        for (auto it = blocks_to_write.begin(); it != blocks_to_write.end(); ++it)
```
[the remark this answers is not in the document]
angela28chen: I thought that since I'm using an ordered map (blocks_to_write), this would already be in ascending secondary_id order (and not insertion order). I will definitely make a unit test for this case though, thanks for pointing it out.

third_party/gfxreconstruct/framework/decode/dive_block_data.h
angela28chen: I haven't considered multi block units. This one is just named single block to distinguish it from DiveBlockData (which is meant to hold a bunch of single blocks. If you have different ideas for naming, any suggestions are welcome.

third_party/gfxreconstruct/framework/decode/dive_block_data.h
angela28chen: Sure, I think that might be easier naming to understand so I will change it. I picked names to describe where the actual encoded bytes of the block exist, but I think original/modified is a bit easier to read.

third_party/gfxreconstruct/framework/decode/dive_block_data.h
angela28chen: There is a difference, in replay order, for a given primary index, the secondary index -100 block would be first, then the -1 block, then 0 (original block). There shouldn't be two blocks with the same order (only case is modified block at secondary index 0 and the original block, in which case the original block is ignored in replay).

I think in the future DiveBlockData may need some functionality for returning/overwriting/clearing all modifications for a given primary_id, depending on how we implement what-ifs. My current idea is that the flow for "advanced whatifs" would be something like this:

1. UI queries DiveBlockData to get the blocks and displays them in a list.
2. UI displays draggable modification blocks anchored to a given primary_id. User makes whatever changes they desire and then press a button to save the modifications to new GFXR file
3. UI translates those blocks into adjacent secondary_ids. If there are 3 blocks before the original it becomes (-3,-2-1) and if there are 2 it becomes (-2, -1) and so on. 
4. These new modifications overwrite the modifications map in DiveBlockData associated with that given primary_id

third_party/gfxreconstruct/framework/decode/dive_block_data.cpp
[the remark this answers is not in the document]
wangra-google: sounds good, thanks

third_party/gfxreconstruct/framework/decode/dive_block_data.h
wangra-google: thanks

third_party/gfxreconstruct/framework/decode/dive_block_data.h
wangra-google: Maybe simply `IDiveBlock`?

angela28chen:
Note: blocked by https://github.com/google/dive/pull/234 because I want to hook this feature up to the CLI for testing

gfxr_ext/decode/dive_block_data.cpp
```diff
@@ -74,63 +161,136 @@ bool DiveBlockData::FinalizeOriginalBlocksMapSizes()
...
+    modifications_map_[primary_id].erase(secondary_id);
+    return true;
+}
+
+bool DiveBlockData::TraverseBlocks(BlockVisitor& visitor) const
+{
+    // Go through block-by-block in order of primary_id
+    for (uint32_t primary_id = 0; primary_id < original_blocks_map_.size(); primary_id++)
+    {
+        GFXRECON_LOG_INFO("UNICORN primary (%d)", primary_id);
```
wangra-google, line 245: curious why UNICORN here?

wangra-google (approved):

gfxr_ext/decode/dive_block_data.cpp
```diff
@@ -74,63 +161,136 @@ bool DiveBlockData::FinalizeOriginalBlocksMapSizes()
...
-            return false;
+            blocks_to_write[0] = original_blocks_map_.at(primary_id);
+        }
+
+        // For a given primary_id, go through block-by-block in order of secondary_id
+        for (auto it = blocks_to_write.begin(); it != blocks_to_write.end(); ++it)
+        {
+            int32_t                     secondary_id = it->first;
+            std::shared_ptr<IDiveBlock> current_modification_ptr = it->second;
+            GFXRECON_LOG_INFO("UNICORN secondary (%d, %d)", primary_id, secondary_id);
```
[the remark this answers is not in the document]
angela28chen: removed

wangra-google (approved):

PR #225, opened 2025-06-03
Fix clang build error on Linux

RenfengLiu:
This fixed the error "fatal error: 'util/defines.h' file not found". And also added CI workflow to build wit clang-14

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,7 +65,7 @@ jobs:
           cd build; ctest --output-on-failure -C Release
 
 
-  build_Linux_Gcc_Debug:
+  build_Linux_Debug:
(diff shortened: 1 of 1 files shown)


angela28chen (approved):
LGTM, thanks for fixing this!

shanminchao (dismissed):
Double checked and made sure these changes fixes the build issue on my machine. It does! Approved.

shanminchao:
Not sure why the build_Linux checks are pending? Probably related to this PR...

angela28chen (approved):
That's strange, I think I was able to delete the old lint workflow and add new ones in the same PR. Maybe there's something on the github website that can be done to clear the pending github actions?

PR #226, opened 2025-06-04
Communication protocol for Dive: Messages

elviscapiaq:
- Messages (Request/Response) were added for the socket connection.
- Helper functions for messages were added.

### DIFF
--- a/network/serializable.h
+++ b/network/serializable.h
@@ -21,10 +21,10 @@ limitations under the License.
 
 #include "absl/status/statusor.h"
 
-typedef std::vector<uint8_t> Buffer;
-
 namespace Network
 {
+
+using Buffer = std::vector<uint8_t>;
 
 class ISerializable
 {
--- a/network/socket_connection.cc
+++ b/network/socket_connection.cc
@@ -24,7 +24,7 @@ limitations under the License.
 #    define NOMINMAX
 #    include <ws2tcpip.h>
 #    pragma comment(lib, "Ws2_32.lib")
-typedef SSIZE_T ssize_t;
+using ssize_t = SSIZE_T;
 #else
 #    include <netdb.h>
 #    include <poll.h>
--- a/network/socket_connection.h
+++ b/network/socket_connection.h
@@ -23,12 +23,12 @@ limitations under the License.
 
 #ifdef WIN32
 // On Windows, a socket is a pointer-sized handle to ensure 32/64-bit compatibility.
-typedef uintptr_t SocketType;
+using SocketType = uintptr_t;
 // The value for an invalid socket on Windows is INVALID_SOCKET (~0).
 constexpr SocketType kInvalidSocketValue = ~static_cast<SocketType>(0);
 #else
 // On POSIX systems, a socket is a file descriptor (`int`).
-typedef int SocketType;
+using SocketType = int;
 // Functions that return a file descriptor use -1 to indicate an error.
 constexpr SocketType kInvalidSocketValue = -1;
 #endif


network/messages.cc
```diff
@@ -0,0 +1,334 @@
...
+    if (src.size() < offset + len)
+    {
+        return absl::InvalidArgumentError("Buffer too small for declared string length.");
+    }
+    std::string result(reinterpret_cast<const char*>(src.data() + offset), len);
+    offset += len;
+    return result;
+}
+
+absl::Status HandShakeMessage::Serialize(Buffer& dest) const
```
wangra-google, line 71: does this need to return absl::Status at all if the return value is always OkStatus?

network/messages.cc
```diff
@@ -0,0 +1,334 @@
...
+    major_version = major_or.value();
+
+    absl::StatusOr<uint32_t> minor_or = ReadUint32FromBuffer(src, offset);
+    if (!minor_or.ok())
+    {
+        return minor_or.status();
+    }
+    minor_version = minor_or.value();
+
+    if (offset != src.size())
```
wangra-google, line 96: sorry i am a bit confused here, why does the offset need to have the same size as the src?

wangra-google: oh, i see, it is because the src should only contain the version in this case, right?

network/messages.h
```diff
@@ -0,0 +1,161 @@
...
+
+class HandShakeResponse : public HandShakeMessage
+{
+public:
+    static constexpr uint32_t TYPE_ID = static_cast<uint32_t>(MessageType::HANDSHAKE_RESPONSE);
+
+    uint32_t GetMessageType() const override { return TYPE_ID; }
+};
+
+class EmptyMessage : public ISerializable
```
wangra-google, line 77: out of curiosity, what is this empty message for?

network/messages.cc
```diff
@@ -0,0 +1,334 @@
...
+        return absl::InvalidArgumentError("Handshake message has unexpected trailing data.");
+    }
+    return absl::OkStatus();
+}
+
+absl::Status StringMessage::Serialize(Buffer& dest) const
+{
+    dest.clear();
+    WriteStringToBuffer(str, dest);
+    return absl::OkStatus();
```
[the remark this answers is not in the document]
wangra-google: i guess it is just to be consistent to Deserialize? and there might be cases in the future that Serialize might fail? (not necessarily change, just want to make sure i understand. thanks :) )

elviscapiaq:
> A general comment: better to follow the Dive naming convention by adding `m_` prefix to all class members

Updated to this naming convention.

wangra-google (approved):

network/messages.h
```diff
@@ -0,0 +1,196 @@
...
+// Helper to read a string (length + data) from the buffer.
+absl::StatusOr<std::string> ReadStringFromBuffer(const Buffer& src, size_t& offset);
+
+enum class MessageType : uint32_t
+{
+    HANDSHAKE_REQUEST = 1,
+    HANDSHAKE_RESPONSE = 2,
+    PING_MESSAGE = 3,
+    PONG_MESSAGE = 4,
+    CAPTURE_REQUEST = 5,
```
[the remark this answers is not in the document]
elviscapiaq: Actually, it's PM4 capture. We are going to replace the current gRPC implementation, which only performs PM4 capture when using the server. We can discuss Gfxr capture using sockets once this implementation is complete. I've updated the naming to be explicit (e.g., PM4_CAPTURE_REQUEST). Thanks for pointing this out.

wangra-google (approved):

network/serializable.h
```diff
@@ -0,0 +1,46 @@
...
+*/
+
+#pragma once
+
+#include <cstdint>
+#include <vector>
+
+#include "absl/status/statusor.h"
+
+typedef std::vector<uint8_t> Buffer;
```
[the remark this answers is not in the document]
elviscapiaq: Updated. Also, I am adding a new commit to make clear the changes.

wangra-google (approved):

PR #229, opened 2025-06-04
Capture PM4 data on Gfxr replay

RenfengLiu:
This is the First step towards doing the PM4 capture on gfxr replay.
- it added the wrap script to the replay apk, so that the libwrap.so can be pre-loaded if the flag `--dump_pm4` is present
- It current only used the dive_client_cli to do the replay and download the PM4 capture to the specified local path. 

The are several limitation that needs to be address once we have the socket integrated to the replay, so that we can:
1. No UI integration yet.
2. control the start/stop of the capture. currently we start/stop when the apk start/end to run. 
3. set the name of the capture file, currently it's done by the setprop command. 
4. no error reporting back if things not going correctly. 
5. interface to control libwrap will need to updated once we have socket integrated. 

To trigger the pm4 dump on replay, using the following command:
```
build/bin/dive_client_cli --device localhost:43805 --command gfxr_replay --gfxr_replay_file_path /data/local/tmp/gfxr_capture/com.google.bigwheels.project_sample_04_cube.debug_trim_trigger_20250531T230841.gfxr  --dump_pm4 --download_path /usr/local/google/home/renfeng/traces
```

### DIFF
--- a/network/CMakeLists.txt
+++ b/network/CMakeLists.txt
@@ -18,9 +18,9 @@ cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 
 project(network)
 
-set(NETWORK_SRCS socket_connection.cc messages.cc)
+set(NETWORK_SRCS socket_connection.cc messages.cc tcp_client.cc)
 
-set(NETWORK_HDRS socket_connection.h serializable.h messages.h)
+set(NETWORK_HDRS platform_net.h socket_connection.h serializable.h messages.h tcp_client.h)
 
 add_library(network SHARED ${NETWORK_SRCS} ${NETWORK_HDRS})
 
--- a/network/messages.cc
+++ b/network/messages.cc
@@ -16,12 +16,6 @@ limitations under the License.
 
 #include "messages.h"
 
-#ifdef WIN32
-#    include <winsock2.h>
-#else
-#    include <netinet/in.h>
-#endif
-
 constexpr uint32_t kMaxPayloadSize = 16 * 1024 * 1024;
 
 namespace Network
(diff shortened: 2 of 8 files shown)


capture_service/constants.h
```diff
@@ -30,14 +30,19 @@ inline constexpr char kXrLayerName[] = "XR_APILAYER_dive";
 inline constexpr char kTargetPath[] = "/data/local/tmp";
 inline constexpr char kGfxrTargetPath[] = "/data/data/";
 inline constexpr char kGfxrReconPyPath[] = "third_party/gfxreconstruct/android/scripts/gfxrecon.py";
-inline constexpr char kDeviceCaptureDirectory[] = "/sdcard/Download/";
+inline constexpr char kDeviceCaptureDirectory[] = "/sdcard/Download";
```
wangra-google, line 33: nit: maybe better to be `kDeviceCapturePath` just to be consistent with other consts in this file?

capture_service/constants.h
```diff
@@ -30,14 +30,19 @@ inline constexpr char kXrLayerName[] = "XR_APILAYER_dive";
...
 inline constexpr char kManifestFileName[] = "XrApiLayer_dive.json";
 inline constexpr char kManifestFilePath[] = "/system/etc/openxr/1/api_layers/implicit.d/";
```
wangra-google, line 35: i guess we could remove the `/` at the end to be consistent with other path?

third_party/freedreno/wrap/wrap-util.c
```diff
@@ -206,8 +206,7 @@ void rd_start(int device_fd, const char *name, const char *fmt, ...)
 	testnum = getenv("TESTNUM");
 	if (testnum) {
 		n = strtol(testnum, NULL, 0);
-		// GOOGLE: add open_count to the filename.
```
wangra-google, line 209: is there any reason why you remove this comment?

wangra-google (approved):

GrantComm (approved):

PR #234, opened 2025-06-05
Make new host CLI tool for GFXR file related functionality

angela28chen:
Moving the gfxr-file related CLI commands inside `divecli` to a new CLI tool `host_cli` to make use of abseil flag library

### DIFF
--- a/host_cli/CMakeLists.txt
+++ b/host_cli/CMakeLists.txt
@@ -14,8 +14,6 @@
 # limitations under the License.
 #
 
-cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
-
 project(host_cli)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
--- a/host_cli/data_core_wrapper.cpp
+++ b/host_cli/data_core_wrapper.cpp
@@ -20,9 +20,7 @@
 #include "dive_core/capture_data.h"
 #include "dive_core/data_core.h"
 
-namespace Dive
-{
-namespace HostCli
+namespace Dive::HostCli
 {
 
 DataCoreWrapper::DataCoreWrapper()
@@ -69,5 +67,4 @@ absl::Status DataCoreWrapper::WriteNewGfxrFile(const std::string& new_gfxr_file_
     return absl::OkStatus();
 }
 
-}  // namespace HostCli
-}  // namespace Dive
+}  // namespace Dive::HostCli
--- a/host_cli/data_core_wrapper.h
+++ b/host_cli/data_core_wrapper.h
@@ -23,9 +23,7 @@
 
 #include "absl/status/status.h"
 
-namespace Dive
-{
-namespace HostCli
+namespace Dive::HostCli
 {
 
 // Initializes DataCore and provides access to it, also stores relevant info for operations
@@ -43,5 +41,4 @@ private:
     std::unique_ptr<Dive::DataCore> m_data_core = nullptr;
 };
 
-}  // namespace HostCli
-}  // namespace Dive
+}  // namespace Dive::HostCli
--- a/host_cli/data_core_wrapper_test.cpp
+++ b/host_cli/data_core_wrapper_test.cpp
@@ -18,9 +18,7 @@
 
 #include <gtest/gtest.h>
 
-namespace Dive
-{
-namespace HostCli
+namespace Dive::HostCli
 {
 namespace
 {
@@ -36,5 +34,4 @@ TEST(DataCoreWrapperTest, CheckInitialization)
 // DataCore
 
 }  // namespace
-}  // namespace HostCli
-}  // namespace Dive
+}  // namespace Dive::HostCli
--- a/host_cli/main.cpp
+++ b/host_cli/main.cpp
@@ -51,9 +51,9 @@ absl::Status ValidateFlags()
     {
         std::filesystem::path input_fp = input_file_path;
         std::string           ext = input_fp.extension().string();
-        if ((ext == "") || std::find(kAllowedInputFileExtensions.begin(),
-                                     kAllowedInputFileExtensions.end(),
-                                     ext) == kAllowedInputFileExtensions.end())
+        if ((ext.empty()) || std::find(kAllowedInputFileExtensions.begin(),
+                                       kAllowedInputFileExtensions.end(),
+                                       ext) == kAllowedInputFileExtensions.end())
         {
             return absl::InvalidArgumentError(
             absl::StrFormat("unrecognizable extension provided for --input_file_path: %s", ext));
@@ -62,7 +62,7 @@ absl::Status ValidateFlags()
     }
 
     std::string output_gfxr_path = absl::GetFlag(FLAGS_output_gfxr_path);
-    if (output_gfxr_path != "")
+    if (!output_gfxr_path.empty())
     {
         if (input_file_ext != ".gfxr")
         {
@@ -122,7 +122,7 @@ int main(int argc, char **argv)
         }
 
         std::string output_gfxr_path = absl::GetFlag(FLAGS_output_gfxr_path);
-        if (output_gfxr_path == "")
+        if (!output_gfxr_path.empty())
         {
             // Nothing further to do with the loaded .gfxr file
             return 0;


angela28chen:
I think this will save me time when adding more features to the modification framework. The old CLI code doesn't parse flags well and basically checks them positionally, which isn't something that will scale well when more flags are added. But I need to add more flags (such as for #224 I want a repeated/list flag to indicate which blocks to delete by ID)

I want to eventually move all the functionality from divecli over, but in the interest of saving time I'm not doing that now.

.gitignore
```diff
@@ -11,6 +11,9 @@
 /compile_commands.json
 /.cache/
 
+# ctest
+Testing/
```
wangra-google, line 15: is it intended to be submitted?

wangra-google (approved):

.gitignore
angela28chen: Oops, will remove this. I thought ctest generated a `Testing/` folder in the dive root dir but it actually does it inside `build/` when you run it properly from there.

wangra-google (dismissed):

host_cli/data_core_wrapper_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+{
+
+TEST(DataCoreWrapperTest, CheckInitialization)
+{
+    DataCoreWrapper data_core_wrapper;
+    ASSERT_EQ(data_core_wrapper.IsDataCoreInitialized(), true);
+    ASSERT_EQ(data_core_wrapper.IsGfxrLoaded(), false);
+}
+
+// TODO : Write more tests if it's possible to hook up actual files from tests/ or with a mock
```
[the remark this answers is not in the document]
angela28chen: Thanks for the context, I'll keep this in mind

host_cli/main.cpp
```diff
@@ -0,0 +1,141 @@
...
+          "",
+          "If specified, a new .gfxr file will be generated from the original file "
+          "(--input_file_path) and any specified modifications");
+
+absl::Status ValidateFlags()
+{
+    std::string input_file_ext = "";
+
+    std::string input_file_path = absl::GetFlag(FLAGS_input_file_path);
+    if (input_file_path != "")
```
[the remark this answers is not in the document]
angela28chen: I wanted to set this up because the intent is that in the future the `--input_file_path` will be used for files of multiple types (namely .dive but also other things) with branching behavior depending on file type. And file type being recognized by this CLI via the extension, so we expect files to end in .gfxr = they must be a gfxr capture.

The reason for this is that from what I understand of `divecli` it offers a lot of functionality based on what is fed in. It can perform tasks on `.dive` and `.bin` and raw strings passed through the command line arguments, but generally only one type per call. My idea is all files will be passed with `--input_file_path` (separate flag for raw string) and the behaviour will branch depending on what the input is.

host_cli/data_core_wrapper.h
```diff
@@ -0,0 +1,47 @@
...
+namespace Dive
+{
+namespace HostCli
+{
+
+// Initializes DataCore and provides access to it, also stores relevant info for operations
+class DataCoreWrapper
+{
+public:
+    DataCoreWrapper();
```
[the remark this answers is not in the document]
angela28chen: I think I see what you're getting at, but I'm not sure about making this into a factory function because I think that I want to use this `DataCoreWrapper` class to keep `main.cpp` short and implement more things in here instead (mainly the things about modifying the GFXR data before writing the new file). 

I want to implement and merge this PR + the following PR with delete block functionality, and then consider a class refactoring after that if you think it's a good idea still. Would that work?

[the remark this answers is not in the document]
angela28chen: But I will try to remove `m_loaded_gfxr` and just check inside m_data_core for that info

host_cli/data_core_wrapper.cpp
```diff
@@ -0,0 +1,69 @@
...
+namespace Dive
+{
+namespace HostCli
+{
+
+DataCoreWrapper::DataCoreWrapper()
+{
+    // Initialize DataCore
+    Dive::LogConsole log;
+    m_data_core = std::make_unique<Dive::DataCore>(&log);
```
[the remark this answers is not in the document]
angela28chen: Thanks, fixed

host_cli/main.cpp
```diff
@@ -0,0 +1,143 @@
...
+                                     ext) == kAllowedInputFileExtensions.end())
+        {
+            return absl::InvalidArgumentError(
+            absl::StrFormat("unrecognizable extension provided for --input_file_path: %s", ext));
+        }
+        input_file_ext = ext;
+    }
+
+    std::string output_gfxr_path = absl::GetFlag(FLAGS_output_gfxr_path);
+    if (output_gfxr_path != "")
```
footballhead, line 65: consider using `!empty()` here too
```suggestion
    if (!output_gfxr_path.empty())
```

host_cli/main.cpp
```diff
@@ -0,0 +1,143 @@
...
+absl::Status ValidateFlags()
+{
+    std::string input_file_ext = "";
+
+    std::string input_file_path = absl::GetFlag(FLAGS_input_file_path);
+    if (!input_file_path.empty())
+    {
+        std::filesystem::path input_fp = input_file_path;
+        std::string           ext = input_fp.extension().string();
+        if ((ext == "") || std::find(kAllowedInputFileExtensions.begin(),
```
footballhead, line 54: consider using `empty` here too
```suggestion
        if ((ext.empty()) || std::find(kAllowedInputFileExtensions.begin(),
```

host_cli/data_core_wrapper_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+*/
+
+#include "data_core_wrapper.h"
+
+#include <gtest/gtest.h>
+
+namespace Dive
+{
+namespace HostCli
+{
```
footballhead, line 24: You can combine there onto one line

```suggestion
namespace Dive::HostCli
{
```

host_cli/data_core_wrapper.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include "dive_core/capture_data.h"
+#include "dive_core/data_core.h"
+
+#include "absl/status/status.h"
+
+namespace Dive
+{
+namespace HostCli
+{
```
footballhead, line 29: You can combine there onto one line
```suggestion
namespace Dive::HostCli
{
```

host_cli/data_core_wrapper.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "data_core_wrapper.h"
+
+#include "dive_core/log.h"
+#include "dive_core/capture_data.h"
+#include "dive_core/data_core.h"
+
+namespace Dive
+{
+namespace HostCli
+{
```
footballhead, line 26: You can combine there onto one line
```suggestion
namespace Dive::HostCli
```

host_cli/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
+
+project(host_cli)
```
footballhead, line 19: My understanding is that these are only required when trying to build the directory standalone like

```
cmake host_cli ...
```

Otherwise they're extra noise in the CMakeLists

host_cli/main.cpp
angela28chen: Fixed

host_cli/main.cpp
```diff
@@ -0,0 +1,141 @@
...
+int main(int argc, char **argv)
+{
+    absl::SetProgramUsageMessage(
+    absl::StrCat("This CLI tool is intended to provide access to the dive_core"
+                 "\nlibrary for utility and for testing. Currently it supports"
+                 "\nmanipulation of .gfxr files. Sample usage:\n\n",
+                 argv[0],
+                 " --helpfull"));
+    absl::ParseCommandLine(argc, argv);
+    absl::Status res = ValidateFlags();
```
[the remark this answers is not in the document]
elviscapiaq: That is what I meant. Thanks for updating it!

elviscapiaq (approved):

PR #235, opened 2025-06-09
Dive communication protocol: Client implementation

elviscapiaq:
- Threading:
    - First thread: Main thread to send requests to the server.
    - Second thread: A keep-alive thread checks the connection periodically via a ping-pong mechanism.
- A timeout parameter was added to Recv function to allow pingServer to use it.

### DIFF
--- a/network/messages.cc
+++ b/network/messages.cc
@@ -16,13 +16,6 @@ limitations under the License.
 
 #include "messages.h"
 
-#ifdef WIN32
-#    include <winsock2.h>
-#    undef SendMessage
-#else
-#    include <netinet/in.h>
-#endif
-
 constexpr uint32_t kMaxPayloadSize = 16 * 1024 * 1024;
 
 namespace Network
--- a/network/platform_net.h
+++ b/network/platform_net.h
@@ -1,0 +1,47 @@
+/*
+Copyright 2025 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#pragma once
+
+#ifdef WIN32
+#    define _WINSOCK_DEPRECATED_NO_WARNINGS
+#    define NOMINMAX
+#    include <winsock2.h>
+#    include <ws2tcpip.h>
+#    pragma comment(lib, "Ws2_32.lib")
+
+// On Windows, a socket is a pointer-sized handle to ensure 32/64-bit compatibility.
+using SocketType = uintptr_t;
+// The value for an invalid socket on Windows is INVALID_SOCKET (~0).
+constexpr SocketType kInvalidSocketValue = ~static_cast<SocketType>(0);
+
+using ssize_t = SSIZE_T;
+
+#    undef SendMessage
+
+#else
+#    include <netdb.h>
+#    include <poll.h>
+#    include <sys/un.h>
+#    include <unistd.h>
+#    include <netinet/in.h>
+
+// On POSIX systems, a socket is a file descriptor (`int`).
+using SocketType = int;
+// Functions that return a file descriptor use -1 to indicate an error.
+constexpr SocketType kInvalidSocketValue = -1;
+
+#endif
--- a/network/tcp_client.cc
+++ b/network/tcp_client.cc
@@ -18,10 +18,13 @@ limitations under the License.
 
 #include "tcp_client.h"
 
+namespace
+{
 constexpr uint32_t kKeepAliveIntervalSec = 5;
 constexpr uint32_t kPingTimeoutMs = 5000;
 constexpr uint32_t kHandshakeMajorVersion = 1;
 constexpr uint32_t kHandshakeMinorVersion = 0;
+}  // namespace
 
 namespace Network
 {
--- a/network/CMakeLists.txt
+++ b/network/CMakeLists.txt
@@ -20,7 +20,7 @@ project(network)
 
 set(NETWORK_SRCS socket_connection.cc messages.cc tcp_client.cc)
 
-set(NETWORK_HDRS socket_connection.h serializable.h messages.h tcp_client.h)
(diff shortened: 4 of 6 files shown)


elviscapiaq:
I fixed the error for WIndows. The `Winsocket` library has a macro called `SendMessage` which is a function name I am also using in `messages.cc`, so this Windows macro was replacing `SendMessage` to `SendMessageA` and I was getting 'no definition of symbol found ...'. For many hours I assumed the error was linking the symbols for the DLL on WIndows. it was painful but I learned something new from Win Sockets.

network/tcp_client.cc
```diff
@@ -0,0 +1,490 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <chrono>
+
+#include "tcp_client.h"
+
+constexpr int      kKeepAliveIntervalSec = 5;
```
wangra-google, line 21: curious why this is a `int` not an `uint32_t`?

network/tcp_client.cc
```diff
@@ -0,0 +1,490 @@
...
+    }
+}
+
+ClientStatus TcpClient::GetClientStatus() const
+{
+    std::lock_guard<std::mutex> lock(m_status_mutex);
+    return m_status;
+}
+
+ClientStatus TcpClient::SetClientStatus(ClientStatus status)
```
wangra-google, line 484: when do we need the return value? and why do we return `return m_status = status;`

network/tcp_client.cc
```diff
@@ -0,0 +1,490 @@
...
+        }
+        return absl::Status(handshake_status.code(),
+                            absl::StrCat("Connect: Handshake fail: ", handshake_status.message()));
+    }
+
+    std::cout << "Client: Connected & StartKeepAlive." << std::endl;
+    auto keep_alive_status = StartKeepAlive();
+    if (!keep_alive_status.ok())
+    {
+        if (absl::IsAlreadyExists(keep_alive_status))
```
wangra-google, line 98: do we need to call `m_connection.reset();` here?

network/tcp_client.cc
```diff
@@ -0,0 +1,490 @@
...
+        {
+            std::cout << keep_alive_status.message() << std::endl;
+        }
+        else
+        {
+            SetClientStatus(ClientStatus::STATUS_ERROR);
```
wangra-google, line 104: do we need to call m_connection.reset(); here?

network/tcp_client.cc
```diff
@@ -0,0 +1,490 @@
...
+    SetClientStatus(ClientStatus::CONNECTED);
+
+    std::cout << "Client: Connected & handshaking." << std::endl;
+    auto handshake_status = PerformHandshake();
+    if (!handshake_status.ok())
+    {
+        SetClientStatus(ClientStatus::STATUS_ERROR);
+        if (m_connection)
+        {
+            m_connection->Close();
```
wangra-google, line 87: it seems that m_connection->Close(); always come with m_connection.reset();, maybe a helper func to group those 2 calls?

network/tcp_client.cc
```diff
@@ -0,0 +1,490 @@
...
+                return !m_keep_alive.running.load();
+            }))
+        {
+            std::cout << "KeepAliveLoop: Client Keep-Alive Thread: Stop signaled via CV."
+                      << std::endl;
+            break;
+        }
+        lk.unlock();
+
+        if (!m_keep_alive.running.load())
```
wangra-google, line 441: isn't the `wait_for` itself returning true if !m_keep_alive.running.load() was the reason it woke up? do you still need this extra check here?

wangra-google (dismissed):

network/messages.cc
```diff
@@ -18,6 +18,7 @@ limitations under the License.
 
 #ifdef WIN32
 #    include <winsock2.h>
+#    undef SendMessage
```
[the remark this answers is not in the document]
angela28chen: I think we can consider:

1. Rename all functions that have names that are interpreted as windows macros (current strategy)
2. Have a common header file with the undefs of common windows macros and include that header with all dive files

I don't think we should add `#undef` on a file by file case, because every time this `SendMessage` is called in a different file wouldn't it cause issues?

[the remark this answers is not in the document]
elviscapiaq: I looked into macros on Windows. It seems even if you use a namespace, the problem persists.
The issue is that the preprocessor runs before the compiler, replacing `SendMessage` to `SendMessageA`, then when the compiler runs it sees `SendMessageA `rather than `SendMessage`.  Therefore. I decided to go with the second recomendation from Angela, because I haven't found a consistent name replacement for `SendMessage` and also we can just put everything related to platform (Windows/POSIX) includes and configurations in this header file. I called this `network_net.h`. Also, we cannot just put `undef` in a header file without taking care of the `Winsocket `library, it means, we need to put `undef ` after the `winsocket `library is included.

network/tcp_client.cc
```diff
@@ -0,0 +1,476 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <chrono>
+
+#include "tcp_client.h"
+
+constexpr uint32_t kKeepAliveIntervalSec = 5;
```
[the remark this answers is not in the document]
elviscapiaq: Thanks, updated!

wangra-google (approved):

network/platform_net.h
```diff
@@ -0,0 +1,47 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#pragma once
+
+#ifdef WIN32
+#    define _WINSOCK_DEPRECATED_NO_WARNINGS
+#    define NOMINMAX
```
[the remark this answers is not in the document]
elviscapiaq: I think `#define WIN32_LEAN_AND_MEAN`  excludes the Winsock API by default, but will take into consideration for the next PR. Thanks for this information!

network/messages.cc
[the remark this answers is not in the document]
elviscapiaq: Given that the majority of people agree with adding a header to undef SendMessage, I will merge this PR, but we can talk in the next meeting if we really want to avoid this. Note that, whenever we use any part of Sockets, then it uses the header file undefinig the SendMessage windows macro, so I would not expect issues in the future but I could be wrong.

PR #236, opened 2025-06-09
Clean up CMakeLists.txt

angela28chen:
Changes:
* Gather global cmake properties in root `CMakeLists.txt`
* Remove global property about verbose makefile since we prefer ninja
* Remove unnecessary checks for subdirectory `CMakeLists.txt` files before `add_subdirectory()` calls
* Remove unnecessary check for grpc tar file

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,8 +19,7 @@ cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 project(dive)
 
 # Global properties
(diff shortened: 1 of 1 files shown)


wangra-google:
> Changes:
> 
> * Gather global cmake properties in root `CMakeLists.txt`
> * Remove unnecessary checks for subdirectory `CMakeLists.txt` files before `add_subdirectory()` calls
> * Remove unnecessary check for grpc tar file

Thanks for cleaning up the cmake files :D

elviscapiaq (dismissed):

PR #239, opened 2025-06-13
Migrate dive-specific gfxr changes to gfxr_ext/

angela28chen:
As discussed, moving dive-specific GFXR changes to under dive source code as much as possible, while still keeping the GFXR style. Now `gfxr_ext/` libraries are how all other Dive targets access gfxr libraries.

Changes:
- Note that this PR builds in about the same time as without it
- Moved `DiveBlockData` and `DiveFileProcessor` out from `third_party/gfxreconstruct`
  - Had to change some integer types to get it to build because Dive has stricter rules around warnings than GFXR
  - Had to keep the non-full path includes setup otherwise `gfxrecon-replay` won't build
  - The formatting is now under the same clang-format rules as the rest of dive sources and also covered by the lint presubmit
- Adding initial unit tests for `gfxr_ext/`
- Added includes per-target and eliminated some unnecessary include directories.
- Moved the `psapi` linked library dependency to `gfxr_ext` so that it doesn't need to be added to all targets depending on dive_core
- New build dependency graph:
```
gfxrecon_decode
^
gfxr_decode_ext_lib
^
dive_core
gfxr_decode_ext_lib_test (windows only)
gfxr_dump_resources
gfxrecon-replay (under third_party/gfxreconstruct, android only)
```

Minor changes:
- Fix a bug with gfxr replay not working without `DiveBlockData` registered
- Expanded `archive.h` include to full path because VS compiler couldn't find it for some reason

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,9 +99,6 @@ if (NOT ANDROID)
     add_subdirectory(third_party/googletest)
 endif()
 
-# For allowing full path includes
-include_directories(${CMAKE_CURRENT_SOURCE_DIR})
-
 add_subdirectory(network)
 add_subdirectory(capture_service)
 add_subdirectory(layer)
--- a/dive_core/CMakeLists.txt
+++ b/dive_core/CMakeLists.txt
@@ -207,6 +207,9 @@ target_link_libraries(${PROJECT_NAME} PRIVATE freedreno_ir3)
 target_link_libraries(${PROJECT_NAME} PRIVATE 
     gfxr_decode_ext_lib
 )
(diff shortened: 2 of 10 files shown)


angela28chen:
I should also look at the other places in dive that depend on dive_core and have additional include dirs added

angela28chen:
Not in this PR:
- Planning to expand on the unit tests in a future CL
- Start expanding includes to use the full path

angela28chen:
I think we should also look into how the X11 header can be properly separated from other things (mostly gtest and absl where it causes collisions). Maybe we can have some kind of library that exposes only part of vulkan headers and have dive_core depend on that? I'm not sure how difficult this would be.

PR #240, opened 2025-06-16
Add memory header to main_window.h

angela28chen:
Without this, Dive fails to build at HEAD on Linux for some compilers.


footballhead (approved):
nit: PR message could mention that this fixes a build problem for some compilers. This gives more context for people looking through the git history in the future.

wangra-google (approved):

PR #241, opened 2025-06-17
Fix multiple issues for Capture UI & CLI on Windows

wangra-google:
- remove \r at the end of capture file name
- use / instead of \ to separate path and file name
- set the default filter to debuggable apks
- auto close the downloading window when the download is done

### DIFF
--- a/capture_service/android_application.cc
+++ b/capture_service/android_application.cc
@@ -137,7 +137,7 @@ absl::Status AndroidApplication::ParsePackage()
 absl::Status AndroidApplication::Start()
 {
     RETURN_IF_ERROR(m_dev.Adb().Run("shell input keyevent KEYCODE_WAKEUP"));
-    RETURN_IF_ERROR(m_dev.Adb().Run(absl::StrFormat("shell am start -S -W %s -n %s/%s ",
+    RETURN_IF_ERROR(m_dev.Adb().Run(absl::StrFormat("shell am start -S -W %s -n %s/%s",
                                                     m_command_args,
                                                     m_package,
                                                     m_main_activity)));
@@ -268,10 +268,14 @@ absl::Status AndroidApplication::GfxrSetup()
     RETURN_IF_ERROR(m_dev.Adb().Run(
     absl::StrFormat("shell settings put global gpu_debug_layer_app %s", m_package)));
 
-    std::string gfxr_capture_directory = std::string(kDeviceCapturePath) + "/" +
-                                         m_gfxr_capture_file_directory;
+    std::string gfxr_capture_directory = absl::StrCat(kDeviceCapturePath,
+                                                      "/",
+                                                      m_gfxr_capture_file_directory);
 
-    std::string capture_file_location = gfxr_capture_directory + "/" + m_package + ".gfxr";
+    std::string capture_file_location = absl::StrCat(gfxr_capture_directory,
+                                                     "/",
+                                                     m_package,
+                                                     ".gfxr");
     RETURN_IF_ERROR(CreateGfxrDirectory(gfxr_capture_directory));
 
     RETURN_IF_ERROR(
--- a/capture_service/dive_client_cli.cc
+++ b/capture_service/dive_client_cli.cc
@@ -345,8 +345,10 @@ bool trigger_capture(Dive::DeviceManager& mgr)
 absl::Status is_capture_directory_busy(Dive::DeviceManager& mgr,
                                        const std::string&   gfxr_capture_directory)
 {
-    std::string on_device_capture_directory = Dive::kDeviceCapturePath + gfxr_capture_directory;
-    std::string command = "shell lsof " + on_device_capture_directory;
+    std::string                 on_device_capture_directory = absl::StrCat(Dive::kDeviceCapturePath,
+                                                           "/",
+                                                           gfxr_capture_directory);
+    std::string                 command = "shell lsof " + on_device_capture_directory;
     absl::StatusOr<std::string> output = mgr.GetDevice()->Adb().RunAndGetResult(command);
 
     if (!output.ok())
@@ -369,10 +371,11 @@ absl::Status is_capture_directory_busy(Dive::DeviceManager& mgr,
 bool retrieve_gfxr_capture(Dive::DeviceManager& mgr, const std::string& gfxr_capture_directory)
 {
     std::filesystem::path download_path = absl::GetFlag(FLAGS_download_path);
-    std::filesystem::path target_download_path(download_path.string() + "/" +
-                                               gfxr_capture_directory);
-    std::filesystem::path on_device_capture_directory = Dive::kDeviceCapturePath +
-                                                        gfxr_capture_directory;
+    std::filesystem::path target_download_path(
+    absl::StrCat(download_path.string(), "/", gfxr_capture_directory));
+    std::filesystem::path on_device_capture_directory = absl::StrCat(Dive::kDeviceCapturePath,
+                                                                     "/",
+                                                                     gfxr_capture_directory);
 
     std::cout << "Retrieving capture..." << std::endl;
     // Check if the target directory already exists on the local machine.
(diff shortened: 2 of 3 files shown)


GrantComm (approved):

capture_service/android_application.cc
```diff
@@ -266,10 +268,10 @@ absl::Status AndroidApplication::GfxrSetup()
     RETURN_IF_ERROR(m_dev.Adb().Run(
     absl::StrFormat("shell settings put global gpu_debug_layer_app %s", m_package)));
 
-    std::string capture_file_location = kDeviceCapturePath + m_gfxr_capture_file_directory + "/" +
-                                        m_package + ".gfxr";
+    std::string gfxr_capture_directory = std::string(kDeviceCapturePath) + "/" +
+                                         m_gfxr_capture_file_directory;
```
footballhead, line 272: consider absl::StrCat, they claim that its more efficient

https://abseil.io/docs/cpp/guides/strings#abslstrcat-and-abslstrappend-for-string-concatenation

```suggestion
    std::string gfxr_capture_directory = absl::StrCat(kDeviceCapturePath, "/", m_gfxr_capture_file_directory);
```

capture_service/android_application.cc
```diff
@@ -266,10 +268,10 @@ absl::Status AndroidApplication::GfxrSetup()
...
 
-    std::string gfxr_capture_directory = kDeviceCapturePath + m_gfxr_capture_file_directory;
+    std::string capture_file_location = gfxr_capture_directory + "/" + m_package + ".gfxr";
```
footballhead, line 274: consider absl::StrCat, they claim that its more efficient

https://abseil.io/docs/cpp/guides/strings#abslstrcat-and-abslstrappend-for-string-concatenation

```suggestion
    std::string capture_file_location = absl::StrCat(gfxr_capture_directory, "/", m_package , ".gfxr");
```

capture_service/android_application.cc
```diff
@@ -137,8 +137,10 @@ absl::Status AndroidApplication::ParsePackage()
 absl::Status AndroidApplication::Start()
 {
     RETURN_IF_ERROR(m_dev.Adb().Run("shell input keyevent KEYCODE_WAKEUP"));
-    RETURN_IF_ERROR(m_dev.Adb().Run(
-    absl::StrFormat("shell am start -S -W %s -n %s/%s ", m_command_args, m_package, m_main_activity)));
+    RETURN_IF_ERROR(m_dev.Adb().Run(absl::StrFormat("shell am start -S -W %s -n %s/%s ",
```
footballhead, line 140: Looks like there's some whitespace that can be trimmed from the command:
```suggestion
    RETURN_IF_ERROR(m_dev.Adb().Run(absl::StrFormat("shell am start -S -W %s -n %s/%s",
```

ui/trace_window.cpp
```diff
@@ -765,11 +766,11 @@ bool GfxrCaptureWorker::areTimestampsCurrent(Dive::AndroidDevice     *device,
 {
     std::vector<std::string> current_time_stamps;
     std::string              get_first_current_timestamp_command = "shell stat -c %Y " +
-                                                      (m_capture_path / m_file_list[0].data())
-                                                      .string();
+                                                      m_capture_path.string() + "/" +
```
footballhead, line 769: I don't see an explicit dependency on absl in the CMake file but it looks like the rest of the UI uses absl. Given that, we might prefer absl::StrCat here

ui/trace_window.cpp
```diff
@@ -1093,7 +1106,8 @@ void TraceDialog::RetrieveGfxrCapture()
 
     if (m_gfxr_capture_file_local_directory_input_box->text() == "")
     {
-        m_gfxr_capture_file_local_directory_input_box->setText("./gfxr_capture");
+        m_gfxr_capture_file_local_directory_input_box->setText(
+        "./" + QString::fromUtf8(Dive::kDefaultCaptureFolderName));
     }
     else
```
footballhead, line 1112: is `local_gfxr_capture_directory_path` used? if not we can remove the `else` here

ui/trace_window.cpp
```diff
@@ -1134,6 +1149,11 @@ void TraceDialog::RetrieveGfxrCapture()
 
 void TraceDialog::OnGFXRCaptureAvailable(QString const &capture_path)
 {
+    QProgressDialog *progress_bar = findChild<QProgressDialog *>("gfxr_download_progress");
```
footballhead, line 1152: What behavior were you seeing? Based on the docs, this should be taken care of by `setAutoReset` combined with `setAutoClose`:

> autoReset : bool
>
> This property holds whether the progress dialog calls reset() as soon as value() equals maximum()
>
> \- https://doc.qt.io/archives/qt-5.15/qprogressdialog.html#autoReset-prop

> autoClose : bool
>
> This property holds whether the dialog gets hidden by reset()
>
> https://doc.qt.io/archives/qt-5.15/qprogressdialog.html#autoClose-prop

ui/trace_window.cpp
```diff
@@ -795,6 +796,17 @@ absl::StatusOr<int64_t> GfxrCaptureWorker::getGfxrCaptureDirectorySize(Dive::And
 
     m_file_list = absl::StrSplit(std::string(ls_output->data()), '\n');
```
footballhead, line 797: Noticed that the `!ls_output.ok()` doesn't return the error status. Should the function continue if `ls` fails?

footballhead, line 797: Should we consider splitting `\n` vs `\r\n` based on platform? This feels like something abseil should already support...

capture_service/android_application.cc
wangra-google: fixed

ui/trace_window.cpp
wangra-google: fixed

ui/trace_window.cpp
wangra-google: added return

capture_service/dive_client_cli.cc
```diff
@@ -424,10 +425,20 @@ bool retrieve_gfxr_capture(Dive::DeviceManager& mgr, const std::string& gfxr_cap
...
-        std::string source_file = (on_device_capture_directory / file.data()).string();
+        std::string file = file_with_trailing;
+        // Windows-style line endings use \r\n. When absl::StrSplit splits by \n, the \r remains at
+        // the end of each line if the input string originated from a Windows-style line ending.
+        if (!file.empty() && file.back() == '\r')
+        {
+            file.pop_back();
+        }
+
+        std::string target_file = absl::StrCat(target_download_path.string(), "/", file.data());
```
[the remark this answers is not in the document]
wangra-google: I did the change to make sure it is consistent with the UI code.

ui/trace_window.cpp
```diff
@@ -898,9 +915,10 @@ void GfxrCaptureWorker::run()
     // Retrieve each file in the capture directory (capture file and asset file).
     for (std::string file : m_file_list)
     {
-        std::string target_file = (m_target_capture_path / file.data()).string();
-        std::string source_file = (m_capture_path / file.data()).string();
-        auto        retrieve_file = device->RetrieveTrace(source_file, target_file);
+        std::string target_file = absl::StrCat(m_target_capture_path.string(), "/", file.data());
```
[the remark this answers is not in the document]
wangra-google: I did the change to make sure it is consistent with the UI code.

capture_service/dive_client_cli.cc
```diff
@@ -369,7 +369,8 @@ absl::Status is_capture_directory_busy(Dive::DeviceManager& mgr,
 bool retrieve_gfxr_capture(Dive::DeviceManager& mgr, const std::string& gfxr_capture_directory)
 {
     std::filesystem::path download_path = absl::GetFlag(FLAGS_download_path);
-    std::filesystem::path target_download_path(download_path / gfxr_capture_directory);
+    std::filesystem::path target_download_path(
+    absl::StrCat(download_path.string(), "/", gfxr_capture_directory));
     std::filesystem::path on_device_capture_directory = Dive::kDeviceCapturePath +
                                                         gfxr_capture_directory;
```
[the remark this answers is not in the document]
wangra-google: fixed

capture_service/android_application.cc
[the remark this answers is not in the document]
angela28chen: I'm guessing this is a change that's affected by the changes below (adding strcat), and we have some kind of clang-format setting that wants vertical align even across blocks?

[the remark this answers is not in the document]
wangra-google: oh, interesting. good to know.

PR #242, opened 2025-06-17
Dive communication protocol: Server implementation

elviscapiaq:
- The server listens on a Unix domain with abstract namespace.
- There are two threads: A main thread and a thread to accept and handle 
  the unique client connection.

### DIFF
--- a/network/socket_connection.h
+++ b/network/socket_connection.h
@@ -79,16 +79,4 @@ private:
     int        m_accept_timout_ms;
 };
 
-struct SocketConnectionDeleter
-{
-    void operator()(SocketConnection* conn) const
-    {
-        if (conn)
-        {
-            conn->Close();
-            delete conn;
-        }
-    }
-};
-
 }  // namespace Network
--- a/network/unix_domain_server.cc
+++ b/network/unix_domain_server.cc
@@ -131,8 +131,7 @@ absl::Status UnixDomainServer::Start(const std::string& server_address)
                                          conn_status.message()));
     }
 
-    m_listen_connection = std::unique_ptr<SocketConnection, SocketConnectionDeleter>(
-    (*std::move(connection)).release());
+    m_listen_connection = *std::move(connection);
     m_is_running.store(true);
     m_server_thread = std::thread(&UnixDomainServer::AcceptAndHandleClientLoop, this);
     return absl::OkStatus();
@@ -189,8 +188,7 @@ void UnixDomainServer::AcceptAndHandleClientLoop()
 
             {
                 std::lock_guard<std::mutex> lk(m_client_mutex);
-                m_client_connection = std::unique_ptr<SocketConnection, SocketConnectionDeleter>(
-                (*std::move(acc_connection)).release());
+                m_client_connection = *std::move(acc_connection);
             }
             LOGI("AcceptAndHandleClientLoop: New client accepted.");
             m_handler->OnConnect();
--- a/network/unix_domain_server.h
+++ b/network/unix_domain_server.h
@@ -71,9 +71,9 @@ private:
     void ResetClientConnection();
 
     // Server connection.
-    std::unique_ptr<SocketConnection, SocketConnectionDeleter> m_listen_connection;
+    std::unique_ptr<SocketConnection> m_listen_connection;
     // The only one client connection at a time.
-    std::unique_ptr<SocketConnection, SocketConnectionDeleter> m_client_connection;
+    std::unique_ptr<SocketConnection> m_client_connection;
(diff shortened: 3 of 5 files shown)


network/log.h
wangra-google, line 1: this one seems to be the same as the `dive/capture_service/log.h`? maybe we could put this header somewhere shared by both projects instead of duplicating it for each project?

network/unix_domain_server.cc
```diff
@@ -0,0 +1,247 @@
...
+        LOGI("DefaultMessageHandler::HandleMessage: Null message.");
+        return;
+    }
+    if (!client_conn)
+    {
+        LOGI("DefaultMessageHandler::HandleMessage: Null client connection.");
+        return;
+    }
+
+    switch (static_cast<MessageType>(message->GetMessageType()))
```
wangra-google, line 44: is there a reason why we want to return a uint32_t from GetMessageType(), can we return MessageType directly to avoid the casting?

network/unix_domain_server.cc
```diff
@@ -0,0 +1,247 @@
...
+{
+
+DefaultMessageHandler::DefaultMessageHandler() {}
+
+void DefaultMessageHandler::OnConnect()
+{
+    LOGI("DefaultMessageHandler::OnConnect()");
+}
+
+void DefaultMessageHandler::HandleMessage(std::unique_ptr<ISerializable> message,
```
wangra-google, line 30: a more general question: are some of the failures in this function critical failures that would need early return in AcceptAndHandleClientLoop()?

network/unix_domain_server.h
```diff
@@ -0,0 +1,79 @@
...
+    void HandleMessage(std::unique_ptr<ISerializable> message,
+                       SocketConnection*              client_conn) override;
+    void OnDisconnect() override;
+};
+
+class UnixDomainServer
+{
+public:
+    // Constructs the server, taking ownership of the provided IMessageHandler.
+    explicit UnixDomainServer(
```
wangra-google, line 47: The constructor's default argument for handler creates a null pointer, which will cause a crash if the server is initialized without a specific handler. Maybe better with something like this?
```
explicit UnixDomainServer(
    std::unique_ptr<IMessageHandler> handler = std::make_unique<DefaultMessageHandler>());
    ```

File: network/socket_connection.h [hunk 4]
```diff
@@ -79,4 +79,16 @@ class SocketConnection
     int        m_accept_timout_ms;
 };
 
+struct SocketConnectionDeleter
```
angela28chen, line 82: Is this special deleter necessary when the `SocketConnection` destructor also calls `Close()`?

network/unix_domain_server.h
```diff
@@ -0,0 +1,79 @@
...
+
+private:
+    // The primary run loop for the server's worker thread.
+    void AcceptAndHandleClientLoop();
+
+    // Resets the client connection to prepare for a new client.
+    void ResetClientConnection();
+
+    std::unique_ptr<SocketConnection, SocketConnectionDeleter> m_listen_connection;
+    std::unique_ptr<SocketConnection, SocketConnectionDeleter> m_client_connection;
```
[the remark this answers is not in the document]
elviscapiaq: I added class documentation, also for the important class members. Thanks for pointing it out.

network/unix_domain_server.cc
```diff
@@ -0,0 +1,247 @@
...
+
+void UnixDomainServer::AcceptAndHandleClientLoop()
+{
+    while (m_is_running.load())
+    {
+        if (!m_client_connection)
+        {
+            if (!m_listen_connection || !m_listen_connection->IsOpen())
+            {
+                if (m_is_running.load())
```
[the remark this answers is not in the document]
elviscapiaq: I simplified it a little bit.Thanks for pointing it out!

network/log.h
wangra-google: thanks

wangra-google (approved):

network/socket_connection.h
elviscapiaq: I tested and it worked as you mentioned. I didn't realize that reset also goes out of scope calling the destructor.
I removed `SocketConnectionDeleter` as we don't need it anymore. Thank you for poiting it out!

elviscapiaq:
I will update our handling of StatusOr in a separate PR to follow the Abseil tip. This will avoid confusion as there are several parts to this change.

wangra-google (approved):

angela28chen (approved):
Thanks for discussiong unique_ptr with me, I also learned more about it!

PR #244, opened 2025-06-17
Adds gfxr vulkan command hierarchy

GrantComm:
Adds the GFXR Vulkan command hierarchy which will be used to display Vulkan commands and arguments from GFXR in the dive UI. 

The following is what the GFXR Vulkan command hierarchy will be displayed as when a .gfxr file is loaded. Follow-up pr integrates the UI-specific components.
Screenshot 2025-06-17 at 10 22 30 AM

### DIFF
--- a/dive_core/data_core.cpp
+++ b/dive_core/data_core.cpp
@@ -53,22 +53,19 @@ bool DataCore::CreateCommandHierarchy(bool is_gfxr_capture)
         }
         return true;
     }
-    else
-    {
-        // Optional: Reserve the internal vectors based on the number of pm4 packets in the capture
-        // This is an educated guess that each PM4 packet results in x number of associated
-        // field/register nodes. Overguessing means more memory used during creation. Underguessing
-        // means more allocations. For big captures, this is easily in the multi-millions, so
-        // pre-reserving the space is a signficiant performance win
-        uint64_t reserve_size = m_capture_metadata.m_num_pm4_packets * 10;
+    // Optional: Reserve the internal vectors based on the number of pm4 packets in the capture
+    // This is an educated guess that each PM4 packet results in x number of associated
+    // field/register nodes. Overguessing means more memory used during creation. Underguessing
+    // means more allocations. For big captures, this is easily in the multi-millions, so
+    // pre-reserving the space is a signficiant performance win
+    uint64_t reserve_size = m_capture_metadata.m_num_pm4_packets * 10;
 
-        // Command hierarchy tree creation
-        CommandHierarchyCreator cmd_hier_creator(m_capture_metadata.m_command_hierarchy,
-                                                 m_capture_data);
-        if (!cmd_hier_creator.CreateTrees(true, std::make_optional(reserve_size)))
-        {
-            return false;
-        }
+    // Command hierarchy tree creation
+    CommandHierarchyCreator cmd_hier_creator(m_capture_metadata.m_command_hierarchy,
+                                             m_capture_data);
+    if (!cmd_hier_creator.CreateTrees(true, std::make_optional(reserve_size)))
+    {
+        return false;
     }
     return true;
 }


third_party/gfxreconstruct/framework/decode/decoder_util.h
```diff
@@ -49,7 +49,7 @@ bool IsComplete(std::vector<T>& consumers, uint64_t block_index)
     return consumers.empty();
 }
 
-static VkQueue GetDeviceQueue(const encode::VulkanDeviceTable* device_table,
```
wangra-google, line 52: is this from integration of the latest gfxr?

dive_core/gfxr_vulkan_command_hierarchy.h
```diff
@@ -0,0 +1,56 @@
...
+                     std::string &&desc,
+                     char         *metadata_ptr = nullptr,
+                     uint32_t      metadata_size = 0);
+    void     AddChild(CommandHierarchy::TopologyType type,
+                      uint64_t                       node_index,
+                      uint64_t                       child_node_index);
+    bool OnCommand(uint32_t submit_index, DiveAnnotationProcessor::VulkanCommandInfo vk_cmd_info);
+    uint64_t           m_cur_submit_node_index = 0;
+    uint64_t           m_cur_command_buffer_node_index = 0;
+    CommandHierarchy  *m_command_hierarchy_ptr = nullptr;
```
wangra-google, line 49: is it possible to have a ref instead of pointer? otherwise we will need to check if the pointer is nullptr every time we try to use it. (which is currently missing)

dive_core/gfxr_vulkan_command_hierarchy.h
```diff
@@ -0,0 +1,56 @@
...
+    const CaptureData *m_capture_data_ptr = nullptr;
```
wangra-google, line 50: same as above

dive_core/gfxr_vulkan_command_hierarchy.h
```diff
@@ -0,0 +1,56 @@
...
+    // This is a list of child indices per node, ie. topology info
+    // Once parsing is complete, we will create a topology from this
+    DiveVector<DiveVector<uint64_t>> m_node_children[CommandHierarchy::kTopologyTypeCount][2];
```
wangra-google, line 53: can you use an enum or constexpr to define `2` here?

third_party/OpenXR-SDK
wangra-google, line 1: just a general comment: . if it is too complicated to move the submodule updates from this PR, you can keep them in this PR, but it would be clearer to have a separate PR for updating the submodules in the future. :)

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -0,0 +1,263 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+#include "gfxr_vulkan_command_hierarchy.h"
+#include "dive_strings.h"
+
+namespace Dive
+{
+GfxrVulkanCommandHierarchyCreator::GfxrVulkanCommandHierarchyCreator() {}
```
wangra-google, line 19: better to add "//--------------------------------------------------------------------------------------------------" after this func to separate functions from each other.

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -0,0 +1,263 @@
...
+    }
+    return node_index;
+}
+
+//--------------------------------------------------------------------------------------------------
+void GfxrVulkanCommandHierarchyCreator::AddChild(CommandHierarchy::TopologyType type,
+                                                 uint64_t                       node_index,
+                                                 uint64_t                       child_node_index)
+{
+    DIVE_ASSERT(node_index < m_node_children[type][0].size());
```
wangra-google, line 83: i am wondering if we could use enum for the index so that it is clear what list are we trying to add in this case.

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -0,0 +1,263 @@
...
+    m_node_children[type][0][node_index].push_back(child_node_index);
+}
+
+void GfxrVulkanCommandHierarchyCreator::GetArgs(const nlohmann::ordered_json &j,
+                                                uint64_t                      curr_index,
+                                                const std::string            &current_path)
+{
+    if (j.is_object())
```
wangra-google, line 91: nit: i think it is clearer to use something like `json` as the var name, `j` seems a bit too similar as an iterator.

third_party/OpenXR-SDK
GrantComm: Moved the submodule updates to #247

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: Is json_args clearer?

wangra-google: sure, much better, thanks :)

third_party/OpenXR-SDK
wangra-google: great! thanks a lot

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -0,0 +1,263 @@
...
+    else
+    {
+        uint64_t vk_cmd_index = AddNode(NodeType::kGfxrVulkanCommandNode,
+                                        vk_cmd_string_stream.str());
+        GetArgs(vk_cmd_info.GetArgs(), vk_cmd_index, "");
+        AddChild(CommandHierarchy::TopologyType::kSubmitTopology,
+                 m_cur_command_buffer_node_index,
+                 vk_cmd_index);
+    }
+    return true;
```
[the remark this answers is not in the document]
GrantComm: Removed and changed function return type to void.

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: Added enum ChildListType. Is that the kind of enum you were looking for?

dive_core/gfxr_vulkan_command_hierarchy.h
GrantComm: Added kNumChildrenLists and for the gfxr case, only set to 1 since there are no need for a second set of children nodes with more than one parent each as is the case in the CommandHierarchy.

dive_core/data_core.h
```diff
@@ -67,10 +68,10 @@ class DataCore
     CaptureData::LoadResult LoadCaptureData(const char *file_name);
 
     // Parse the capture to generate info that describes the capture
-    bool ParseCaptureData();
+    bool ParseCaptureData(bool is_gfxr_capture = false);
```
wangra-google, line 71: i feel it is better to not have the default value to force the user know explicitly what value is passed in.

dive_core/data_core.cpp
```diff
@@ -49,59 +49,84 @@ CaptureData::LoadResult DataCore::LoadCaptureData(const char *file_name)
...
-bool DataCore::ParseCaptureData()
+bool DataCore::ParseCaptureData(bool is_gfxr_capture)
 {
     if (m_progress_tracker)
     {
         m_progress_tracker->sendMessage("Processing command buffers...");
     }
 
-    if (!CreateMetaData())
+    if (is_gfxr_capture)
```
wangra-google, line 112: i guess this can be simplified as 

```
        if (!CreateMetaData(is_gfxr_capture))
        {
            return false;
        }

        if (!CreateCommandHierarchy(is_gfxr_capture))
        {
            return false;
        }
```

dive_core/gfxr_vulkan_command_hierarchy.h
```diff
@@ -0,0 +1,61 @@
...
+    void     OnGfxrSubmit(uint32_t                                   submit_index,
+                          const DiveAnnotationProcessor::SubmitInfo &submit_info);
+    uint64_t AddNode(NodeType type, std::string &&desc);
+    void     AddChild(CommandHierarchy::TopologyType type,
+                      uint64_t                       node_index,
+                      uint64_t                       child_node_index);
+    void OnCommand(uint32_t submit_index, DiveAnnotationProcessor::VulkanCommandInfo vk_cmd_info);
+    uint64_t          m_cur_submit_node_index = 0;
+    uint64_t          m_cur_command_buffer_node_index = 0;
+    CommandHierarchy &m_command_hierarchy;
```
wangra-google, line 53: Thanks a lot for making the change of replacing pointers to ref in GfxrVulkanCommandHierarchyCreator, can you make similar change to CommandHierarchyCreator? (if you prefer, this can be in a different PR) Thanks :)

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -0,0 +1,252 @@
...
+                GetArgs(val, object_node_index, "");
+            }
+            else if (val.is_array())
+            {
+                uint64_t array_node_index = AddNode(NodeType::kGfxrVulkanCommandArgNode,
+                                                    key.c_str());
+                AddChild(CommandHierarchy::TopologyType::kSubmitTopology,
+                         curr_index,
+                         array_node_index);
+                for (size_t i = 0; i < val.size(); ++i)
```
wangra-google, line 101: can you add some comments on the 3 different cases? thanks

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -0,0 +1,252 @@
...
+            {
+                std::ostringstream vk_cmd_arg_string_stream;
+                vk_cmd_arg_string_stream << element;
+                uint64_t arg_index = AddNode(NodeType::kGfxrVulkanCommandArgNode,
+                                             vk_cmd_arg_string_stream.str());
+                AddChild(CommandHierarchy::TopologyType::kSubmitTopology, curr_index, arg_index);
+            }
+        }
+    }
+    else
```
wangra-google, line 161: what is this 3rd case?

dive_core/gfxr_vulkan_command_hierarchy.h
```diff
@@ -0,0 +1,61 @@
...
+class GfxrVulkanCommandHierarchyCreator : public IEmulateCallbacks
+{
+public:
+    GfxrVulkanCommandHierarchyCreator(CommandHierarchy &command_hierarchy,
+                                      CaptureData      &capture_data);
+
+    bool CreateTrees();
+
+private:
+    enum ChildListType
```
wangra-google, line 33: Thanks for making this change! can you apply the same change (use enum) to CommandHierarchyCreator? (again, you can put the change in a different PR if you feel it is too much for this PR) thanks

dive_core/gfxr_vulkan_command_hierarchy.h
```diff
@@ -0,0 +1,61 @@
...
+                      uint64_t                       node_index,
+                      uint64_t                       child_node_index);
+    void OnCommand(uint32_t submit_index, DiveAnnotationProcessor::VulkanCommandInfo vk_cmd_info);
+    uint64_t          m_cur_submit_node_index = 0;
+    uint64_t          m_cur_command_buffer_node_index = 0;
+    CommandHierarchy &m_command_hierarchy;
+    CaptureData      &m_capture_data;
+    // This is a list of child indices per node, ie. topology info
+    // Once parsing is complete, we will create a topology from this
+    DiveVector<DiveVector<uint64_t>> m_node_children[CommandHierarchy::kTopologyTypeCount]
```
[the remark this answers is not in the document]
GrantComm: Removed.

dive_core/gfxr_vulkan_command_hierarchy.h
GrantComm: Sure! Will put in a separate PR.

dive_core/data_core.h
```diff
@@ -84,7 +85,7 @@ class DataCore
 
 private:
     // Create command hierarchy from the captured data
-    bool CreateCommandHierarchy();
+    bool CreateCommandHierarchy(bool is_gfxr_capture = false);
```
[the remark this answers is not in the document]
GrantComm: Done.

dive_core/data_core.h
GrantComm: Done.

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: Removed.

dive_core/data_core.cpp
```diff
@@ -49,59 +49,84 @@ CaptureData::LoadResult DataCore::LoadCaptureData(const char *file_name)
...
-    CommandHierarchyCreator cmd_hier_creator(*state_tracker);
-    if (!cmd_hier_creator.CreateTrees(&m_capture_metadata.m_command_hierarchy,
-                                      m_capture_data,
-                                      true,
-                                      reserve_size,
-                                      m_log_ptr))
+    if (is_gfxr_capture)
     {
-        return false;
+        GfxrVulkanCommandHierarchyCreator vk_cmd_creator(m_capture_metadata.m_command_hierarchy,
```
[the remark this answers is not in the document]
GrantComm: At the moment, it would be either pm4 or the gfxr vulkan. The check for whether or not it is a gfxr_capture decides which command hierarchy creator to use.

angela28chen:
Some overall structure concerns:
- I remember we talked about this briefly before, but would it be possible to do some kind of restructuring that reflects how different the GFXR and PM4 information is? Maybe we could make CaptureData into a base class and then make separate derived classes for GFXR and PM4 captures, or some other kind of separation?
- I think this would help with:
  - Understanding which members inside CaptureData are PM4-only or GFXR-only or shared.
  - Detangling the loading of GFXR and PM4 files, and establishing what combination of files can be loaded/displayed at the same time.
- I feel like we can also consider separating some parts of DataCore to have different behaviour depending on which capture data object, rather than hinging on passing in a boolean argument "is_gfxr_capture".

I think it's a good opportunity to restructure some of dive_core now rather than directly adding all the GFXR stuff into a system that was originally only designed to fit PM4 data.

wangra-google:
> Some overall structure concerns:
> 
> * I remember we talked about this briefly before, but would it be possible to do some kind of restructuring that reflects how different the GFXR and PM4 information is? Maybe we could make CaptureData into a base class and then make separate derived classes for GFXR and PM4 captures, or some other kind of separation?
> * I think this would help with:
>   
>   * Understanding which members inside CaptureData are PM4-only or GFXR-only or shared.
>   * Detangling the loading of GFXR and PM4 files, and establishing what combination of files can be loaded/displayed at the same time.
> * I feel like we can also consider separating some parts of DataCore to have different behaviour depending on which capture data object, rather than hinging on passing in a boolean argument "is_gfxr_capture".
> 
> I think it's a good opportunity to restructure some of dive_core now rather than directly adding all the GFXR stuff into a system that was originally only designed to fit PM4 data.

I was about to say the same. It a bit messy to mix gfxr and pm4 into 1 class, and it would be nice to split them into 2 classes as Angela suggested. However, i feel that it might be a bigger change which might be better to be done in a different PR. (Grant's another PR of pm4 correlation also has some dependencies on this, so maybe it is better to submit this to make sure the vulkan tree is working with the pm4 correlation functional, then Grant can work on the refactoring if the classes) What do you both think?

GrantComm:
> > Some overall structure concerns:
> > 
> > * I remember we talked about this briefly before, but would it be possible to do some kind of restructuring that reflects how different the GFXR and PM4 information is? Maybe we could make CaptureData into a base class and then make separate derived classes for GFXR and PM4 captures, or some other kind of separation?
> > * I think this would help with:
> >   
> >   * Understanding which members inside CaptureData are PM4-only or GFXR-only or shared.
> >   * Detangling the loading of GFXR and PM4 files, and establishing what combination of files can be loaded/displayed at the same time.
> > * I feel like we can also consider separating some parts of DataCore to have different behaviour depending on which capture data object, rather than hinging on passing in a boolean argument "is_gfxr_capture".
> > 
> > I think it's a good opportunity to restructure some of dive_core now rather than directly adding all the GFXR stuff into a system that was originally only designed to fit PM4 data.
> 
> I was about to say the same. It a bit messy to mix gfxr and pm4 into 1 class, and it would be nice to split them into 2 classes as Angela suggested. However, i feel that it might be a bigger change which might be better to be done in a different PR. (Grant's another PR of pm4 correlation also has some dependencies on this, so maybe it is better to submit this to make sure the vulkan tree is working with the pm4 correlation functional, then Grant can work on the refactoring if the classes) What do you both think?

I think the refactor makes sense as well. I will probably need to restructure some of my UI changes to accompany that refactor too.

dive_core/data_core.cpp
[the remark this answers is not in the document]
wangra-google: as we discussed, this will be refactored in future PR to split pm4 and vulkan into 2 classes

dive_core/gfxr_vulkan_command_hierarchy.cpp
wangra-google: thanks

wangra-google (dismissed):

GrantComm:
> > > > Some overall structure concerns:
> > > > 
> > > > * I remember we talked about this briefly before, but would it be possible to do some kind of restructuring that reflects how different the GFXR and PM4 information is? Maybe we could make CaptureData into a base class and then make separate derived classes for GFXR and PM4 captures, or some other kind of separation?
> > > > * I think this would help with:
> > > >   
> > > >   * Understanding which members inside CaptureData are PM4-only or GFXR-only or shared.
> > > >   * Detangling the loading of GFXR and PM4 files, and establishing what combination of files can be loaded/displayed at the same time.
> > > > * I feel like we can also consider separating some parts of DataCore to have different behaviour depending on which capture data object, rather than hinging on passing in a boolean argument "is_gfxr_capture".
> > > > 
> > > > I think it's a good opportunity to restructure some of dive_core now rather than directly adding all the GFXR stuff into a system that was originally only designed to fit PM4 data.
> > > 
> > > 
> > > I was about to say the same. It a bit messy to mix gfxr and pm4 into 1 class, and it would be nice to split them into 2 classes as Angela suggested. However, i feel that it might be a bigger change which might be better to be done in a different PR. (Grant's another PR of pm4 correlation also has some dependencies on this, so maybe it is better to submit this to make sure the vulkan tree is working with the pm4 correlation functional, then Grant can work on the refactoring if the classes) What do you both think?
> > 
> > 
> > I think the refactor makes sense as well. I will probably need to restructure some of my UI changes to accompany that refactor too.
> 
> OK, it seems like we're on the same page here. I'll leave it up to grant on what order to refactor/implement new things on.

Ok, I have started to refactor and test with UI. Will open up a separate PR for that when it is ready.

dive_core/gfxr_vulkan_command_hierarchy.h
angela28chen: For some issues that will be addressed in future PRs:
1. This 
2. Maybe the gfxr_ext migration
3. Larger refactor with CaptureData

Maybe it would be a good idea to track those with internal bugs? I would find (3) useful as well since some areas of what I'm working on will touch the CaptureData

dive_core/command_hierarchy.h
```diff
@@ -390,6 +399,7 @@ class CommandHierarchyCreator : public IEmulateCallbacks
 {
 public:
     CommandHierarchyCreator(EmulateStateTracker &state_tracker);
+    CommandHierarchyCreator();
```
angela28chen, line 402: I have a bit of trouble understanding why the default constructor is added here, could you elaborate?

dive_core/gfxr_vulkan_command_hierarchy.h
```diff
@@ -0,0 +1,53 @@
...
+*/
+
+// =====================================================================================================================
+// The VulkanCommandHierarchy class parses and creates a tree of Nodes in the command buffer. The
+// primary client for this class is the Model class for the Vulkan command and argument Views in the
+// UI.
+// =====================================================================================================================
+
+#include "dive_core/command_hierarchy.h"
+
```
angela28chen, line 21: nit: I think this should also directly include the header that defines `IEmulateCallbacks`

dive_core/data_core.cpp
```diff
@@ -49,57 +49,72 @@ CaptureData::LoadResult DataCore::LoadCaptureData(const char *file_name)
...
+                                                         m_capture_data);
+        if (!vk_cmd_creator.CreateTrees())
+        {
+            return false;
+        }
+    }
```
angela28chen, line 62: Consider adding `return true;` here for the `is_gfxr_capture` case to early return and avoid having the `else` indented block below

dive_core/common/emulate_pm4.h
```diff
@@ -30,6 +30,7 @@
 #include "adreno.h"
 #include "dive_core/common/pm4_packets/pfp_pm4_packets.h"
 #include "dive_core/stl_replacement.h"
+#include "dive_core/common/dive_annotation_processor.h"
```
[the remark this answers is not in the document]
GrantComm: Done

dive_core/common/emulate_pm4.cpp
```diff
@@ -917,4 +932,28 @@ bool IEmulateCallbacks::ProcessSubmits(const DiveVector<SubmitInfo> &submits,
...
+const IMemoryManager                                                    &mem_manager)
+{
+    for (uint32_t submit_index = 0; submit_index < submits.size(); ++submit_index)
+    {
+        const DiveAnnotationProcessor::SubmitInfo &submit_info = *submits[submit_index];
+
+        OnGfxrSubmit(submit_index, submit_info);
+
+        EmulatePM4                                             emu;
+        DiveVector<DiveAnnotationProcessor::VulkanCommandInfo> vkcmds;
```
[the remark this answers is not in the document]
GrantComm: Removed

dive_core/data_core.cpp
GrantComm: Done

dive_core/gfxr_vulkan_command_hierarchy.h
GrantComm: Done.

dive_core/command_hierarchy.h
GrantComm: Removed. Was left over from previous iteration.

wangra-google (approved):

lrz_validator/CMakeLists.txt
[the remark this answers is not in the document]
angela28chen: I was asking this because I see you added `NOMINMAX` in this PR for a lot of targets outside lrz_validator as well?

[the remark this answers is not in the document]
GrantComm: Oh sorry, yes there are min/max conflicts in the cli and ui target as well.

angela28chen (commented):
Sorry I think I'm getting confused, is this change (Adding GFXR Vulkan Command Hierarchy) blocked by our discussion around whether we want Vulkan and PM4 nodes in the same CommandHierarchy lists, or is that a separate issue that can be addressed in the upcoming changes about adding support for "joint" Command Hierarchies?

GrantComm:
> Sorry I think I'm getting confused, is this change (Adding GFXR Vulkan Command Hierarchy) blocked by our discussion around whether we want Vulkan and PM4 nodes in the same CommandHierarchy lists, or is that a separate issue that can be addressed in the upcoming changes about adding support for "joint" Command Hierarchies?

This change is not necessarily blocked by the Vulkan and PM4 nodes in the same CommandHierarchy lists. There will need to be an update to this once we all agree on the Vulkan and PM4 nodes since the nodes are added to CommandHierarchy's m_nodes member via AddGfxrNodes. If we had the vulkan node vector and pm4 node vector, then the gfxrcommandhierarchy creator would only add nodes to the vulkan node vector.

angela28chen (dismissed):
LGTM but better to hold off on submitting until we make a decision about if Vulkan and PM4 nodes should be in the same vectors or not

GrantComm:
> LGTM but better to hold off on submitting until we make a decision about if Vulkan and PM4 nodes should be in the same vectors or not

Ok sounds good

lrz_validator/CMakeLists.txt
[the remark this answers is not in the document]
angela28chen: I raised this because I was wondering if you set the compile definitions NOMINMAX on the target dive_core then maybe you don't need to change all the CMakeLists.txt files for targets built using dive_core? Or does it not work like that? I vaguely remember something similar happening with where we included psapi lib, the compiler doesn't complain about it missing for dive_core library but gives errors for all the other targets built using it.

angela28chen (dismissed):

wangra-google (approved):

dive_core/command_hierarchy.cpp
```diff
@@ -387,6 +393,16 @@ uint64_t CommandHierarchy::Nodes::AddNode(NodeType type, std::string &&desc, Aux
     return m_node_type.size() - 1;
 }
 
+//--------------------------------------------------------------------------------------------------
+uint64_t CommandHierarchy::Nodes::AddGfxrNode(NodeType type, std::string &&desc)
+{
+    DIVE_ASSERT(m_node_type.size() == m_description.size());
+
+    m_node_type.push_back(type);
+    m_description.push_back(std::move(desc));
```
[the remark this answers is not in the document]
GrantComm: When just loading a gfxr file this would not be needed because the m_aux_info is not used in that context. Would you want this change here or in the pr that introduces the mixed pm4 and gfxr command hierarchy?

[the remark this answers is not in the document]
GrantComm: Ok, added the dummy AuxInfo object and pushed it to the m_aux_info array.

wangra-google (dismissed):

dive_core/capture_data.cpp
```diff
@@ -1227,6 +1240,13 @@ const DiveVector<SubmitInfo> &CaptureData::GetSubmits() const
     return m_submits;
 }
 
+//--------------------------------------------------------------------------------------------------
+const std::vector<std::unique_ptr<DiveAnnotationProcessor::SubmitInfo>> &
+CaptureData::GetGfxrSubmits() const
+{
+    return std::move(m_gfxr_submits);
```
wangra-google, line 1247: why do you use std::move here? isn't `GetGfxrSubmits() const` a read-only access?

GrantComm: Removed.

dive_core/common/emulate_pm4.cpp
```diff
@@ -916,5 +916,4 @@ bool EmulateCallbacksBase::ProcessSubmits(const DiveVector<SubmitInfo> &submits,
     }
     return true;
 }
-
```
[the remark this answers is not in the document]
GrantComm: Undone.

dive_core/capture_data.cpp
```diff
@@ -26,8 +26,10 @@
 #include "dive_core/command_hierarchy.h"
 #include "dive_core/common/common.h"
 #include "freedreno_dev_info.h"
+#include "generated/generated_vulkan_dive_consumer.h"
```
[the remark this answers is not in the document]
GrantComm: Added the full path.

dive_core/common/emulate_pm4.cpp
```diff
@@ -27,6 +27,7 @@
 #include "emulate_pm4.h"
 #include "memory_manager_base.h"
 #include "pm4_info.h"
+#include "gfxr_ext/decode/dive_annotation_processor.h"
```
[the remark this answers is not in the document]
GrantComm: Removed.

dive_core/data_core.cpp
```diff
@@ -40,51 +41,67 @@ CaptureData::LoadResult DataCore::LoadCaptureData(const char *file_name)
...
+        {
+            return false;
+        }
     }
     return true;
 }
 
 //--------------------------------------------------------------------------------------------------
-bool DataCore::CreateMetaData()
+bool DataCore::CreateMetaData(bool is_gfxr_capture)
```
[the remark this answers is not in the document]
GrantComm: Removed the bool parameter and added the check where the function is called.

wangra-google (approved):

PR #256, opened 2025-06-24
Moves dive_annotation processor

GrantComm:
Moves dive_annotation processor from dive_core/common to gfxr_ext/decode.

### DIFF
--- a/gfxr_ext/decode/dive_annotation_processor_test.cpp
+++ b/gfxr_ext/decode/dive_annotation_processor_test.cpp
@@ -41,9 +41,6 @@ MATCHER_P2(SubmitInfoEq, expected_name, expected_command_buffer_count, "")
            arg->GetCommandBufferCount() == expected_command_buffer_count;
 }
 
-class DiveAnnotationProcessorTest : public testing::Test
-{};
-
 TEST(WriteBlockEndTest, SingleSubmitCreatesOneSubmitWithNoCommands)
 {
     DiveAnnotationProcessor processor;


dive_core/generateAdrenoHeader.py
```diff
@@ -474,8 +474,8 @@ def outputPackets(pm4_packet_file_h, registers_et_root, domains):
   pm4_packet_file_h.close();
 
   # lint
-  print("formatting " + "clang-format -i -style=file " + sys.argv[3])
-  os.system("clang-format -i -style=file " + sys.argv[3])
+  print("formatting " + "clang-format-18 -i -style=file " + sys.argv[3])
```
footballhead, line 477: I don't have this in my PATH, I got LLVM 18.1.8 from https://releases.llvm.org/download.html. The build output also looks like it couldn't find this

GrantComm: Ah, sorry! My windows machine wasn't configured properly. Removed.

gfxr_ext/decode/dive_annotation_processor.cpp
```diff
@@ -18,9 +18,9 @@
 #include <cstdint>
 #include <iostream>
 #include <ostream>
-#include "third_party/gfxreconstruct/framework/decode/api_decoder.h"
-#include "third_party/gfxreconstruct/framework/util/logging.h"
-#include "third_party/gfxreconstruct/framework/util/output_stream.h"
+#include "decode/api_decoder.h"
```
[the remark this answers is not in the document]
angela28chen: For files under `gfxr_ext/`, since the library they're built into is then used in gfxrecon replay project, they currently need to be paths relative to the framework dir for it to build. I think it's fine to use relative paths for this PR. When I was setting this up I experimented with full paths but couldn't figure out that part, maybe if someone can figure that out we can change all include paths under gfxr_ext at once?

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+ limitations under the License.
+*/
+
+#include "dive_annotation_processor.h"
+
+#include <gtest/gtest.h>
+
+namespace gfxrecon::decode
+{
+namespace
```
footballhead, line 23: +1 for anonymous namespace. All tests should use one

gfxr_ext/decode/dive_annotation_processor.cpp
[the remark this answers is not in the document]
footballhead: Gotcha. thanks for the explanation

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+    processor_.WriteBlockEnd(submit_data_2);
+
+    auto submits = processor_.getSubmits();
+    ASSERT_EQ(submits.size(), 2);
+    EXPECT_TRUE(AreSubmitInfoEqual(submits[0].get(), "vkQueueSubmit", 1));
+    ASSERT_EQ(submits[0]->GetVulkanCommands().size(), 4);
+    EXPECT_TRUE(AreVulkanCommandInfoEqual(submits[0]->GetVulkanCommands()[1],
+                                          "vkCmdDraw",
+                                          1,
+                                          nlohmann::ordered_json()));
```
angela28chen, line 91: Should the expected parameters here and below be expressed in terms of the DiveFunctionData variables? Such as rather than "vkCmdDispatch" it is `cmd_data_3.GetFunctionName()`?

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "dive_annotation_processor.h"
+
+#include <gtest/gtest.h>
+
+namespace gfxrecon::decode
+{
+namespace
+{
+
+class DiveAnnotationProcessorTestFixture : public testing::Test
```
footballhead, line 26: I don't think you need a fixture for this. The methods can be free functions instead (or matchers if you want to get fancy!). `processor_` can be instantiated in the test body so that it's easier to read.

The preference tends to be to avoid fixtures. https://abseil.io/tips/122

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+                            uint32_t expected_command_buffer_count)
+    {
+        return actual != nullptr && actual->GetSubmitText() == expected_name &&
+               actual->GetCommandBufferCount() == expected_command_buffer_count;
+    }
+};
+
+TEST_F(DiveAnnotationProcessorTestFixture, WriteBlockEndSingleSubmit)
+{
+    gfxrecon::util::DiveFunctionData data("vkQueueSubmit", 0, 1, { { "arg1", 1 } });
```
footballhead, line 53: Since this local is only referenced one, consider inlining it into the call below https://abseil.io/tips/161

footballhead, line 53: Consider function argument comments to give these numbers some more meaning: https://google.github.io/styleguide/cppguide.html#Function_Argument_Comments

```suggestion
    gfxrecon::util::DiveFunctionData data("vkQueueSubmit", /*cmd_buffer_index=*/0, /*block_index=*/1, { { "arg1", 1 } });
```

footballhead, line 53: Are `args` necessary for this test? If not, consider removing them to keep the test more focused
```suggestion
    gfxrecon::util::DiveFunctionData data("vkQueueSubmit", 0, 1, {});
```

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+    processor_.WriteBlockEnd(data);
+
+    auto submits = processor_.getSubmits();
+    ASSERT_EQ(submits.size(), 1);
```
footballhead, line 57: nit: The SizeIs() matcher better expresses intent here

```suggestion
    ASSERT_THAT(submits, SizeIs(1));
```

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+    bool AreSubmitInfoEqual(const DiveAnnotationProcessor::SubmitInfo* actual,
+                            const std::string&                         expected_name,
+                            uint32_t expected_command_buffer_count)
+    {
+        return actual != nullptr && actual->GetSubmitText() == expected_name &&
+               actual->GetCommandBufferCount() == expected_command_buffer_count;
+    }
+};
+
+TEST_F(DiveAnnotationProcessorTestFixture, WriteBlockEndSingleSubmit)
```
footballhead, line 51: nit: consider test names that summarize the behaviour and include the expected outcome so that (a) the code is self-documenting and the reader can more easily understand the behavior and (b) the test remains focused. It's OK for test names to be long since you only need to write them once; their value comes from being read over and over in test output and by future maintainers

gfxr_ext/decode/dive_annotation_processor_test.cpp
GrantComm: Done

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+protected:
+    // Helper function to compare VulkanCommandInfo
+    bool AreVulkanCommandInfoEqual(const DiveAnnotationProcessor::VulkanCommandInfo& actual,
+                                   const std::string&                                expected_name,
+                                   uint32_t                                          expected_index,
+                                   const nlohmann::ordered_json&                     expected_args)
+    {
+        return actual.GetVkCmdName() == expected_name && actual.GetVkCmdIndex() == expected_index &&
+               actual.GetArgs() == expected_args;
+    }
```
footballhead, line 40: consider using a matcher here since it will print out which field did not match. you have two options:

```suggestion
    auto VulkanCommandInfoEqual(const DiveAnnotationProcessor::VulkanCommandInfo& actual,
                                   const std::string&                                expected_name,
                                   uint32_t                                          expected_index,
                                   const nlohmann::ordered_json&                     expected_args)
    {
        return AllOf(Eq(actual.GetVkCmdName(), expected_name), Eq(actual.GetVkCmdIndex(), expected_index) &&
               Eq(actual.GetArgs() , expected_args);
    }
```

or
```suggestion
    MATCHER_P3(VulkanCommandInfoEqual, 
                                   const std::string&                                expected_name,
                                   uint32_t                                          expected_index,
                                   const nlohmann::ordered_json&                     expected_args, "")
    {
        EXPECT_EQ(arg.GetVkCmdName(), expected_name);
        EXPECT_EQ(arg.GetVkCmdIndex(), expected_index);
        EXPECT_EQ(arg.GetArgs(), expected_args);
        return true;
    }
```

then you'd use it like:

```c++
    EXPECT_THAT(submits[0]->GetVulkanCommands()[1], VulkanCommandInfoEqual(
                                          cmd_data_2.GetFunctionName(),
                                          /*expected_index=*/1,
                                          nlohmann::ordered_json()));
```

(https://google.github.io/googletest/gmock_cook_book.html#extending-gmock)

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+                            uint32_t expected_command_buffer_count)
+    {
+        return actual != nullptr && actual->GetSubmitText() == expected_name &&
+               actual->GetCommandBufferCount() == expected_command_buffer_count;
+    }
+};
+
+TEST_F(DiveAnnotationProcessorTest, WriteBlockEnd_SingleSubmit_CreatesOneSubmitWithNoCommands)
+{
+    DiveAnnotationProcessor processor_;
```
footballhead, line 54: nit: don't need the trailing underscore, it doesn't add any readability value as a local variable

```suggestion
    DiveAnnotationProcessor processor;
```

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+    processor_.WriteBlockEnd(gfxrecon::util::DiveFunctionData("vkQueueSubmit",
+                                                              /*cmd_buffer_index=*/0,
+                                                              /*block_index=*/1,
+                                                              {}));
+
+    auto submits = processor_.getSubmits();
+    ASSERT_THAT(submits, SizeIs(1));
+    EXPECT_TRUE(AreSubmitInfoEqual(submits[0].get(), "vkQueueSubmit", 0));
+    EXPECT_TRUE(submits[0]->GetVulkanCommands().empty());
```
footballhead, line 63: Can use a more specific matcher here

```suggestion
    EXPECT_THAT(submits[0]->GetVulkanCommands(), IsEmpty());
```

The goal is to make these are close to readable english statements

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+    bool AreSubmitInfoEqual(const DiveAnnotationProcessor::SubmitInfo* actual,
+                            const std::string&                         expected_name,
+                            uint32_t expected_command_buffer_count)
+    {
+        return actual != nullptr && actual->GetSubmitText() == expected_name &&
+               actual->GetCommandBufferCount() == expected_command_buffer_count;
+    }
+};
+
+TEST_F(DiveAnnotationProcessorTest, WriteBlockEnd_SingleSubmit_CreatesOneSubmitWithNoCommands)
```
footballhead, line 52: gtests asks for tests names to avoid underscores: https://google.github.io/googletest/primer.html#simple-tests

In this case there are two schools of thought:

1. "DiveAnnotationProcessorTest" is redundant since it's in the file name (dive_annotation_processor_test.cpp). So the suite name should be the name of the function being tested:

```suggestion
TEST_F(WriteBlockEndTest, SingleSubmitCreatesOneSubmitWithNoCommands)
```

2. Using the function name as the suite name is bad since it ignores the fact that often we don't just test one function

```suggestion
TEST_F(DiveAnnotationProcessorTest, CreatesOneSubmitWithNoCommandsAfterWriteBlockEndWithSingleSubmit)
```

Pick whichever one makes you happy :)

PR #257, opened 2025-06-25
Remove the requirement for root access for gfxr capture

RenfengLiu:
root access is only needed for PM4 capture for now. In the future if we transit fully to the capture PM4 on replay mode, we can remove the requirement of needing root access.


wangra-google (approved):

footballhead (approved):
Thanks for doing this!

PR #258, opened 2025-06-27
Improve reliability of replay-with-dump script

footballhead:
A bunch of small improvements to replay-with-dump.sh which makes it more reliable and ergonomic:

- Make GFXR mandatory and GFXA optional. GFXA is not required for all captures (see the ones in //tests/gfxr_captures)
- Don't assume the replay APK has been installed; install it!
- Uninstall replay first to avoid potential cert mismatch
- Ensure replay app has permissions to use `/sdcard/Download`. Instead of requiring permissions, I tested `/data/local/tmp`. However, not all devices can access that directory. `/sdcard/Download` with external storage permissions proved to be more consistently reliable on my test devices.
- Wait a little bit for replay APK to launch so that `pidof` can track when it's done. There's still the corner case where the app takes >5s to launch (e.g. the screen is off) but that can't be solved without modifying gfxrecon.py to use `am start -W` (and even that has an internal timeout)
- Clean up `$PUSH_DIR` to avoid clogging up the device with temp files.

While these all could individually been their own PR, they needed to be tested together. So it made sense to commit them together.

### DIFF
--- a/scripts/replay-with-dump.bat
+++ b/scripts/replay-with-dump.bat
@@ -18,7 +18,7 @@
 setlocal
 
 IF "%~1"=="" (
-    echo Usage: replay-with-dump.bat GFXR GFXA
+    echo Usage: replay-with-dump.bat GFXR [GFXA]
     exit /b 1
 )
 


wangra-google:
> A bunch of small improvements to replay-with-dump.sh which makes it more reliable and ergonomic:
> 
> * Make GFXR mandatory and GFXA optional. GFXA is not required for all captures (see the ones in //tests/gfxr_captures)
> * Don't assume the replay APK has been installed; install it!
> * Ensure replay app has permissions to use `/sdcard/Download`. Instead of requiring permissions, I tested `/data/local/tmp`. However, not all devices can access that directory. `/sdcard/Download` with external storage permissions proved to be more consistently reliable on my test devices.
> * Wait a little bit for replay APK to launch so that `pidof` can track when it's done. There's still the corner case where the app takes >5s to launch (e.g. the screen is off) but that can't be solved without modifying gfxrecon.py to use `am start -W` (and even that has an internal timeout)
> * Clean up `$PUSH_DIR` to avoid clogging up the device with temp files.
> 
> While these all could individually been their own PR, they needed to be tested together. So it made sense to commit them together.

Thanks for making the change. Could you apply the same changes to the Windows bat file as well so that both scripts are consistent?

Also I am thinking maybe it is better to move those 2 scripts into the script folder so that all scripts are at the same location? What do you think?

footballhead:
> Also I am thinking maybe it is better to move those 2 scripts into the script folder so that all scripts are at the same location? What do you think?

Make sense to me. I've moved them in https://github.com/google/dive/pull/264 and will rebase this onto that once it's in.

wangra-google:
> > Also I am thinking maybe it is better to move those 2 scripts into the script folder so that all scripts are at the same location? What do you think?
> 
> Make sense to me. I've moved them in #264 and will rebase this onto that once it's in.

Thanks :)

wangra-google (dismissed):

wangra-google (dismissed):

wangra-google (dismissed):

wangra-google (approved):

footballhead:
> Thanks for making the change. Could you apply the same changes to the Windows bat file as well so that both scripts are consistent?

Done 

> Also I am thinking maybe it is better to move those 2 scripts into the script folder so that all scripts are at the same location? What do you think?

done

wangra-google:
> > Thanks for making the change. Could you apply the same changes to the Windows bat file as well so that both scripts are consistent?
> 
> Done
> 
> > Also I am thinking maybe it is better to move those 2 scripts into the script folder so that all scripts are at the same location? What do you think?
> 
> done

Thanks!

PR #268, opened 2025-07-04
Ask the device for architecture if not given

footballhead:
The device knows its architecture. If an architecture isn't specified, ask the device for it. This is more ergonomic (saves a lot of typing).


angela28chen (approved):
Thanks! This seems more convenient

wangra-google (approved):

PR #269, opened 2025-07-08
Add GPU time support for Replay

wangra-google:
- Add gpu time support for Replay. Note that the frame gpu time is calculated by accumulating cmd gpu time which might not be correct since cmds might overlapping with each other. (We will measure cmd/renderpass gpu time separately so that we could associate them with vulkan calls)
- Extract gpu time code from the runtime layer, and share the code with Replay
- Minor fix in application.h in gfxr code to fix a warning


third_party/gfxreconstruct/framework/application/application.h
```diff
@@ -84,7 +84,9 @@ class Application final
 
     void StopRunning() { running_ = false; }
 
-    uint32_t GetCurrentFrameNumber() const
+    // GOOGLE: fix a warning of possible loss of data
+    // file_processor_->GetCurrentFrameNumber() returns a uint64_t 
+    uint64_t GetCurrentFrameNumber() const
```
footballhead, line 89: In the spirit of small CLs, this could have been a separate CL

gpu_time/gpu_time.h
```diff
@@ -0,0 +1,147 @@
...
+#include <string>
+#include <vector>
+#include <deque>
+#include <unordered_map>
+#include <limits>
+
+namespace Dive
+{
+
+class GPUTime
```
footballhead, line 30: In the spirit of small CLs, this refactor could have been a separate CL

gfxr_ext/decode/dive_vulkan_replay_consumer.h
```diff
@@ -0,0 +1,121 @@
...
+
+#include "generated/generated_vulkan_replay_consumer.h"
+#include "gpu_time.h"
+#include <set>
+#include <vector>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+class DiveVulkanReplayConsumer : public VulkanReplayConsumer
```
footballhead, line 28: In the sprit of small CLs, adding this class could have been a separate CL

PR #278, opened 2025-07-10
Use Abseil flags in gfxr_dump_resources

footballhead:
I want to add a flag to gfxr_dump_resources so this PR introduces the dependency and some scaffolding. The actual flag will come in a later PR.

I renamed main.cpp to gfxr_dump_resources_main.cpp since `--help` filters for flags based on very specific file names. See `ContainsHelppackageFlags()`


angela28chen:
> I renamed main.cpp to gfxr_dump_resources_main.cpp since --help filters for flags based on very specific file names. See ContainsHelppackageFlags()

I ran into absl flags not showing up in "--help" but I never realized this is a matter of how the main cpp file is named. Good to know!

PR #281, opened 2025-07-11
Dive communication protocol: Adds unit tests to TcpClient and UnixDomainServer

elviscapiaq:
- The tests assume the client and server are running on a PC (Linux), as we are not using an Android device. However, this creates a limitation: we cannot directly test the TcpClient against the UnixDomainServer because they use incompatible communication protocols. The TcpClient uses TCP/IP sockets, while the UnixDomainServer uses Unix domain sockets (UDS).
- To test the UnixDomainServer, we will start an instance of it and use a basic, raw socket client to verify its behavior, such as handling connections and basic messages.
- To test the TcpClient, we create a minimal TCP server.
- Adds `network` namespace to `messages_test.cc` to simplify and aligned with `tcp_client_test.cc` and `unix_domain_server_test.cc`. 

**Note:** The unit tests do not work on Windows because the UnixDomainServer listens only on UNIX OS (Android/Linux).

### DIFF
--- a/network/CMakeLists.txt
+++ b/network/CMakeLists.txt
@@ -81,7 +81,7 @@ if(NOT ANDROID)
       absl::statusor
     )
     add_test(NAME unix_domain_server_test COMMAND unix_domain_server_test)
-    set_tests_properties(unix_domain_server_test PROPERTIES TIMEOUT 2)
+    set_tests_properties(unix_domain_server_test PROPERTIES TIMEOUT 5)
 
     add_executable(tcp_client_test tcp_client_test.cc)
     target_link_libraries(tcp_client_test PRIVATE
@@ -92,6 +92,6 @@ if(NOT ANDROID)
       absl::statusor
     )
     add_test(NAME tcp_client_test COMMAND tcp_client_test)
-    set_tests_properties(tcp_client_test PROPERTIES TIMEOUT 2)
+    set_tests_properties(tcp_client_test PROPERTIES TIMEOUT 5)
   endif()
 endif()
--- a/network/messages_test.cc
+++ b/network/messages_test.cc
@@ -75,11 +75,7 @@ TEST(MessagesTest, WriteAndReadString)
 
 TEST(MessagesTest, SerializeAndDeserializeHandShakeMessage)
 {
-<<<<<<< HEAD
-    Network::HandshakeRequest request;
-=======
-    HandShakeRequest  request;
->>>>>>> fb7b93d0 (Dive communication protocol: Adds unit tests to TcpClient and UnixDomainServer)
+    HandshakeRequest request;
     request.SetMajorVersion(345612);
     request.SetMinorVersion(567348);
     Buffer buf;
@@ -87,11 +83,7 @@ TEST(MessagesTest, SerializeAndDeserializeHandShakeMessage)
     ASSERT_TRUE(status.ok());
     ASSERT_EQ(request.GetMessageType(), MessageType::HANDSHAKE_REQUEST);
 
-<<<<<<< HEAD
-    Network::HandshakeResponse response;
-=======
     HandShakeResponse response;
->>>>>>> fb7b93d0 (Dive communication protocol: Adds unit tests to TcpClient and UnixDomainServer)
     status = response.Deserialize(buf);
     ASSERT_TRUE(status.ok());
     ASSERT_EQ(response.GetMessageType(), MessageType::HANDSHAKE_RESPONSE);


network/messages_test.cc
```diff
@@ -19,138 +19,140 @@ limitations under the License.
...
     ASSERT_TRUE(status.ok());
-    ASSERT_EQ(res_serialize.GetMessageType(), Network::MessageType::DOWNLOAD_FILE_RESPONSE);
-    Network::DownloadFileResponse res_deserialize;
+    ASSERT_EQ(res_serialize.GetMessageType(), MessageType::DOWNLOAD_FILE_RESPONSE);
+    DownloadFileResponse res_deserialize;
     status = res_deserialize.Deserialize(buf);
     ASSERT_TRUE(status.ok());
-    ASSERT_EQ(res_deserialize.GetMessageType(), Network::MessageType::DOWNLOAD_FILE_RESPONSE);
+    ASSERT_EQ(res_deserialize.GetMessageType(), MessageType::DOWNLOAD_FILE_RESPONSE);
     ASSERT_EQ(res_serialize.GetFound(), res_deserialize.GetFound());
```
wangra-google, line 156: curious would it be better to use EXPECT_ for GetFound()/GetErrorReason()/GetFileSizeStr()?

network/tcp_client_test.cc
```diff
@@ -0,0 +1,281 @@
...
+};
+
+TEST_F(TcpClientTest, IsConnected)
+{
+    EXPECT_TRUE(client.IsConnected());
+    client.Disconnect();
+    EXPECT_FALSE(client.IsConnected());
+}
+
+TEST_F(TcpClientTest, ConnectionFails)
```
wangra-google, line 233: does this test need the server/client connection code in the Setup/TearDown?

network/tcp_client_test.cc
```diff
@@ -0,0 +1,281 @@
...
+#include <gtest/gtest.h>
+#include <filesystem>
+#include <fstream>
+
+namespace Network
+{
+namespace
+{
+
+const std::string kTestHost = "127.0.0.1";
```
footballhead, line 29: Static variables with non-trivial destructors (like `std::string`) can be dangerous since they may still be used after they are destroyed. See https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables for more info. Here's some more information from Abseil: https://abseil.io/tips/140#approaches-that-work-in-source-files

Here, a better option would be `std::string_view`.  Also, consider `constexpr`.

```suggestion
constexpr std::string_view kTestHost = "127.0.0.1";
```

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,145 @@
...
+        {
+            return connection.status();
+        }
+        return *std::move(connection);
+    }
+
+    std::unique_ptr<UnixDomainServer> server;
+};
+
+TEST_F(UnixDomainServerTest, StartAndStop)
```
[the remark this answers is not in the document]
elviscapiaq: Thanks for pointing it out. Updated!

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,145 @@
...
+
+    // Verify response.
+    ASSERT_EQ((*response_msg)->GetMessageType(), MessageType::HANDSHAKE_RESPONSE);
+    auto* response = dynamic_cast<HandShakeResponse*>((*response_msg).get());
+    ASSERT_NE(response, nullptr);
+    EXPECT_EQ(response->GetMajorVersion(), request.GetMajorVersion());
+    EXPECT_EQ(response->GetMinorVersion(), request.GetMinorVersion());
+}
+
+TEST_F(UnixDomainServerTest, PingPong)
```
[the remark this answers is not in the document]
elviscapiaq: Done.

network/messages_test.cc
```diff
@@ -19,173 +19,176 @@ limitations under the License.
...
+    Buffer         buf;
+    const uint32_t write_value_1 = 123456703;
+    WriteUint32ToBuffer(write_value_1, buf);
     size_t offset = 0;
-    auto   read_value = Network::ReadUint32FromBuffer(buf, offset);
-    ASSERT_TRUE(read_value.ok());
-    ASSERT_EQ(write_value, *read_value);
+    auto   read_value_1 = ReadUint32FromBuffer(buf, offset);
+    ASSERT_TRUE(read_value_1.ok());
+    ASSERT_EQ(write_value_1, *read_value_1);
```
footballhead, line 35: Consider using status_matchers like #285.

```suggestion
    ASSERT_THAT(Network::ReadUint32FromBuffer(buf, offset), IsOkAndHolds(write_value_1));
```

It's more succinct, readable, and produces better output. Before

```
[ RUN      ] MessagesTest.WriteAndReadUint32
/usr/local/google/home/hitchens/git/dive/network/messages_test.cc:35: Failure
Value of: read_value.ok()
  Actual: false
Expected: true

[  FAILED  ] MessagesTest.WriteAndReadUint32 (0 ms)
```

After:

```
[ RUN      ] MessagesTest.WriteAndReadUint32
/usr/local/google/home/hitchens/git/dive/network/messages_test.cc:34: Failure
Value of: Network::ReadUint32FromBuffer(buf, offset)
Expected: is OK and has a value that is equal to 123456703
  Actual: ( INVALID_ARGUMENT: Buffer too small to read a uint32_t. ), which has status INVALID_ARGUMENT: Buffer too small to read a uint32_t.

[  FAILED  ] MessagesTest.WriteAndReadUint32 (0 ms)
```

network/CMakeLists.txt
```diff
@@ -68,4 +69,28 @@ if(NOT ANDROID)
...
+  if(NOT WIN32)
+    add_executable(unix_domain_server_test unix_domain_server_test.cc)
+    target_link_libraries(unix_domain_server_test PRIVATE
+      network
+      gtest
+      gtest_main
+      absl::status
+      absl::statusor
+    )
+    add_test(NAME unix_domain_server_test COMMAND unix_domain_server_test)
```
footballhead, line 82: Since gtest_discover_tests is [intended to replace add_test](https://cmake.org/cmake/help/latest/module/GoogleTest.html), I'd suggest we use that here. Unless we have a reason, at which point I'd appreciate a comment.

PR #285, opened 2025-07-16
Build Abseil TESTONLY libs, e.g. status_matchers

footballhead:
Now that we have a more up-to-date Abseil, we have access to status_matchers. status_matchers is very useful for writing succinct, readable tests.

I've modified one of the existing tests to demonstrate its use and value. Example output before:

```
[ RUN      ] MessagesTest.WriteAndReadUint32
/usr/local/google/home/hitchens/git/dive/network/messages_test.cc:35: Failure
Value of: read_value.ok()
  Actual: false
Expected: true

[  FAILED  ] MessagesTest.WriteAndReadUint32 (0 ms)
```

After:

```
[ RUN      ] MessagesTest.WriteAndReadUint32
/usr/local/google/home/hitchens/git/dive/network/messages_test.cc:34: Failure
Value of: Network::ReadUint32FromBuffer(buf, offset)
Expected: is OK and has a value that is equal to 123456703
  Actual: ( INVALID_ARGUMENT: Buffer too small to read a uint32_t. ), which has status INVALID_ARGUMENT: Buffer too small to read a uint32_t.

[  FAILED  ] MessagesTest.WriteAndReadUint32 (0 ms)
```

The suggested change includes:

- the function that failed
- the expected return value of that function
- the actual return value with lots of detail

### DIFF
--- a/network/messages_test.cc
+++ b/network/messages_test.cc
@@ -23,7 +23,7 @@ limitations under the License.
 namespace
 {
 
-using absl_testing::IsOkAndHolds;
+using ::absl_testing::IsOkAndHolds;
 
 TEST(MessagesTest, WriteAndReadUint32)
 {


footballhead:
I've added @elviscapiaq more as an FYI

angela28chen (approved):

PR #287, opened 2025-07-18
End support for perfetto

angela28chen:
Removing perfetto code and submodule to reduce maintenance burden now that we have decided not to support perfetto traces.


footballhead (approved):

angela28chen:
waiting for @wangra-google approval before submitting, just in case there is a use for the perfetto code

PR #300, opened 2025-07-28
use Dive::GPUTime for runtime layer to measure gpu time

wangra-google:


GrantComm (approved):

shanminchao:
I wasn't involved in the previous PR related to this, so I need a little context here.
Is this just refactoring a previously implemented vulkan layer that measures timing to use a new Dive gputimer class?
Is there any functional difference resulting from this change?
What is the ultimate goal of these changes?

wangra-google:
> I wasn't involved in the previous PR related to this, so I need a little context here. Is this just refactoring a previously implemented vulkan layer that measures timing to use a new Dive gputimer class? Is there any functional difference resulting from this change? What is the ultimate goal of these changes?

just refactoring to move the common code to gpu_time classes so that it can be shared by runtime layer and replay apk since the time measuring code is the same for both

shanminchao:
> > I wasn't involved in the previous PR related to this, so I need a little context here. Is this just refactoring a previously implemented vulkan layer that measures timing to use a new Dive gputimer class? Is there any functional difference resulting from this change? What is the ultimate goal of these changes?
> 
> just refactoring to move the common code to gpu_time classes so that it can be shared by runtime layer and replay apk since the time measuring code is the same for both

Gotcha!

shanminchao (approved):

PR #303, opened 2025-07-30
use Dive::GPUTime for replay to measure gpu time

wangra-google:
using Dive::GPUTime for replay to measure gpu time.
- currently the gpu time measurement is always enabled during replay. I will be adding a flag in the gputime class to disable the measurement by default in a separate PR.

### DIFF
--- a/capture_service/dive_client_cli.cc
+++ b/capture_service/dive_client_cli.cc
@@ -372,54 +372,17 @@ absl::Status is_capture_directory_busy(Dive::DeviceManager& mgr,
 bool retrieve_gfxr_capture(Dive::DeviceManager& mgr, const std::string& gfxr_capture_directory)
 {
     std::filesystem::path download_dir = absl::GetFlag(FLAGS_download_dir);
-    std::filesystem::path target_download_dir(
-    absl::StrCat(download_dir.string(), "/", gfxr_capture_directory));
-    std::filesystem::path on_device_capture_directory = absl::StrCat(Dive::kDeviceCapturePath,
-                                                                     "/",
-                                                                     gfxr_capture_directory);
+
+    // Need to explicitly use forward slash so that this works on Windows targetting Android
+    std::string on_device_capture_directory = absl::StrCat(Dive::kDeviceCapturePath,
+                                                           "/",
+                                                           gfxr_capture_directory);
 
     std::cout << "Retrieving capture..." << std::endl;
-    // Check if the target directory already exists on the local machine.
-    if (!std::filesystem::exists(target_download_dir))
-    {
 
-        std::error_code ec;
-        if (!std::filesystem::create_directories(target_download_dir, ec))
-        {
-            std::cout << "Error creating directory: " << ec << std::endl;
-            return false;
-        }
-    }
-    else
-    {
-        // If the target directory already exists on the local machine, append a number to it to
-        // differentiate.
-        int                   counter = 1;
-        std::filesystem::path newDirPath;
-        while (true)
-        {
-            newDirPath = std::filesystem::path(target_download_dir.generic_string() + "_" +
-                                               std::to_string(counter));
-            if (!std::filesystem::exists(newDirPath))
-            {
-                std::error_code ec;
-
-                if (!std::filesystem::create_directories(newDirPath, ec))
-                {
-                    std::cout << "Error creating directory: " << ec << std::endl;
-                    return false;
-                }
-                target_download_dir = newDirPath;
-                break;
-            }
-            counter++;
-        }
-    }
-
-    // Retrieve the names of the files in the capture directory on the device.
-    std::string                 command = "shell ls " + on_device_capture_directory.string();
+    // Retrieve the list of files in the capture directory on the device.
+    std::string command = absl::StrFormat("shell ls %s", on_device_capture_directory);
     absl::StatusOr<std::string> output = mgr.GetDevice()->Adb().RunAndGetResult(command);
-
     if (!output.ok())
     {
         std::cout << "Error getting capture_file name: " << output.status().message() << std::endl;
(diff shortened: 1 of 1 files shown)


gfxr_ext/decode/CMakeLists.txt
```diff
@@ -44,7 +46,9 @@ if (WIN32)
   target_link_libraries(gfxr_decode_ext_lib psapi)
 endif()
 target_include_directories(gfxr_decode_ext_lib PUBLIC
-  ${CMAKE_CURRENT_SOURCE_DIR}/../../third_party/gfxreconstruct/framework)
+  ${CMAKE_CURRENT_SOURCE_DIR}/../../third_party/gfxreconstruct/framework
+  ${CMAKE_CURRENT_SOURCE_DIR}/../../gpu_time
```
angela28chen, line 50: If you're adding a dependency of gfxr_decode_ext_lib on gpu_time, maybe instead of `<DIVE_ROOT>/gpu_time` those files should be under `<DIVE_ROOT>/gfxr_ext/decode` instead, in the same way we have arranged files like `dive_block_data.cpp`? It can also be a separate library if necessary, which is then depended on by `gfxr_decode_ext_lib`. I think this would be more consistent and simplify this PR's changes to `third_party/gfxreconstruct`. The reason we set up `gfxr_ext` is to better isolate which parts of dive libraries are used in GFXR tools and to keep our build dependency graphs more simple if possible.

angela28chen: We could keep gpu_time still as its own library but have it under `<DIVE_ROOT>/gfxr_ext`?

Dependencies:
```
gpu_time <- gfxr_decode_ext
gpu_time <- VkLayer_rt_dive
```
I think this would still minimize changes under third_party/gfxreconstruct. It looks like gpu_time is used wherever gfxr_decode_ext_lib is used.

wangra-google: VkLayer_rt_dive is not part of gfxr_ext, and gpu_time is more like a util class used by gfxr_decode_ext, so i feel it is a bit misleading to put it under gfxr_ext. gfxr_ext is for code that specifically extends GFXReconstruct, while gpu_time is a more general utility used by multiple parts of the project. If we put gpu_time under gfxr_ext, it suggests that gpu_time is only relevant to the GFXReconstruct extension, which is false. VkLayer_rt_dive also depends on gpu_time, and this project is not part of gfxr_ext. This could confuse other developers and make it harder to find the code or understand its purpose.

angela28chen: I am not suggesting to put VkLayer_rt_dive under gfxr_ext, but I am suggesting to put gpu_time under there because it is a util class that is used by gfxr_decode_ext. I think avoiding any dependencies of gfxr_decode_ext on other Dive code is the reason we created the gfxr_ext dir, but dependencies of other dive code on gfxr_ext is fine and is already happening via dive_core, and it just can happen more explicitly with VkLayer_rt_dive. My impression is that stuff under gfxr_ext should not be viewed as "code only useful for gfxr" but rather "we centralize all the dependencies third_party/gfxreconstruct has on dive code within this folder"

wangra-google: yes, i understand you would prefer moving gpu_time into gfxr_ext (not VkLayer_rt_dive under gfxr_ext). And, you would like to centralize all dependencies that GFXR has on Dive code within the gfxr_ext folder. While centralizing dependencies can be helpful for a clean build process, it often comes at the cost of clear project structure and reusability. Putting gpu_time inside gfxr_ext creates a hidden dependency: VkLayer_rt_dive now depends on a library that appears to be specific to GFXR. This can cause confusion and make the code harder to navigate.

dive_core has some dependencies of gfxr_decode_ext which is fine since dive_core is using features from gfxr_decode_ext. gpu_time is an independent util lib which is not part of gfxr. VkLayer_rt_dive also doesn't have any dependency on gfxr. putting gpu_time under gfxr_ext would limit our project's future potential. For example, if we wanted to share VkLayer_rt_dive as a standalone project, we could easily do so with our current setup. It only depends on the independent gpu_time library. If gpu_time were tied to gfxr_ext, we would have to restructure the project just to share a single component.

I feel that keeping gpu_time as a separate, top-level library gives us the best of both worlds: a clear and logical project structure that is also flexible for future development. :)

angela28chen: OK, thanks for explaining your thoughts behind the choice.

gfxr_ext/decode/dive_vulkan_replay_consumer.h
```diff
@@ -0,0 +1,119 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef GFXRECON_DECODE_VULKAN_DIVE_CONSUMER_H
+#define GFXRECON_DECODE_VULKAN_DIVE_CONSUMER_H
+
+#include "generated/generated_vulkan_replay_consumer.h"
+#include "gpu_time.h"
```
angela28chen, line 21: Should we consider using full path include here?

gfxr_ext/decode/CMakeLists.txt
```diff
@@ -35,6 +35,8 @@ add_library(gfxr_decode_ext_lib
   dive_block_data.cpp
   dive_file_processor.h
   dive_file_processor.cpp
+  dive_vulkan_replay_consumer.h
+  dive_vulkan_replay_consumer.cpp
 )
 target_link_libraries(gfxr_decode_ext_lib
   gfxrecon_decode
```
angela28chen, line 42: Can gpu_time be linked to the gfxr_decode_ext_lib target here so that it doesn't have to be linked inside all the gfxreconstruct CMakeLists.txt?

gfxr_ext/decode/dive_vulkan_replay_consumer.h
wangra-google: sure, added the full path

gfxr_ext/decode/CMakeLists.txt
wangra-google: Thanks for the suggestion. That was actually the original version i did, it is not perfect since the gradle build of the gfx_replay bypasses the root add_subdirectory(gpu_time), i had to do some trick in the cmake in gfxr_ext/decode which looks a bit odd. but comparing the 2 solutions, not modifying all the gfxreconstruct CMakeLists.txt is probably cleaner, so keep this solution.

gfxr_ext/decode/dive_vulkan_replay_consumer.h
```diff
@@ -0,0 +1,119 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef GFXRECON_DECODE_VULKAN_DIVE_CONSUMER_H
+#define GFXRECON_DECODE_VULKAN_DIVE_CONSUMER_H
+
+#include "generated/generated_vulkan_replay_consumer.h"
+#include "../../gpu_time/gpu_time.h"
```
angela28chen, line 21: I think it might be better to do this for consistency:
```
#include "gpu_time/gpu_time.h"
```
CMakeLists.txt
```
target_include_directories(gfxr_decode_ext_lib PUBLIC
  ${CMAKE_CURRENT_SOURCE_DIR}/../../third_party/gfxreconstruct/framework
${CMAKE_CURRENT_SOURCE_DIR}/../../
)
```

angela28chen: Or actually `target_include_directories(gfxr_decode_ext_lib ../..)` might just work, I'm not sure

wangra-google: done

GrantComm (approved):

third_party/gfxreconstruct/tools/replay/android_main.cpp
```diff
@@ -167,7 +170,8 @@ void android_main(struct android_app* app)
                                                       replay_options.block_index_from,
                                                       replay_options.block_index_to);
 
-                gfxrecon::decode::VulkanReplayConsumer vulkan_replay_consumer(application, replay_options);
+                // GOOGLE: replace VulkanReplayConsumer with dive specific DiveVulkanReplayConsumer
+                gfxrecon::decode::DiveVulkanReplayConsumer vulkan_replay_consumer(application, replay_options);
```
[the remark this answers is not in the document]
wangra-google: the replay consumer is in our code base (gfxr_ext/decode/dive_vulkan_replay_consumer.h/cpp), this is the actual apk main.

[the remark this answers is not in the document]
shanminchao: Ok sounds good, thanks for the info!

PR #304, opened 2025-07-30
Produce error during GFXR capture retrieval if the on-device directory is empty indicating that capture has failed

angela28chen:


angela28chen:
I've tested this on windows and linux.

Note that this doesn't fix the issue where if you have 2 or more captures in a single session, they will be pulled multiple times because every "retrieval" is seeking to create a new folder. I suggest to fix this we should be looking at making the captures in different folders on the device.

Rather than multiple captures under `/sdcard/Download/gfxr_capture/` we can consider `/sdcard/Download/gfxr_capture/gfxr_capture_0` where each subfolder holds one pair of gfxr and gfxa files.

wangra-google (approved):

PR #305, opened 2025-07-31
New trace_stats CLI tool

shanminchao:
This represents the initial set of stats
Additional stats to be added in the future

image

### DIFF
--- a/trace_stats/main.cpp
+++ b/trace_stats/main.cpp
@@ -81,6 +81,13 @@
                                                    (uint64_t)0);                          \
     }
 
+#define GATHER_RESOLVES(type)            \
+    do                                   \
+    {                                    \
+        stats_list[kTotalResolves]++;    \
+        stats_list[k##type##Resolves]++; \
+    } while (0)
+
 void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &ostream)
 {
     // Number of draw calls in BINNING, DIRECT, and TILED
@@ -107,7 +114,6 @@ void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &o
         kDirectDraws,
         kTiledDraws,
         kDispatches,
-        kResolves,
         kSyncs,
         kDepthTestEnabled,
         kDepthWriteEnabled,
@@ -133,6 +139,10 @@ void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &o
         kMinGPRs,
         kMaxGPRs,
         kMedianGPRs,
+        kTotalResolves,
+        kSysmemToGmemResolves,
+        kGmemToSysmemResolves,
+        kClearGmemResolves,
         kNumStats
     };
     const char *stats_desc[kNumStats] = {
@@ -140,7 +150,6 @@ void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &o
         "Num Draws (DIRECT)",
         "Num Draws (TILED)",
         "Num Dispatches",
-        "Num Resolves",
         "Num Syncs",
         "Num Draws with Depth Test Enabled",
         "Num Draws with Depth Write Enabled",
@@ -166,6 +175,10 @@ void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &o
         "\tMin GPRs in a single shader",
         "\tMax GPRs in a single shader",
         "\tMedian GPRs in a single shader",
+        "Total resolves",
+        "\tSysMem to Gmem Resolves",
+        "\tGmem to SysMem Resolves",
+        "\tGmem Clears",
     };
 #ifndef NDEBUG
     // Ensure all elements of stats_desc are initialized
@@ -240,8 +253,6 @@ void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &o
         }
         else if (info.m_type == Dive::EventInfo::EventType::kDispatch)
             stats_list[kDispatches]++;
-        else if (info.m_type == Dive::EventInfo::EventType::kResolve)
-            stats_list[kResolves]++;
         else if (info.m_type == Dive::EventInfo::EventType::kSync)
             stats_list[kSyncs]++;
 
@@ -294,6 +305,13 @@ void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &o
             window_scissor.m_br_y = event_state_it->WindowScissorBRY();
             window_scissors.insert(window_scissor);
         }
+
+        if (info.m_type == Dive::EventInfo::EventType::kSysmemToGmemResolve)
+            GATHER_RESOLVES(SysmemToGmem);
+        else if (info.m_type == Dive::EventInfo::EventType::kGmemToSysmemResolve)
+            GATHER_RESOLVES(GmemToSysmem);
+        else if (info.m_type == Dive::EventInfo::EventType::kClearGmem)
+            GATHER_RESOLVES(ClearGmem);
     }
 
     GATHER_TOTAL_MIN_MAX_MEDIAN(event_num_indices, Indices);


dive_core/capture_event_info.cpp
```diff
@@ -343,4 +343,38 @@ std::string Util::GetEventString(const IMemoryManager &mem_manager,
     return string_stream.str();
 }
 
+//--------------------------------------------------------------------------------------------------
+uint32_t Util::GetVertexCount(const IMemoryManager &mem_manager,
```
wangra-google, line 347: name is a bit misleading, should be GetIndexCount?

trace_stats/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+
+if(MSVC)
+  set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} /MP")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+endif()
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+
+file(GLOB HDR_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.h")
```
wangra-google, line 32: sorry i know this is a copy version of the one in the lrz_validator, but i think file(GLOB ...) is generally discouraged in CMake. better to explicitly declare the files

set(SRC_FILES
  main.cpp
)

trace_stats/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+file(GLOB SRC_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
+
+include_directories(${THIRDPARTY_DIRECTORY}/Vulkan-Headers/include
```
wangra-google, line 35: i think it is recommended to use target_include_directories, something like:

target_include_directories(${PROJECT_NAME} PRIVATE
  ${THIRDPARTY_DIRECTORY}/Vulkan-Headers/include
  ${CMAKE_SOURCE_DIR}
  ${CMAKE_BINARY_DIR}
  ${LibArchive_INCLUDE_DIRS})

trace_stats/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+  ${CMAKE_SOURCE_DIR}
+  ${CMAKE_BINARY_DIR}
+  ${LibArchive_INCLUDE_DIRS})
+
+set(LIB_SRC_FILES ${SRC_FILES})
```
wangra-google, line 40: i think line 40 & 41 are not needed.

trace_stats/main.cpp
```diff
@@ -0,0 +1,430 @@
+/*
+ Copyright 2024 Google LLC
```
wangra-google, line 2: 2025

trace_stats/main.cpp
```diff
@@ -0,0 +1,430 @@
...
+    {
+        const Dive::EventInfo &info = meta_data.m_event_info[i];
+        if (info.m_type == Dive::EventInfo::EventType::kDraw)
+        {
+            if (info.m_render_mode == Dive::RenderModeType::kBinning)
+                stats_list[kBinningDraws]++;
+            else if (info.m_render_mode == Dive::RenderModeType::kDirect)
+                stats_list[kDirectDraws]++;
+            else if (info.m_render_mode == Dive::RenderModeType::kTiled)
+                stats_list[kTiledDraws]++;
```
wangra-google, line 252: i am wondering if it is also interesting to have a stat to show the number of drawcalls per tile in tiled mode? otherwise the user needs to do the manual division of number of tiles.

dive_core/capture_event_info.cpp
shanminchao: That's what I had it originally (along with m_num_indices). However, I realized that this could be misleading for non-indexed draws, and what we really want to know is the number of vertices being rendered. So that's why I changed it. If you think it makes more sense the other way, let me know!

trace_stats/main.cpp
shanminchao: Yeah I'm actually working on printing more info about the different passes in the next PR
I noticed the lack of clarity here too :)

trace_stats/CMakeLists.txt
shanminchao: Cleaned up the CMakeLists.txt

trace_stats/main.cpp
shanminchao: Done!

trace_stats/CMakeLists.txt
shanminchao: Done!

shanminchao:
> > This represents the initial set of stats Additional stats to be added in the future
> > image
> 
> could be interesting to output number of tiles as well.

I actually had this bit of code before:

        ostream << "\tNum Draws (TILED) / Num Draws (BINNING) = "
                << (float)stats_list[kTiledDraws] / stats_list[kBinningDraws]
                << " (Number of Tiles?)" << std::endl;

But then I realized the # returned here doesn't reflect the # of actual tiles. It was off by a factor of 2x, I think.
So I realized this needed a bit of investigation to get a more accurate #, and therefore will be part of a future PR

trace_stats/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+endif()
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+
+add_executable(${PROJECT_NAME} "main.cpp")
+target_link_libraries(${PROJECT_NAME} PRIVATE dive_core)
+target_include_directories(${PROJECT_NAME} PRIVATE
+  ${THIRDPARTY_DIRECTORY}/Vulkan-Headers/include
```
wangra-google, line 35: is vulkan header needed?

dive_core/capture_event_info.cpp
shanminchao: I see. When you think vertex count, you think of the size of the vertex buffer.
In my head, I thought of vertex count as the number of vertices processed by the vertex shader. 
Two different ways of looking at it.

I'm fine with changing it. I'm not particularly fixed on any terminology. Just have to keep in mind the "Index" in this case also includes non-index draws.

trace_stats/CMakeLists.txt
shanminchao: Yes. I get this without it:

15>------ Build started: Project: trace_stats, Configuration: Debug x64 ------
15>main.cpp
15>C:\src\dive\third_party\Vulkan-Headers\include\vulkan\vulkan_core.h(9537,10): error C1083: Cannot open include file: 'vk_video/vulkan_video_codec_h264std.h': No such file or directory

dive_core/capture_event_info.cpp
shanminchao: Changed as requested!

trace_stats/main.cpp
```diff
@@ -0,0 +1,430 @@
...
+        kSyncs,
+        kDepthTestEnabled,
+        kDepthWriteEnabled,
+        kEarlyZ,
+        kLateZ,
+        kEarlyLRZLateZ,
+        kLrzEnabled,
+        kLrzWriteEnabled,
+        kCullModeEnabled,
+        kTotalVertices,
```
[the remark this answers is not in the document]
wangra-google: those stats names should probably be updated with the name of indices as well.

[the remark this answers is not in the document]
shanminchao: Oh oops! Forgot about those. Corrected...

dive_core/capture_event_info.h
```diff
@@ -67,6 +77,9 @@ struct EventInfo
     // References of each shader used in the event.
     std::vector<ShaderReference> m_shader_references;
 
+    // Number of vertices processed, for draw calls
```
wangra-google, line 80: maybe `vertices` in this comment as well?

wangra-google (approved):

dive_core/capture_event_info.h
wangra-google: Thanks

PR #307, opened 2025-08-01
Avoid inserting timestamps to secondary command buffers

wangra-google:
We only insert timestamps for primary command buffers since it is common to reuse secondary command buffers within a frame in real apps. (we still dont support reusing primary command buffers within a frame since it is a less common usage case )

Also removed the FreeSameCommandBufferTwiceFails test since is a vk wrong behavior not gpu_time one, should be cought by the vulkan validation

Add a new test to check the case where the same cmd got submitted 2 times

### DIFF
--- a/capture_service/dive_client_cli.cc
+++ b/capture_service/dive_client_cli.cc
@@ -372,54 +372,17 @@ absl::Status is_capture_directory_busy(Dive::DeviceManager& mgr,
 bool retrieve_gfxr_capture(Dive::DeviceManager& mgr, const std::string& gfxr_capture_directory)
 {
     std::filesystem::path download_dir = absl::GetFlag(FLAGS_download_dir);
-    std::filesystem::path target_download_dir(
-    absl::StrCat(download_dir.string(), "/", gfxr_capture_directory));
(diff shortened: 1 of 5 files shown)


gpu_time/gpu_time.cpp
```diff
@@ -223,15 +225,23 @@ GPUTime::GpuTimeStatus GPUTime::OnAllocateCommandBuffers(
 const VkCommandBufferAllocateInfo* pAllocateInfo,
 VkCommandBuffer*                   pCommandBuffers)
 {
+    // The cache should not contain secondary command buffers
+    if (pAllocateInfo->level == VK_COMMAND_BUFFER_LEVEL_SECONDARY)
+    {
+        return GPUTime::GpuTimeStatus();
```
shanminchao, line 231: Would it be better to still add secondary command buffer into m_cmds, but mark it with a flag indicating it's a secondary command buffer and therefore does not have any timestamps allocated?

This way, your error checks later (i.e. "...is not in the cmd cache!") can still be invoked for any unexpected command buffers, but ignored for secondary ones.

shanminchao: Just as long as you considered it, it's fine.

PR #313, opened 2025-08-08
[mesa] Fix the dump_bos by adding recursive mutex in log_gpuaddr

wangra-google:
the bo_lock mutex cannot prevent the cases where log_gpuaddr is called in other threads. We have encountered the cases where we have 2 consecutive RD_GPUADDR blocks followed by RD_BUFFER_CONTENTS.


shanminchao (approved):
It's unclear to me whether the mutex is the right or best solution. Rob Clark needs to be the one that really reviews this. So please create a CL/PR in the freedreno repo for this fix as well.

PR #317, opened 2025-08-14
Adds Perf Counter Tab View

GrantComm:
Adds perf counter tab view that displays perf counters in a searchable table.

Screenshot 2025-08-13 at 11 25 34 PM

### DIFF
--- a/ui/command_tab_view.cpp
+++ b/ui/command_tab_view.cpp
@@ -62,7 +62,7 @@ CommandTabView::CommandTabView(const Dive::CommandHierarchy &command_hierarchy, 
     main_layout->addWidget(m_search_bar);
     main_layout->addWidget(m_command_buffer_view);
     setLayout(main_layout);
-    m_search_bar->setTreeView(m_command_buffer_view);
+    m_search_bar->setView(m_command_buffer_view);
 
     QObject::connect(m_search_trigger_button,
                      SIGNAL(clicked()),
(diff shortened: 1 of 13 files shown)


GrantComm:
> > Adds perf counter tab view that displays perf counters in a searchable table.
> > Screenshot 2025-08-13 at 11 25 34 PM
> 
> Thanks for adding this UI support. A more general question, do you think if it is feasible to use the plugin system to add this?

Follow up on the external conversation had outside of the pr. Yes, I think that it would be possible to use the plugin system to do this but would require significant changes to the existing setup of the main window. In this case, this would include being able to access the QTabWidget, modifying how the changing of tabs functions for the hiding/unhiding of search widgets, being able to access certain slots to connect the tab to the main window, and modifying how the dynamic search shortcut (Ctrl + Shift + F) functions since it also activates a search based on the current tab index

wangra-google:
> > > Adds perf counter tab view that displays perf counters in a searchable table.
> > > Screenshot 2025-08-13 at 11 25 34 PM
> > 
> > 
> > Thanks for adding this UI support. A more general question, do you think if it is feasible to use the plugin system to add this?
> 
> Follow up on the external conversation had outside of the pr. Yes, I think that it would be possible to use the plugin system to do this but would require significant changes to the existing setup of the main window. In this case, this would include being able to access the QTabWidget, modifying how the changing of tabs functions for the hiding/unhiding of search widgets, being able to access certain slots to connect the tab to the main window, and modifying how the dynamic search shortcut (Ctrl + Shift + F) functions since it also activates a search based on the current tab index

k, thanks, yeah, we can see how to do that properly in a future PR.

PR #318, opened 2025-08-15
collect performance metrics with replay

RenfengLiu:
To collect the metrics, first compile the metrics plugin and put the output folder under install `src/dive/install/dive_profiling_plugin`

and then run the following command, the output would be `com.google.bigwheels.project_sample_04_cube.debug_trim_trigger_20250807T180622.csv`

```
./install/dive_client_cli --device R32X50020WJ --command profile_replay --gfxr_replay_file_path /data/local/tmp/1/com.google.bigwheels.project_sample_04_cube.debug_trim_trigger_20250807T180622.gfxr --metrics=PERCENT_SHADERS_BUSY
```

Note, the loop replay is currently broken likely due to the previous subtree merege, to test this feature, check out an  older version of dive(such as 1b8a69dc5960a2fb55ac9afe1718038e0ff66e3c) and cherry pick this commit.

### DIFF
--- a/capture_service/dive_client_cli.cc
+++ b/capture_service/dive_client_cli.cc
@@ -42,6 +42,7 @@ enum class Command
     kNone,
     kGfxrCapture,
     kGfxrReplay,
+    kProfileReplay,
     kListDevice,
     kListPackage,
     kRunPackage,
@@ -86,6 +87,11 @@ bool AbslParseFlag(absl::string_view text, Command* command, std::string* error)
         *command = Command::kGfxrReplay;
         return true;
     }
+    if (text == "profile_replay")
+    {
+        *command = Command::kProfileReplay;
+        return true;
+    }
     if (text.empty())
     {
         *command = Command::kNone;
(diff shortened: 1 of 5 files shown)


capture_service/dive_client_cli.cc
```diff
@@ -625,12 +638,8 @@ bool process_input(Dive::DeviceManager& mgr)
     return true;
 }
 
-bool deploy_and_run_gfxr_replay(Dive::DeviceManager& mgr,
-                                const std::string&   device_serial,
-                                const std::string    gfxr_replay_capture,
-                                const std::string    gfxr_replay_flags)
+bool deploy_gfxr_replay(Dive::DeviceManager& mgr, const std::string& device_serial)
```
wangra-google, line 641: is there a reason that you prefer snake case here as the func name?

GrantComm (approved):

capture_service/dive_client_cli.cc
wangra-google: k

PR #319, opened 2025-08-15
trace_stats: Fix incorrect indices reported + cleanup

shanminchao:


dive_core/capture_event_info.cpp
```diff
@@ -361,7 +361,7 @@ uint32_t Util::GetIndexCount(const IMemoryManager &mem_manager,
     {
         PM4_CP_DRAW_INDX packet;
         DIVE_VERIFY(mem_manager.RetrieveMemoryData(&packet, submit_index, va_addr, sizeof(packet)))
-        index_count += packet.bitfields2.NUM_INDICES;
```
wangra-google, line 364: if i understand correctly, this doesn't affect the result since index_count is initialized to 0, so this is only for code cleanness?

dive_core/capture_event_info.cpp
```diff
@@ -371,8 +371,8 @@ uint32_t Util::GetIndexCount(const IMemoryManager &mem_manager,
         DIVE_VERIFY(mem_manager.RetrieveMemoryData(&packet,
                                                    submit_index,
                                                    va_addr,
-                                                   dword_count * sizeof(uint32_t)))
-        index_count += packet.bitfields2.NUM_INDICES;
+                                                   (dword_count + 1) * sizeof(uint32_t)))
```
wangra-google, line 374: why +1?

dive_core/capture_event_info.cpp
shanminchao: Yeah this part is for code cleanliness

dive_core/capture_event_info.cpp
shanminchao: Because the dword_count is for the body of the packet. There's also a dword for the header.

shanminchao: It's the count value from the header. I wonder if maybe we shouldn't rename this to header_dword_count to make it more clear?

wangra-google: yeah, would be better to either add some comments on why +1 is needed or name it slightly differently to reflect the reason.

shanminchao: I'm just gonna pass the whole header in. That should make it unambiguous what we're doing!

wangra-google: even with this change, maybe still better with comments to explain the header.count + 1. :)

GrantComm (approved):

PR #323, opened 2025-08-19
GFXR: Demote --dump-resources dynamic offset check

footballhead:
Since "entries that are not used by a pipeline can have undefined descriptors", replace the assert in vkCmdBindDescriptorSets with a warning.

If vkUpdateDescriptorSets is not called, VulkanDescriptorSetBindingInfo.buffer_info will not have contain a VulkanDescriptorTypeBufferInfo for that binding or array element (despite the descriptor set layout). This assert should likely take more context into consideration.

### DIFF
--- a/capture_service/android_application.cc
+++ b/capture_service/android_application.cc
@@ -30,7 +30,7 @@ limitations under the License.
 namespace Dive
 {
 
-int GetIndention(const std::string &line)
+int GetIndentation(const std::string &line)
 {
     int indention = 0;
     for (int i = 0; i < line.size(); i++)
@@ -72,14 +72,14 @@ std::string ParsePackageForActivity(const std::string &input, const std::string 
         if (absl::StrContains(line, "Non-Data Actions"))
         {
             non_data_action_found = true;
-            indention = GetIndention(line);
+            indention = GetIndentation(line);
             continue;
         }
 
         if (!non_data_action_found)
             continue;
 
-        int cur_indention = GetIndention(line);
+        int cur_indention = GetIndentation(line);
         if (cur_indention <= indention)
         {
             non_data_action_found = false;
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -356,7 +356,7 @@ absl::Status AndroidDevice::SetupApp(const std::string    &command,
     return m_app->Setup();
 }
 
-absl::Status AndroidDevice::CleanupAPP()
+absl::Status AndroidDevice::CleanupApp()
 {
     m_app = nullptr;
     return absl::OkStatus();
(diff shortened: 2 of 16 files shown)


third_party/gfxreconstruct/framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -2033,13 +2033,41 @@ void DrawCallsDumpingContext::BindDescriptorSets(
...
+                               "dynamicOffsetCount=%u but only found dynamic_offset_index=%u. It's likely that a "
+                               "descriptor for the set is undefined. Since \"entries that are not used by "
+                               "a pipeline can have undefined descriptors\", this might be OK.",
+                               dynamicOffsetCount,
+                               dynamic_offset_index);
+        }
+    }
+    else if (dynamic_offset_index)
+    {
+        assert(!dynamic_offset_index);
```
wangra-google, line 2069: not really a big deal, but strange that the were using assert instead of GFXRECON_ASSERT here

third_party/gfxreconstruct/framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -2033,13 +2033,41 @@ void DrawCallsDumpingContext::BindDescriptorSets(
...
+        {
+            GFXRECON_LOG_DEBUG("DrawCallsDumpingContext::BindDescriptorSets: Did not process all dynamic offsets. "
+                               "dynamicOffsetCount=%u but only found dynamic_offset_index=%u. It's likely that a "
+                               "descriptor for the set is undefined. Since \"entries that are not used by "
+                               "a pipeline can have undefined descriptors\", this might be OK.",
+                               dynamicOffsetCount,
+                               dynamic_offset_index);
+        }
+    }
+    else if (dynamic_offset_index)
```
angela28chen, line 2067: I'm not really following for this section:
```
else if (dynamic_offset_index)
{
  assert(!dynamic_offset_index);
}
```
if dynamic_offset_index == 0 --> no assert
if dynamic_offset_index >0 --> triggers assert

could it just be this then?
```
else
{
  assert(!dynamic_offset_index);
}
```

third_party/gfxreconstruct/framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -2033,13 +2033,41 @@ void DrawCallsDumpingContext::BindDescriptorSets(
...
+    {
+        assert(!dynamic_offset_index);
+    }
```
footballhead, line 2070:
```suggestion
    else
    {
        assert(!dynamic_offset_index);
    }
```

angela28chen (approved):

PR #324, opened 2025-08-19
GFXR: If queue family reused, ignore protected queue

footballhead:
When creating a device, you must specify which queues to create. Apps are allowed to provide the same a queue family index twice when it wants both a non-protected queue (flags = 0) and a protected queue (flags = 1).

For now, we suspect that the protected queue is used for niche situations. Therefore: replace the assert with a warning, and prefer the non-protected queue.

Currently, capture doesn't use the flags that are stored so this code could be removed entirely. Replay uses the flags in some places like the VulkanResourceInitializer to get a queue on which to submit work.

### DIFF
--- a/third_party/gfxreconstruct/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/third_party/gfxreconstruct/framework/decode/vulkan_replay_consumer_base.cpp
@@ -3176,8 +3176,25 @@ VkResult VulkanReplayConsumerBase::PostCreateDeviceUpdateState(VulkanPhysicalDev
     for (uint32_t q = 0; q < create_state.modified_create_info.queueCreateInfoCount; ++q)
     {
         const VkDeviceQueueCreateInfo* queue_create_info = &create_state.modified_create_info.pQueueCreateInfos[q];
-        GFXRECON_ASSERT(device_info->queue_family_creation_flags.find(queue_create_info->queueFamilyIndex) ==
-                        device_info->queue_family_creation_flags.end());
+        // GOOGLE: Given that "two members can share the same queueFamilyIndex if one describes protected-capable
+        // queues and one describes queues that are not protected-capable", an assert is too strict here. Prefer the
+        // queue that is not protected-capable for now.
+        if (const auto queue_family_creation_flags =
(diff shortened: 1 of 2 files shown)


wangra-google (approved):

third_party/gfxreconstruct/framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3160,8 +3160,25 @@ VkResult VulkanReplayConsumerBase::PostCreateDeviceUpdateState(VulkanPhysicalDev
...
+                queue_create_info->queueFamilyIndex,
+                queue_create_info->flags,
+                *queue_family_creation_flags);
+            if (queue_create_info->flags == VK_DEVICE_QUEUE_CREATE_PROTECTED_BIT)
+            {
+                GFXRECON_LOG_INFO("Ignoring create info for protected queue.");
+                continue;
+            }
+        }
         device_info->queue_family_creation_flags[queue_create_info->queueFamilyIndex] = queue_create_info->flags;
```
wangra-google, line 3182: is device_info->queue_family_creation_flags used in GetDeviceQueue? (gfxreconstruct\framework\decode\decoder_util.h) would it be problematic if we always ignore VK_DEVICE_QUEUE_CREATE_PROTECTED_BIT? let's say a protected cmd buffer is created, and sent to the queue, if the queue is not protected because we ignore this flag, could this be problematic?

footballhead: Based on my limited understanding of the code, yes this could be an issue. I don't have a reliable way of testing this though, and I don't have a full understanding of the code. I could spend some more time digging into this or we could keep our eyes peeled for potential issues in the future.

footballhead: Yes that's my feeling as well. I've definitely seen a validation error related to using vkGetDeviceQueue instead of vkGetDeviceQueue2. However, I'm not confident that I've tested protected content enough (if at all) to give you a definitive answer.

GrantComm (approved):

PR #325, opened 2025-08-19
GFXR: Replay FillProcessMemory assuming buffer is strided

footballhead:
Captures with imported AHardwareBuffers are incorrectly treating padding as part of the image. Since the capture does not include bufferRowLength, make some assumptions about how to derive it.


wangra-google:
> Captures with imported AHardwareBuffers are incorrectly treating padding as part of the image. Since the capture does not include bufferRowLength, make some assumptions about how to derive it.

the fix lgtm, should be fine on the gpus that we are working on, but might introduce problems on other gpus if the alignment is larger than 16. we can submit this on our side, but could you follow up with LunarG to see if they could make a real fix on this? Thanks

wangra-google (approved):

PR #327, opened 2025-08-21
build_android: Only build replay and layer

footballhead:
Last subtree update, we got two new Android targets:

- multi-win-replay
- quest_replay

As a result, build times doubled. In order to reduce build times, only compile parts of GFXR that we use. This also means that we don't compile the [test apps](https://github.com/google/dive/blob/main/third_party/gfxreconstruct/TESTING_test_apps.md)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ driver/xgl/builds
 third_party/freedreno/libs/
 third_party/freedreno/obj
 build_android/
+.gradle/
 # Amdvlk auto-generated files
 *.pyc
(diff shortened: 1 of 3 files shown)


footballhead:
Last main Android build took **27m**: https://github.com/google/dive/actions/runs/17959883734/job/51080514162

This Android build took **13m 12s**: https://github.com/google/dive/actions/runs/17979996130/job/51143222902?pr=327

angela28chen (approved):
Thanks for making this change!

PR #332, opened 2025-08-24
Add new arg `--enable-gpu-time` to enable gpu time measurement on Replay

wangra-google:


angela28chen (approved):

angela28chen:
Forgot to ask, has this been tested with and without the `--loop-single-frame` flag?

wangra-google:
> Forgot to ask, has this been tested with and without the `--loop-single-frame` flag?

yes, it works

PR #322, opened 2025-08-15
Dive cmd hier UI

GrantComm:
Displays the Dive Command Hierarchy, which contains pm4 and gfxr data in a single display. This PR builds off of #310, which introduced the Dive Command Hierarchy

Screenshot 2025-08-15 at 3 12 20 PM

In this new display, users can correlate a gfxr draw call with a pm4 draw call:
Screenshot 2025-08-25 at 12 22 32 AM

PR #333, opened 2025-08-25
Adds Gfxr Vulkan Command Draw Call filter

GrantComm:
Adds Gfxr vulkan command draw call filter so that only draw/dispatch and render pass calls are shown when the filter is enabled

Screenshot 2025-08-25 at 12 22 27 PM

Screenshot 2025-08-25 at 12 22 27 PM

### DIFF
--- a/ui/gfxr_vulkan_command_filter_proxy_model.cpp
+++ b/ui/gfxr_vulkan_command_filter_proxy_model.cpp
@@ -107,5 +107,11 @@ bool GfxrVulkanCommandFilterProxyModel::filterAcceptsRow(int                sour
         }
     }
 
+    // Do not include non-gfxr submits and their descendents.
+    if (m_command_hierarchy->GetNodeType(node_index) == Dive::NodeType::kSubmitNode)
+    {
+        return false;
+    }
+
     return true;
 }
--- a/ui/main_window.cpp
+++ b/ui/main_window.cpp
@@ -644,6 +644,9 @@ bool MainWindow::LoadDiveFile(const std::string &file_name)
     m_gfxr_vulkan_commands_filter_proxy_model->setSourceModel(
     m_gfxr_vulkan_command_hierarchy_model);
 
+    m_gfxr_vulkan_commands_filter_proxy_model->setSourceModel(
+    m_gfxr_vulkan_command_hierarchy_model);
+
     m_command_hierarchy_model->BeginResetModel();
 
     // Reset the tab widget.


GrantComm:
Based on #322. Please begin review at [fed0776](https://github.com/google/dive/pull/333/commits/fed07765a88b60b4a7780d78433c333e723b854a)

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -46,6 +46,24 @@ DiveAnnotationProcessor::VulkanCommandInfo vk_cmd_info)
                  m_cur_submit_node_index,
                  cmd_buffer_index);
     }
+    else if (vk_cmd_info.GetVkCmdName().find("vkCmdDraw") != std::string::npos)
```
wangra-google, line 49: do you need other draw & dispatch cmds?
vkCmdDrawIndirect
vkCmdDrawIndexed
vkCmdDrawIndexedIndirect
vkCmdDispatch
vkCmdDispatchIndirect

ui/dive_tree_view.h
```diff
@@ -53,6 +53,9 @@ class DiveFilterModel : public QSortFilterProxyModel
 
     DiveFilterModel(const Dive::CommandHierarchy &command_hierarchy, QObject *parent = nullptr);
     void SetMode(FilterMode filter_mode);
+    void CollectGfxrDrawCallIndices(const QModelIndex &parent_index = QModelIndex());
+    std::vector<uint64_t> GetPm4DrawCallIndices() { return m_pm4_draw_call_indices; }
```
wangra-google, line 57: do you need return a copy? or const std::vector<uint64_t>& could be good enough?

ui/gfxr_vulkan_command_arguments_tab_view.cpp
```diff
@@ -98,33 +98,35 @@ void GfxrVulkanCommandArgumentsTabView::OnSelectionChanged(const QModelIndex &in
...
+            source_index = proxy_model->mapToSource(index);
+        }
     }
 
-    // Map the index using the correct proxy model.
-    QModelIndex source_index = proxy_model->mapToSource(index);
+    // Always use the source_index, regardless of whether a proxy was involved.
+    m_arg_proxy_model->SetTargetParentSourceIndex(source_index);
+
+    uint32_t column_count = (uint32_t)m_command_hierarchy_model->columnCount(QModelIndex());
```
wangra-google, line 121: use static_cast<uint32_t> here

ui/gfxr_vulkan_command_filter_proxy_model.h
```diff
@@ -29,14 +29,25 @@ class GfxrVulkanCommandFilterProxyModel : public QSortFilterProxyModel
...
                                       const Dive::CommandHierarchy *command_hierarchy = nullptr);
+    void SetFilter(FilterMode filter_mode);
 
 protected:
     bool filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const override;
 
 private:
+    void ApplyNewFilterMode(FilterMode new_mode);
+
     const Dive::CommandHierarchy *m_command_hierarchy;
```
wangra-google, line 49: just to be consistent. if you prefer to initialize members in the header, better to do the init for all members.

GrantComm: Done.

ui/dive_tree_view.h
```diff
@@ -62,6 +65,8 @@ public slots:
 private:
     const Dive::CommandHierarchy &m_command_hierarchy;
     FilterMode                    m_filter_mode = kNone;
+    mutable std::vector<uint64_t> m_pm4_draw_call_indices;
```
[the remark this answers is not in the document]
wangra-google: i think we should avoid using mutable if possible since it breaks the logical const-ness of the function. I am wondering if you could populate these 2 vectors after endResetModel() in `applyNewFilterMode`?

dive_core/gfxr_vulkan_command_hierarchy.cpp
wangra-google: Thanks

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -46,7 +46,8 @@ DiveAnnotationProcessor::VulkanCommandInfo vk_cmd_info)
                  m_cur_submit_node_index,
                  cmd_buffer_index);
     }
-    else if (vk_cmd_info.GetVkCmdName().find("vkCmdDraw") != std::string::npos)
+    else if (vk_cmd_info.GetVkCmdName().find("vkCmdDraw") != std::string::npos ||
```
shanminchao, line 49: It would be better if we actually stored a command enum inside vk_cmd_info so that we can avoid string comparisons here. For captures with lots of commands, string comparisons such as this could increase load times. Furthermore, string comparisons are fundamentally not very robust. Imagine if somewhere down the line they change the capitalization for one of those command names.


However, I don't suggest changing this now. The change is additional work, and it isn't a source of load time woes yet.

wangra-google (approved):

PR #335, opened 2025-08-26
Temp Fix for  RB_BLIT_INFO

wangra-google:
We have encountered the case where rb_blit_info.bitfields.GMEM == 0 but
rb_blit_info.bitfields.UNK0 == 1 The register 0x88e3 has been updated
https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/freedreno/registers/adreno/a6xx.xml?ref_type=heads#L1712
The 1st 2 bits are actually TYPE
If rb_blit_info.bitfields.GMEM bit is set, it means either BLIT_EVENT_CLEAR
or BLIT_EVENT_LOAD In my case, the value means BLIT_EVENT_STORE_AND_CLEAR

The real fix would need integrate latest mesa code

### DIFF
--- a/dive_core/command_hierarchy.h
+++ b/dive_core/command_hierarchy.h
@@ -68,7 +68,9 @@ enum class NodeType
     kGfxrVulkanCommandBufferNode,
     kGfxrVulkanCommandNode,
     kGfxrVulkanCommandArgNode,
-    kGfxrVulkanDrawCommandNode
+    kGfxrVulkanDrawCommandNode,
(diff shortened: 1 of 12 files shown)


wangra-google:
> > curious why not pull from upstream mesa code instead
> 
> i was about to do a quick fix so that Shan can try the new capture quickly before today's meeting, but agree that better to pull the upstream code. will close this pr

actually after thinking about it, i would prefer pulling from upstream after the milestone to avoid potential extra bugs. and we need this fix so that people can start trying Dive internally within the team.

PR #336, opened 2025-08-26
Capture PM4 data at the frame boundary for gfxr replay

RenfengLiu:


gfxr_ext/decode/dive_pm4_capture.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "util/logging.h"
+
+#include <dlfcn.h>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+DivePM4Capture::DivePM4Capture()
+{
+    m_is_initialized = false;
```
wangra-google, line 28: better with initialization list

gfxr_ext/decode/dive_pm4_capture.h
```diff
@@ -0,0 +1,52 @@
...
+public:
+    using CaptureFunc = void (*)();
+
+    static DivePM4Capture& GetInstance()
+    {
+        static DivePM4Capture instance;
+        return instance;
+    }
+
+    void MaybeStartCapture();
```
wangra-google, line 36: nit: i think `Try` is a more common prefix for functions that attempt to perform an action but might fail or not proceed for some reason. 

void TryStartCapture()
void TryStopCapture()

gfxr_ext/decode/dive_file_processor.cpp
```diff
@@ -119,6 +121,7 @@ bool DiveFileProcessor::ProcessStateMarker(const format::BlockHeader& block_head
         state_end_marker_file_offset_ = TellFile(gfxr_file_name_);
         GFXRECON_LOG_INFO("Stored state end marker offset %d", state_end_marker_file_offset_);
         GFXRECON_LOG_INFO("Single frame number %d", GetFirstFrame());
+        DivePM4Capture::GetInstance().MaybeStartCapture();
```
wangra-google, line 124: should we avoid calling this (mainly for avoid calling DivePM4Capture::DivePM4Capture() ) when pm4 capture is not enabled for the Replay session?

wangra-google (dismissed):

wangra-google (approved):

PR #338, opened 2025-08-26
Add functionality for outputting GPU timing stats to a file

angela28chen:
When GPU timing data is collected with flag `--enable-gpu-time`, the stats will be output to file `gpu_time.csv` located in the same dir as the .gfxr capture.

Example usage:
`build\bin\Debug\dive_client_cli.exe --command gfxr_replay --gfxr_replay_file_path /sdcard/Download/shared_dive_internal/com.google.bigwheels.project_fishtornado_xr.debug_trim_trigger.gfxr --gfxr_replay_flags "--loop-single-frame --loop-single-frame-count 3 --enable-gpu-time"`

Example contents:
```
Type,Id,Mean [ms],Median [ms]
Frame,5,57.037,44.513
CommandBuffer,0,0.001,0.000
CommandBuffer,1,0.154,0.133
RenderPass,0,0.154,0.132
CommandBuffer,2,0.001,0.000
CommandBuffer,3,0.001,0.001
CommandBuffer,4,0.018,0.018
CommandBuffer,5,18.405,15.669
CommandBuffer,6,12.525,12.525
RenderPass,1,12.524,12.524
CommandBuffer,7,25.929,16.145
RenderPass,2,25.928,16.144
CommandBuffer,8,0.003,0.003
CommandBuffer,9,0.000,0.000
CommandBuffer,10,0.000,0.000
CommandBuffer,11,0.000,0.000
CommandBuffer,12,0.000,0.000
CommandBuffer,13,0.000,0.000
```

Other changes:
- Index fix for GFXR traces with command buffers containing multiple renderpasses

### DIFF
--- a/gfxr_ext/decode/dive_file_processor.cpp
+++ b/gfxr_ext/decode/dive_file_processor.cpp
@@ -51,7 +51,7 @@ bool DiveFileProcessor::WriteFile(const std::string& name, const std::string& co
     int   result = util::platform::FileOpen(&fd, new_file_path.c_str(), "wb");
     if (result || fd == nullptr)
     {
-        GFXRECON_LOG_ERROR("Failed to open file %s", new_file_path.c_str());
+        GFXRECON_LOG_ERROR("Failed to open file %s, exit code: %d", new_file_path.c_str(), result);
         return false;
     }
 
@@ -63,9 +63,10 @@ bool DiveFileProcessor::WriteFile(const std::string& name, const std::string& co
 
     GFXRECON_LOG_INFO("Wrote file: %s", new_file_path.c_str());
 
-    if (util::platform::FileClose(fd))
+    result = util::platform::FileClose(fd);
+    if (result)
     {
-        GFXRECON_LOG_ERROR("Failed to close file %s", new_file_path.c_str());
+        GFXRECON_LOG_ERROR("Failed to close file %s, exit code: %d", new_file_path.c_str(), result);
         return false;
     }
 
--- a/gfxr_ext/decode/dive_vulkan_replay_consumer.cpp
+++ b/gfxr_ext/decode/dive_vulkan_replay_consumer.cpp
@@ -311,7 +311,7 @@ format::HandleId                            fence)
         if (submit_status.contains_frame_boundary)
         {
             GFXRECON_LOG_INFO(gpu_time_.GetStatsString().c_str());
-            gpu_time_stats_csv_str_ += gpu_time_.GetStatsCSVString();
+            gpu_time_stats_csv_str_ = gpu_time_.GetStatsCSVString();
         }
     }
 }
--- a/gfxr_ext/decode/dive_vulkan_replay_consumer.h
+++ b/gfxr_ext/decode/dive_vulkan_replay_consumer.h
@@ -133,12 +133,16 @@ public:
 
     void SetEnableGPUTime(bool enable) { enable_gpu_time_ = enable; }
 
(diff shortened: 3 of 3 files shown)


third_party/gfxreconstruct/framework/decode/file_processor.h
```diff
@@ -288,7 +286,8 @@ class FileProcessor
     virtual bool ProcessStateMarker(const format::BlockHeader& block_header, format::MarkerType marker_type);
     virtual void StoreBlockInfo() {}
 
-    bool run_without_decoders_ = false;
+    bool        run_without_decoders_ = false;
+    std::string absolute_path_;
```
wangra-google, line 290: why do you prefer to move `absolute_path_` to under // GOOGLE? seems to be a var from LunarG?

gfxr_ext/decode/dive_file_processor.cpp
```diff
@@ -41,6 +42,35 @@ void DiveFileProcessor::SetDiveBlockData(std::shared_ptr<DiveBlockData> p_block_
...
 
+bool DiveFileProcessor::WriteFile(const std::string& name, const std::string& content)
+{
+    std::string new_file_path = absolute_path_ + "/" + name;
+
+    FILE* fd;
+    int   result = util::platform::FileOpen(&fd, new_file_path.c_str(), "wb");
+    if (result || fd == nullptr)
+    {
+        GFXRECON_LOG_ERROR("Failed to open file %s", new_file_path.c_str());
```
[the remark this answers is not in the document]
angela28chen: Could you provide an example of error code being used in `gfxr_ext` or `gfxr`? Sorry I'm not sure what that means in this context, I've just been using the macro GFXRECON_LOG_ERROR() and passing it the error message

third_party/gfxreconstruct/framework/decode/file_processor.h
wangra-google: oh, i see. thanks

gpu_time/gpu_time.cpp
```diff
@@ -303,15 +303,43 @@ std::string GPUTime::GetStatsString() const
             const Stats& renderpass_stats = GetFrameRenderPassTimeStats(renderpass_index);
             ss << "\t\tRenderPass" << renderpass_index << ": \n";
             PopulateStatsString(ss, renderpass_stats, 2);
+            renderpass_index++;
```
angela28chen, line 306: @wangra-google btw does this fix seem alright to you? I wanted to test it on a GFXR trace containing multiple renderpasses within a command buffer but I couldn't find one easily. This change is because I thought there would be a bug in the multi-renderpass case (originally, the code in this inner loop doesn't change between loops).

gfxr_ext/decode/dive_file_processor.cpp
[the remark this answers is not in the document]
angela28chen: Oh I see what you mean, thanks for clarifying, I will fix it

gpu_time/gpu_time.cpp
wangra-google: yes, i noticed this. nice catch! i guess i didn't find this bug because all of our captures do not have multi renderpasses within a command buffer.

gfxr_ext/decode/dive_file_processor.cpp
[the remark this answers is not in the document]
angela28chen: Done

wangra-google (approved):

PR #339, opened 2025-08-27
Updates ui to reflect arg support for pm4 dump

GrantComm:
Updates ui to reflect arg support for pm4 dump. When a user enables pm4 dump,  --loop-single-frame and --loop-single-frame-count arguments are no longer passed to RunReplayApk.

### DIFF
--- a/plugins/CMakeLists.txt
+++ b/plugins/CMakeLists.txt
@@ -57,3 +57,5 @@ target_link_libraries(dive_plugin_loader PRIVATE
 
 # For each new plugin, you can add an add_subdirectory call here.
 add_subdirectory(plugin_test)
+
+install(CODE "file(MAKE_DIRECTORY ${CMAKE_INSTALL_PREFIX}/plugins)")


ui/analyze_window.cpp
```diff
@@ -642,14 +650,12 @@ absl::Status AnalyzeDialog::Pm4Replay(Dive::DeviceManager &device_manager,
     else if (!m_dump_pm4_enabled && m_gpu_time_enabled)
     {
         SetReplayButton("Replaying with gpu_time enabled...", false);
-        SetReplayDownloadDir();
```
wangra-google, line 645: Is this still needed if we want to download a csv that contains the gpu time info?

https://github.com/google/dive/pull/338#discussion_r2304798460

ui/analyze_window.cpp
```diff
@@ -539,10 +539,19 @@ std::string AnalyzeDialog::GetReplayArgs()
         args += " --loop-single-frame-count " + std::to_string(frame_count);
     }
 
-    if (m_gpu_time_enabled)
+    if (m_dump_pm4_enabled && m_gpu_time_enabled)
```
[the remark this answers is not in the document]
GrantComm: Ok, modified it so that the dump pm4 and gpu time replay aren't together

ui/analyze_window.cpp
GrantComm: Moved to AnalyzeDialog::OnReplay() so that the the download directory is set when performing gpu time, perf counter, or dump pm4 replay

wangra-google (approved):

PR #343, opened 2025-08-28
Update the clang version to 19 to match with gLinux

RenfengLiu:
This also fixed the compilation issue with it.

With clang-19, there's compilation error:

```
[build] /usr/local/google/home/renfeng/src/dive/trace_stats/main.cpp:255:73: error: variable 'num_draws_in_pass' set but not used [-Werror,-Wunused-but-set-variable]
[build]   255 |     uint32_t             num_binning_passes = 0, num_tiling_passes = 0, num_draws_in_pass = 0;
[build]       |                                                                         ^
[build] 1 error generated.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,9 +128,6 @@ else()
     endif()
 endif()
 
-option(DIVE_UNIT_TESTS, "Build unit tests for dive")
-
-
 option(DIVE_PYTHON_BINDINGS "Build Python bindings for dive")
 
 if(DIVE_PYTHON_BINDINGS)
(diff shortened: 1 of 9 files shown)


trace_stats/main.cpp
```diff
@@ -252,14 +252,17 @@ void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &o
     std::set<WindowScissor>         window_scissors;
 
     Dive::RenderModeType cur_type = Dive::RenderModeType::kUnknown;
-    uint32_t             num_binning_passes = 0, num_tiling_passes = 0, num_draws_in_pass = 0;
+    uint32_t             num_binning_passes = 0, num_tiling_passes = 0;
+#ifndef NDEBUG
+    uint32_t num_draws_in_pass = 0;
```
footballhead, line 257: FYI: Throwing this out there as an option, but you could also consider `[[maybe_unused]]`. This would preserve the original behavior (small waste in stack space, unnecessary computation) which may be preferrable in some situations if you don't want to use #ifdefs (the additional conditions could hamper readability by making code harder to follow).

footballhead (approved):
LGTM. From what I can tell, `_DEBUG` is MSVC only and tied to MSVC details whereas `NDEBUG` is a C standard (more cross platform). https://stackoverflow.com/questions/2290509/debug-vs-ndebug

wangra-google (approved):

footballhead (approved):

PR #344, opened 2025-08-28
 Do not show cmds from the loading phase in vk tree view

GrantComm:
Modifies file_processor to not show cmds from the loading phase in dive vk tree view.

- Needs testing to verify correct commands are shown/hidden


wangra-google (commented):
This is not going to work since some command buffers can be recorded before the kEndMarker but submitted after this marker

PR #345, opened 2025-08-29
Fixes for GPU time <-> vk correlation

wangra-google:
Fix on Replay apk:
- it seems that the OS is inserting the same empty cmd for left and right eye without VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT. This in theory should trigger a validation error, but not in this case
For now we keep both in the frame cache m_frame_cmds, since it is easier for vk correlation (but the gpu time will be equal to the last cmd that is being submitted). This does not affect timing since it is empty, but needs some further investigation.

Fix on vk tree view:
We should not create the tree view by purely following the block order since
- Some command buffers could be pre-recorded only once, but submitted 2 times in one frame for left and right eye (seens to be allowed in OpenXR Vulkan apps), in this case, we want to show the vk events in both submissions
- The app can call vkBeginCommandBuffer on cmdA, then call vkBeginCommandBuffer on cmdB, then submitt cmdC, then call vkEndCommandBuffer on cmdB and cmdA. In this case, we should keep the vk events in submission order (instead of recording order) so that it is easier to correlate with PM4 and gpu time

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,14 +85,12 @@ jobs:
           sudo apt-get install --yes \
             cmake \
             gcc-14 \
-            clang-14 \
             libsystemd-dev \
             libbsd-dev \
             ninja-build \
             python3-mako \
             libxcb-glx0-dev
           which gcc-14
-          which clang-14
       - name: Generate build files with cmake
         run: |
           cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
@@ -123,14 +121,12 @@ jobs:
           sudo apt-get update --yes
           sudo apt-get install --yes \
             cmake gcc-14 \
-            clang-14 \
             libsystemd-dev \
             libbsd-dev \
             ninja-build \
             python3-mako \
             libxcb-glx0-dev
           which gcc-14
-          which clang-14
       - name: Generate build files with cmake
         run: |
           cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
@@ -160,22 +156,22 @@ jobs:
         run: |
           sudo apt-get update --yes
           sudo apt-get install --yes \
-          cmake clang-14 \
+          cmake clang-19 \
           libsystemd-dev \
           libbsd-dev \
           ninja-build \
           python3-mako \
           libxcb-glx0-dev
-          which clang-14
+          which clang-19
       - name: Generate build files with cmake
         run: |
           cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
             -DCMAKE_BUILD_TYPE=Debug \
-            -DCMAKE_C_COMPILER:FILEPATH=/usr/bin/clang-14 \
-            -DCMAKE_CXX_COMPILER:FILEPATH=/usr/bin/clang++-14 \
+            -DCMAKE_C_COMPILER:FILEPATH=/usr/bin/clang-19 \
+            -DCMAKE_CXX_COMPILER:FILEPATH=/usr/bin/clang++-19 \
             -GNinja \
             -Bbuild .
-      - name: Build Dive host tools with ninja with clang-14
+      - name: Build Dive host tools with ninja with clang-19
         run: |
           ninja -C build
       - name: Run tests
@@ -196,22 +192,22 @@ jobs:
         run: |
           sudo apt-get update --yes
           sudo apt-get install --yes \
-          cmake clang-14 \
+          cmake clang-19 \
           libsystemd-dev \
           libbsd-dev \
           ninja-build \
           python3-mako \
           libxcb-glx0-dev
-          which clang-14
+          which clang-19
       - name: Generate build files with cmake
         run: |
           cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
             -DCMAKE_BUILD_TYPE=Release \
-            -DCMAKE_C_COMPILER:FILEPATH=/usr/bin/clang-14 \
-            -DCMAKE_CXX_COMPILER:FILEPATH=/usr/bin/clang++-14 \
+            -DCMAKE_C_COMPILER:FILEPATH=/usr/bin/clang-19 \
+            -DCMAKE_CXX_COMPILER:FILEPATH=/usr/bin/clang++-19 \
             -GNinja \
             -Bbuild .
-      - name: Build Dive host tools with ninja with clang-14
+      - name: Build Dive host tools with ninja with clang-19
         run: |
           ninja -C build
       - name: Run tests
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,9 +128,6 @@ else()
     endif()
 endif()
 
-option(DIVE_UNIT_TESTS, "Build unit tests for dive")
-
-
 option(DIVE_PYTHON_BINDINGS "Build Python bindings for dive")
 
 if(DIVE_PYTHON_BINDINGS)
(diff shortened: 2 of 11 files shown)


gfxr_ext/decode/dive_annotation_processor.h
```diff
@@ -119,13 +121,14 @@ class DiveAnnotationProcessor : public gfxrecon::decode::AnnotationHandler
     {
     }
 
-    bool WriteBinaryFile(const std::string& filename, uint64_t data_size, const uint8_t* data);
-
-    std::vector<std::unique_ptr<SubmitInfo>> GetSubmits() { return std::move(m_submits); }
+    std::vector<std::unique_ptr<SubmitInfo>> TakeSubmits() { return std::move(m_submits); }
```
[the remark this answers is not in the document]
wangra-google: I rename them to use Take on purpose: it is not a getter here. the getter won't modify members whereas here the function actually use std::move which takes the ownership of the member.

gfxr_ext/decode/dive_annotation_processor.cpp
```diff
@@ -25,26 +25,43 @@
...
+                    const auto& command_buffers = submit["pCommandBuffers"];
+                    for (const auto& cmd_buffer : command_buffers)
+                    {
+                        vk_command_buffer_handles.push_back(cmd_buffer);
+                    }
+                }
+            }
+        }
+        submit_ptr->TakeVkCommandBufferHandles(vk_command_buffer_handles);
+        submit_ptr->TakeNoneCmdVkCommands(m_none_cmd_vk_commands_per_submit_cache);
```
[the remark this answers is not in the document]
wangra-google: because those are the vk events that doesn't belong to any command buffer, but before current submit which need to be shown in the tree view before this submission.

gfxr_ext/decode/dive_annotation_processor.h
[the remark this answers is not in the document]
wangra-google: it is not a good case to use unique_ptr here, it will cause unnecessary heap allocation and indirection. But your comment about ` just the vehicle to pass it over to the command hierarchy code` is valid. I think in this case, using POD-like struct for SubmitInfo (as well as VulkanCommandInfo) is much cleaner.

wangra-google:
> In my opinion it's confusing to have DiveAnnotationProcessor::Take*() mean "the arguments are assigned to the members" but SubmitInfo::Take*() mean "function returns std::move() of the members". My point was the inconsistency of the meaning of "Take" across these two classes. I don't think we must name them as "Get". Though I will point out that currently we have DiveAnnotationProcessor::GetSubmits() and I personally think it's fine for a function prefixed Get to modify the members because it's indicated that this isn't a const function like the getters that do not modify the members.

Good point of the inconsistency, using POD like struct to avoid this. 

You are technically right that without `const` signals that a function can modify an object, but relying on that alone for a function named `Get` creates a confusing and potentially risky API. A developer calling it for a second time would be surprised to get an empty or invalid result, which could lead to subtle bugs. 

The universal convention for `Get` prefix is a query. It implies a read-only operation to retrieve data. Programmers would normally assume that getters are "free" to call multiple times without side effects. `Take` explicitly signal a destructive read. It communicates that you are not just getting the data, but you are also taking ownership of it, leaving the source object modified (often empty).

gfxr_ext/decode/dive_annotation_processor.h
```diff
@@ -32,85 +32,39 @@ class DiveAnnotationProcessor : public gfxrecon::decode::AnnotationHandler
...
-        void                          SetCmdCount(uint32_t cmd_count) { m_cmd_count = cmd_count; }
-        uint32_t                      GetCmdCount() const { return m_cmd_count; }
-        const nlohmann::ordered_json& GetArgs() const { return m_args; }
-
-    private:
-        nlohmann::ordered_json m_args;
-        std::string            m_name;
-        uint32_t               m_index;
-        uint32_t               m_cmd_count;  // Only used by vkBeginCommandBuffers
+        nlohmann::ordered_json args{};
```
angela28chen, line 42: nit: Suggested to use assignment-like syntax for initialization with intended literal value
```
nlohmann::ordered_json args = {};
std::string name = "";
```

gfxr_ext/decode/dive_annotation_processor.h
```diff
@@ -32,85 +32,39 @@ class DiveAnnotationProcessor : public gfxrecon::decode::AnnotationHandler
...
-            m_vulkan_commands = std::move(vulkan_commands);
-        }
-
-    private:
-        std::vector<VulkanCommandInfo> m_vulkan_commands{};
-        std::string                    m_name{ "" };
-        uint32_t                       m_command_buffer_count{ 0 };
+        // Keeps all the vk commands that come before this submit and that are not associated with
+        // any command buffer
+        std::vector<VulkanCommandInfo> none_cmd_vk_commands{};
```
angela28chen, line 56: nit: Suggested to use assignment-like syntax for initialization with intended literal value
```
std::vector<VulkanCommandInfo> none_cmd_vk_commands = {};
std::vector<uint64_t> vk_command_buffer_handles = {};
std::string           name = "";
```

gfxr_ext/decode/dive_annotation_processor.h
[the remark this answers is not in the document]
angela28chen: OK, makes sense, thank you!

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -48,11 +48,10 @@ void GfxrVulkanCommandHierarchyCreator::ConditionallyAddChild(uint64_t node_inde
 
 //--------------------------------------------------------------------------------------------------
 void GfxrVulkanCommandHierarchyCreator::OnCommand(
-uint32_t                                   parent_index,
 DiveAnnotationProcessor::VulkanCommandInfo vk_cmd_info)
```
angela28chen, line 51: Should the argument `vk_cmd_info` be a const reference struct instead of passing by value?

dive_core/gfxr_capture_data.cpp
```diff
@@ -67,7 +67,8 @@ CaptureData::LoadResult GfxrCaptureData::LoadCaptureFile(const std::string& file
         return LoadResult::kFileIoError;
     }
 
-    m_gfxr_submits = dive_annotation_processor.GetSubmits();
+    m_gfxr_submits = dive_annotation_processor.TakeSubmits();
+    m_gfxr_command_buffers = dive_annotation_processor.TakeVkCommandsCache();
```
angela28chen, line 71: Might be a good idea to validate input and assert that neither of these are empty, if that is always expected to be the case? I'm thinking about if the original struct was formed wrong or if Take is called multiple times by mistake.

angela28chen (approved):
Thanks for the additional context on Get/Take, good to know.

I think the POD struct is a good idea here and definitely simplifies the code, thanks for making this change.

gfxr_ext/decode/dive_annotation_processor.h
wangra-google: done

dive_core/gfxr_vulkan_command_hierarchy.cpp
wangra-google: done

dive_core/gfxr_capture_data.cpp
wangra-google: added assert for m_gfxr_submits but not for m_gfxr_command_buffers  since command buffers could be empty.

GrantComm (approved):

dive_core/gfxr_capture_data.cpp
```diff
@@ -105,4 +107,12 @@ GfxrCaptureData::GetGfxrSubmits() const
     return m_gfxr_submits;
 }
 
+//--------------------------------------------------------------------------------------------------
+const std::vector<DiveAnnotationProcessor::VulkanCommandInfo>&
+GfxrCaptureData::GetGfxrCommandBuffers(uint64_t cmd_handle) const
+{
+    DIVE_ASSERT(m_gfxr_command_buffers.count(cmd_handle) != 0);
```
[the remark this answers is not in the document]
wangra-google: will keep it as assert for now (would crash at `m_gfxr_command_buffers.at` if this is happening.) will see if this needs to be changed to error later.

PR #346, opened 2025-09-02
Add code to parse perf metrics results

RenfengLiu:


dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,236 @@
...
+namespace Dive
+{
+namespace
+{
+constexpr int     kPerfMetricsFixFieldCount = 9;
+const char* const kFixedHeaders[] = { "ContextID",   "ProcessID", "FrameID",
+                                      "CmdBufferID", "DrawID",    "DrawType",
+                                      "DrawLabel",   "ProgramID", "LRZState" };
+
+void trim(std::string& s)
```
footballhead, line 38: nit: style-wise, do we prefer `Trim`?
```suggestion
void Trim(std::string& s)
```

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include "dive_core/available_metrics.h"
+
+namespace Dive
+{
+namespace
+{
+constexpr int     kPerfMetricsFixFieldCount = 9;
+const char* const kFixedHeaders[] = { "ContextID",   "ProcessID", "FrameID",
+                                      "CmdBufferID", "DrawID",    "DrawType",
+                                      "DrawLabel",   "ProgramID", "LRZState" };
```
footballhead, line 36: Consider using a `std::array` with CTAD. That way, the size of the array always reflects the number of elements in the array. If it is in a separate variable, one might forget to modify `kFixedHeaders` but not `kPerfMetricsFixFieldCount` (or vice-versa).

Also, consider `constexpr`  as a stronger constness guarantee:
```suggestion
#include <array>
constexpr std::array kFixedHeaders = { "ContextID",   "ProcessID", "FrameID",
                                      "CmdBufferID", "DrawID",    "DrawType",
                                      "DrawLabel",   "ProgramID", "LRZState" };
```
Then, all usages of `kPerfMetricsFixFieldCount` could be replaced with `kFixedHeaders.size()`

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,236 @@
...
+{
+    s.erase(std::find_if(s.rbegin(), s.rend(), [](unsigned char ch) { return !std::isspace(ch); })
+            .base(),
+            s.end());
+    s.erase(s.begin(),
+            std::find_if(s.begin(), s.end(), [](unsigned char ch) { return !std::isspace(ch); }));
+}
+
+// Helper functions for safe string to number conversion
+bool safe_stoull(const std::string& s, uint64_t& out)
```
footballhead, line 48: nit: style-wise, do we prefer something like `SafeStoull`?
```suggestion
bool SafeStoull(const std::string& s, uint64_t& out)
```
(This is similar to how Abseil names their own versions of `atoi` as `absl::SimpleAtoi`)

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,236 @@
...
+{
+    char*       end = nullptr;
+    const char* start = s.c_str();
+    errno = 0;
+    unsigned long val = std::strtoul(start, &end, 10);
+    if (errno == ERANGE || *end != '\0' || start == end)
+    {
+        return false;
+    }
+    out = val;
```
footballhead, line 72: This works but is nuanced so I thought that I'd document it for posterity. Since `sizeof(unsigned long) != sizeof(uint32_t)` on GCC/clang, this assignment will work because the spec guarantees that implicit conversion of unsigned types works via modulo arithmetic.

That said, will we--practically speaking--encounter a string representing an 8-byte integer that MSVC cannot convert (ERANGE) but GCC/clang can?

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+    std::string line;
+    // Read header line
+    if (!std::getline(file, line) || line.empty())
+    {
+        return nullptr;
+    }
+    trim(line);
+
+    std::stringstream header_ss(line);
```
footballhead, line 129: Consider in the future splitting this function up into several functions. Some ideas that come to mind:

- one that parses the header for the metrics
- one that converts a string based on metric type
- one that converts a row into a record

Each of these subfunctions can be tested more thoroughly independently for edge cases. Additionally, this might be more readable due to decreased cognitive complexity.

Google C++ style recommends (if posslbe) splitting up a function if it's more than 40 lines: https://google.github.io/styleguide/cppguide.html

Another way of looking at the problem: clang-tidy has a `readability-function-cognitive-complexity` that penalizes conditionals and loops, among other things: https://clang.llvm.org/extra/clang-tidy/checks/readability/function-cognitive-complexity.html

footballhead (approved):
Would love some unit tests ;)

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,236 @@
...
+    std::ifstream file(file_path);
+    if (!file.is_open())
+    {
+        std::cerr << "Failed to open file: " << file_path << std::endl;
+        return nullptr;
+    }
+
+    std::string line;
+    // Read header line
+    if (!std::getline(file, line) || line.empty())
```
footballhead, line 123: Optional: since `getline` is typically followed by a `Trim`, consider whether it makes sense to have a helper function that does both, something like `bool GetTrimmedLine(std::string&)`

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,236 @@
...
+    unsigned long val = std::strtoul(start, &end, 10);
+    if (errno == ERANGE || *end != '\0' || start == end)
+    {
+        return false;
+    }
+    out = val;
+    return true;
+}
+
+bool SafeStoll(const std::string& s, int64_t& out)
```
footballhead, line 76: Optional: consider whether it makes sense to return instead of using an output parameter:
```
std::optional<int64_t> SafeStol(const std::string&)
```
This is recommended by Abseil tips https://abseil.io/tips/176

I realize that this doesn't necessarily jive with how it's currently being used but thought that I'd point it out as something to consider in the future

footballhead (approved):

footballhead (dismissed):

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,248 @@
...
+{
+    s.erase(std::find_if(s.rbegin(), s.rend(), [](unsigned char ch) { return !std::isspace(ch); })
+            .base(),
+            s.end());
+    s.erase(s.begin(),
+            std::find_if(s.begin(), s.end(), [](unsigned char ch) { return !std::isspace(ch); }));
+}
+
+// Helper functions for safe string to number conversion
+bool SafeStoull(const std::string& s, uint64_t& out)
```
wangra-google, line 48: i feel that the SafeSto... functions are quite similar, maybe better to use template to avoid duplicating the code? something like:

```
template<typename T> bool SafeConvertFromString(const std::string& s, T& out)
{
    if constexpr (std::is_floating_point_v<T>)
    {
        char*       end = nullptr;
        const char* start = s.c_str();
        errno = 0;
        T val = std::strtof(start, &end);
        if (errno == ERANGE || *end != '\0' || start == end)
        {
            return false;
        }
        out = val;
        return true;
    }
    else if constexpr (std::is_integral_v<T>)
    {
        char*       end = nullptr;
        const char* start = s.c_str();
        errno = 0;

        long long val = 0;
        if constexpr (std::is_signed_v<T>)
            val = std::strtoll(start, &end, 10);
        else
            val = std::strtoull(start, &end, 10);

        if (errno == ERANGE || *end != '\0' || start == end)
        {
            return false;
        }

        if (val < std::numeric_limits<T>::min() || val > std::numeric_limits<T>::max())
            return false;

        out = static_cast<T>(val);
        return true;
    }
    return false;
}
```

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,248 @@
...
+            !SafeStoull(fields[1], record.m_process_id) ||
+            !SafeStoull(fields[2], record.m_frame_id) ||
+            !SafeStoull(fields[3], record.m_cmd_buffer_id) ||
+            !SafeStoul(fields[4], record.m_draw_id) || !SafeStoul(fields[5], draw_type) ||
+            !SafeStoul(fields[6], record.m_draw_label) ||
+            !SafeStoull(fields[7], record.m_program_id) || !SafeStoul(fields[8], lrz_state))
+        {
+            continue;  // Skip lines with parsing errors
+        }
+        record.m_draw_type = static_cast<uint8_t>(draw_type);
```
wangra-google, line 191: potential overflow by converting from uint32_t to uint8_t. (which could be caught by SafeConvertFromString() with the std::numeric_limits<T> check)

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,248 @@
...
+        }
+
+        if (fields.size() != kFixedHeaders.size() + metric_names.size())
+        {
+            continue;  // Skip malformed lines
+        }
+
+        PerfMetricsRecord record{};
+        uint32_t          draw_type = 0, lrz_state = 0;
+        if (!SafeStoull(fields[0], record.m_context_id) ||
```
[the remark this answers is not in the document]
wangra-google: And you can later do something like:

```
std::optional<PerfMetricsRecord> record_opt = ParseRecordHeader(fields);
if (!record_opt.has_value())
{
        continue; 
}
PerfMetricsRecord record = std::move(record_opt.value());
```

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,235 @@
...
+        bool all_metrics_parsed = true;
+        for (size_t i = 0; i < metric_infos.size(); ++i)
+        {
+            const std::string& value_str = fields[kFixedHeaders.size() + i];
+            const MetricInfo*  info = metric_infos[i];
+            if (info)
+            {
+                switch (info->m_metric_type)
+                {
+                case MetricType::kCount:
```
wangra-google, line 190: This part can probably be simplified with template as well, something like:

```
template<typename T>
bool ParseAndEmplaceMetric(const std::string& s, std::vector<std::variant<int64_t, float>>& values)
{
    T val{};
    if (SafeConvertFromString(s, val))
    {
        values.emplace_back(val);
        return true;
    }
    return false;
}
```

and this block can be something like:

```
switch (info->m_metric_type)
{
        case MetricType::kCount:
            metric_parsed = ParseAndEmplaceMetric<int64_t>(value_str, record.m_metric_values);
            break;
        case MetricType::kPercent:
            metric_parsed = ParseAndEmplaceMetric<float>(value_str, record.m_metric_values);
            break;
        default:
            metric_parsed = false;
            break;
}
```

wangra-google (approved):

wangra-google (approved):

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,192 @@
...
+    {
+        values.emplace_back(val);
+        return true;
+    }
+    return false;
+}
+
+bool ParseMetrics(const std::vector<std::string>&       fields,
+                  const std::vector<const MetricInfo*>& metric_infos,
+                  std::optional<PerfMetricsRecord>&     record_opt)
```
footballhead, line 99: Since the record is required, consider taking it by reference. This better communicates the intent to the caller and means that you can't accidentally derefence an empty optional:
```suggestion
                  PerfMetricsRecord&     record)
```
It's easy to pass a reference by dereferencing the optional at the callsite (after has_value() has been checked)
```c++
if (ParseMetrics(fields, metric_infos, *record_opt))
```

dive_core/perf_metrics_data.cpp
```diff
@@ -0,0 +1,192 @@
...
+namespace
+{
+constexpr std::array kFixedHeaders = { "ContextID",   "ProcessID", "FrameID",
+                                       "CmdBufferID", "DrawID",    "DrawType",
+                                       "DrawLabel",   "ProgramID", "LRZState" };
+
+bool ParseHeaders(const std::string&              line,
+                  const AvailableMetrics&         available_metrics,
+                  std::vector<std::string>&       metric_names,
+                  std::vector<const MetricInfo*>& metric_infos)
```
footballhead, line 39: Optional: In the future, consider whether you can return the names and infos instead of using output parameters. This Abseil tip goes into more details: https://abseil.io/tips/176

 This can be accomplished using a struct wrapped in an optional:
```suggestion
struct ParseHeadersResult {
    std::vector<std::string> metric_names;
    std::vector<const MetricInfo*> metric_infos;
};
std::optional<ParseHeadersResult> ParseHeaders(const std::string& line, const AvailableMetrics& available_metrics)
```

footballhead (approved):
Thank you for splitting the one function into several helpers. I think it really aids readability

PR #349, opened 2025-09-03
Add string utils

RenfengLiu:


wangra-google (dismissed):

dive_core/tests/string_utils_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+    EXPECT_FALSE(SafeConvertFromString("123.45", val));
+
+    // Test limits
+    EXPECT_TRUE(SafeConvertFromString(std::to_string(std::numeric_limits<int>::max()), val));
+    EXPECT_EQ(val, std::numeric_limits<int>::max());
+    EXPECT_TRUE(SafeConvertFromString(std::to_string(std::numeric_limits<int>::min()), val));
+    EXPECT_EQ(val, std::numeric_limits<int>::min());
+
+    // Out of range
+    long long too_big = (long long)std::numeric_limits<int>::max() + 1;
```
footballhead, line 50: Prefer named casts over C casts since the intent is more explicit and the compiler can catch errors. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast 
```suggestion
    long long too_big = static_cast<long long>(std::numeric_limits<int>::max()) + 1;
```

dive_core/tests/string_utils_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+    EXPECT_FALSE(SafeConvertFromString(std::to_string(too_big), val));
+    long long too_small = (long long)std::numeric_limits<int>::min() - 1;
```
footballhead, line 52: Prefer named casts over C casts since the intent is more explicit and the compiler can catch errors. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast 
```suggestion
    long long too_small = static_cast<long long>(std::numeric_limits<int>::min()) - 1;
```

footballhead (approved):

PR #352, opened 2025-09-04
Fix the vk event tree view

wangra-google:


wangra-google:
> LGTM, is any of the WriteBlockEnd() code covered by unit testing? If so I think you should add a test case to cover this.

sure, i can add the unit test in a different PR

GrantComm (approved):

PR #353, opened 2025-09-04
Add AvailableGpuTiming class

angela28chen:
AvailableGpuTiming class reads gpu timing csv files into memory and validates their contents.

### DIFF
--- a/dive_core/available_gpu_time.cpp
+++ b/dive_core/available_gpu_time.cpp
@@ -105,7 +105,7 @@ bool AvailableGpuTiming::LoadFromStream(std::istream& stream)
     return true;
 }
 
-bool AvailableGpuTiming::LoadLine(uint32_t row, const std::string& line, uint32_t expected_columns)
+bool AvailableGpuTiming::LoadLine(uint32_t row, const std::string& line)
 {
     std::stringstream        ss(line);
     std::string              field;
@@ -120,12 +120,14 @@ bool AvailableGpuTiming::LoadLine(uint32_t row, const std::string& line, uint32_
         return true;
     }
 
-    if (fields.size() != expected_columns)
+    if (fields.size() != kExpectedColumns)
     {
         std::cerr << "Unexpected number of columns: " << fields.size() << std::endl;
         return false;
     }
 
+    // TODO(b/443122531): Improve integer and float parsing here, this has edge cases that aren't
+    // covered
     uint32_t id;
     Stats    stats;
     try
--- a/dive_core/available_gpu_time.h
+++ b/dive_core/available_gpu_time.h
@@ -85,9 +85,7 @@ private:
     bool LoadFromStream(std::istream& stream);
 
     // Load statistics from non-header CSV row
-    bool LoadLine(uint32_t           row,
-                  const std::string& line,
-                  uint32_t           expected_columns = kExpectedColumns);
+    bool LoadLine(uint32_t row, const std::string& line);
 
     // Check m_ordered_entries against info stored in *_stats members
     void Validate();


dive_core/available_gpu_time.h
```diff
@@ -0,0 +1,101 @@
...
+TODO: Retrieve GPU timing info from device to host through network/ rather than using a CSV file.
+*/
+
+namespace
+{
+constexpr char kExpectedHeader[] = "Type,Id,Mean [ms],Median [ms]";
+constexpr int  kExpectedColumns = 4;
+}  // namespace
+
+class AvailableGpuTiming
```
wangra-google, line 40: i am wondering if it is better to add a Reset() func so that this class to be able to use the same object to load different files

dive_core/available_gpu_time.h
```diff
@@ -0,0 +1,101 @@
...
+    // Load statistics from a CSV file and flag as loaded afterwards
+    bool LoadFromCsv(const std::filesystem::path& file_path);
+
+    // Load statistics from a string and flag as loaded afterwards
+    // For unit testing
+    bool LoadFromString(const std::string& full_text);
+
+    // Get the statistic info with the ObjectType and the id (nth object of type ObjectType)
+    // If object_type is kFrame, the object_id value is disregarded
+    std::optional<Stats> GetStats(ObjectType object_type, uint32_t object_id) const;
```
wangra-google, line 75: i find overloading GetStats is a bit confusing here: 
In GetStats(ObjectType object_type, uint32_t object_id), it means the 0-based
In GetStats(uint32_t object_id), it means the 1-based row index from the original file

Comment helps here, but might be better to have different names, something like
```
std::optional<Stats> GetStatsByRow(uint32_t one_based_row_id) const;
std::optional<Stats> GetStatsByType(ObjectType object_type, uint32_t per_type_id) const;
```

dive_core/tests/available_gpu_time_test.cpp
```diff
@@ -0,0 +1,215 @@
...
+{
+    AvailableGpuTiming g;
+    EXPECT_FALSE(g.IsValid());
+    EXPECT_TRUE(g.LoadFromCsv(fp / "mock_gpu_time.csv"));
+    EXPECT_TRUE(g.IsValid());
+
+    std::optional<AvailableGpuTiming::Stats>
+    ret = g.GetStats(AvailableGpuTiming::ObjectType::kFrame, 0);
+    ASSERT_NE(ret, std::nullopt);
+    EXPECT_EQ(ret->mean_ms, 0.345f);
```
wangra-google, line 81: EXPECT_FLOAT_EQ

wangra-google: same for other floating point comparisons

dive_core/available_gpu_time.cpp
```diff
@@ -0,0 +1,265 @@
...
+    if (fields.size() != expected_columns)
+    {
+        std::cerr << "Unexpected number of columns: " << fields.size() << std::endl;
+        return false;
+    }
+
+    uint32_t id = static_cast<uint32_t>(std::stoi(fields[1]));
+
+    Stats stats;
+    stats.mean_ms = std::stof(fields[2]);
```
wangra-google, line 136: what if the data is not a float?

dive_core/tests/available_gpu_time_test.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+TEST(AvailableGpuTiming, LoadFromCsv_Successful)
+{
+    AvailableGpuTiming g;
+    EXPECT_FALSE(g.IsValid());
+    EXPECT_TRUE(g.LoadFromCsv(fp / "mock_gpu_time.csv"));
+    EXPECT_TRUE(g.IsValid());
+}
+
+TEST(AvailableMetrics, LoadFromCsv_MalformedFrameFail)
```
wangra-google, line 37: is there a specific reason that you prefer to use two different names for the test suite: AvailableGpuTiming and AvailableMetrics?

dive_core/available_gpu_time.h
angela28chen: Rather than reset, my intent for loading a separate file is just to discard the object and construct a new one. I don't think that adding `Reset()` would offer a significant performance boost and I think it adds unnecessary complexity?

wangra-google: oh, Reset is not for performance, it is more for an option to reuse the object of this class, otherwise the user need to create a new AvailableGpuTiming if they want to load another one.

dive_core/tests/available_gpu_time_test.cpp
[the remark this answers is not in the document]
angela28chen: Honestly I feel like the current way is not bad. The test case itself is longer but it's more straightforward and you don't have to scroll as much to find out what the testcases / checker function is. Maybe I can change it to this if we add more unit tests and it becomes too long?

dive_core/available_gpu_time.cpp
```diff
@@ -0,0 +1,265 @@
...
+            return false;
+        }
+        row++;
+    }
+
+    Validate();
+    return IsValid();
+}
+
+bool AvailableGpuTiming::LoadFromString(const std::string& full_text)
```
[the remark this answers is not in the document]
angela28chen: Thanks for the suggestion, will do!

dive_core/tests/available_gpu_time_test.cpp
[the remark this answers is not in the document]
wangra-google: sure, i dont really have strong opinion on this. The thing i would avoid in general is to avoid duplicating code if possible. it is very easy to miss some of the copies if any modification is needed in the future.

dive_core/available_gpu_time.h
angela28chen: I think it's probably OK to rely on the destructor/constructor here rather than have a reset func, so I'll just leave it as-is

dive_core/available_gpu_time.h
```diff
@@ -0,0 +1,104 @@
...
+    bool IsValid() const { return m_valid; };
+
+private:
+    // Load statistics from stream
+    bool LoadFromStream(std::istream& stream);
+
+    // Load statistics from non-header CSV row
+    bool LoadLine(uint32_t           row,
+                  const std::string& line,
+                  uint32_t           expected_columns = kExpectedColumns);
```
wangra-google, line 90: nit: you probably don't need the default arg since LoadLine is only called once.

dive_core/available_gpu_time.cpp
```diff
@@ -0,0 +1,291 @@
...
+        std::cerr << "Caught invalid_argument exception: " << e.what() << std::endl;
+        return false;
+    }
+    catch (const std::out_of_range& e)
+    {
+        std::cerr << "Caught out_of_range exception: " << e.what() << std::endl;
+        return false;
+    }
+
+    if (fields[1].find('.') != std::string::npos)
```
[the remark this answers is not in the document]
angela28chen: Do you think it makes sense to add these functions into `Dive::StringUtils`, or another common location?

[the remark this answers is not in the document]
wangra-google: sure, we can add it to Dive::StringUtils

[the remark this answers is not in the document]
angela28chen: It seems like using the <charconv> library is producing a lot of build issues with dive core, I haven't been able to get it to build. I think that I want to go ahead with what I have now and then log a bug for making the parsing more robust later on. I think that since the .csv file is being produced by us, the parsing having some not-covered edge cases is alright for now

dive_core/available_gpu_time.h
angela28chen: Done

wangra-google (approved):

dive_core/available_gpu_time.cpp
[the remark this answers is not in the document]
angela28chen: I linked this thread in the internal bug and I'll revisit when I address that. I think I tried including `<charconv>` into this file directly without doing it through StringUtils and still had issues, so I'm just not going to attempt that right now because we are short on time.

wangra-google (approved):

PR #354, opened 2025-09-04
Adds draw call count to vkBeginCommandBuffer nodes

GrantComm:
Adds draw call count to vkBeginCommandBuffer nodes and renderpass nodes.

Screenshot 2025-09-08 at 2 06 54 PM

### DIFF
--- a/gfxr_ext/decode/dive_annotation_processor_test.cpp
+++ b/gfxr_ext/decode/dive_annotation_processor_test.cpp
@@ -321,49 +321,52 @@ TEST(WriteBlockEndTest, MixedCommandsOnlyCountDrawCalls)
     EXPECT_THAT(draw_counts_map.at(handle).render_pass_draw_call_counts, testing::ElementsAre(2));
 }
 
-TEST(WriteBlockEndTest, DrawCallsAreCountedBothInsideAndOutsideRenderPass)
+TEST(WriteBlockEndTest, DrawCallsAreCorrectlyCountedInMultipleRenderPasses)
 {
     DiveAnnotationProcessor processor;
     uint64_t                handle = 1001;
 
-    // Command Buffer 1 (handle 1001)
+    // Command Buffer (handle 1001)
     processor.WriteBlockEnd(CreateCommandData("vkBeginCommandBuffer", handle, 0, 1));
 
-    // Draw calls outside of a render pass
-    processor.WriteBlockEnd(CreateCommandData("vkCmdDraw", handle, 1, 2));
+    // Begin Render Pass 1
+    processor.WriteBlockEnd(CreateCommandData("vkCmdBeginRenderPass", handle, 0, 2));
 
-    // Begin render pass
-    processor.WriteBlockEnd(CreateCommandData("vkCmdBeginRenderPass", handle, 0, 3));
-
-    // Draw calls inside the render pass
+    // Draw calls inside Render Pass 1
+    processor.WriteBlockEnd(CreateCommandData("vkCmdDraw", handle, 1, 3));
     processor.WriteBlockEnd(CreateCommandData("vkCmdDraw", handle, 2, 4));
-    processor.WriteBlockEnd(CreateCommandData("vkCmdDraw", handle, 3, 5));
 
-    // End render pass
-    processor.WriteBlockEnd(CreateCommandData("vkCmdEndRenderPass", handle, 0, 6));
+    // End Render Pass 1
+    processor.WriteBlockEnd(CreateCommandData("vkCmdEndRenderPass", handle, 0, 5));
 
-    // More draw calls outside of a render pass
-    processor.WriteBlockEnd(CreateCommandData("vkCmdDraw", handle, 4, 7));
+    // Begin Render Pass 2
+    processor.WriteBlockEnd(CreateCommandData("vkCmdBeginRenderPass", handle, 0, 6));
 
-    processor.WriteBlockEnd(CreateCommandData("vkEndCommandBuffer", handle, 0, 8));
+    // Draw calls inside Render Pass 2
+    processor.WriteBlockEnd(CreateCommandData("vkCmdDraw", handle, 3, 7));
+    processor.WriteBlockEnd(CreateCommandData("vkCmdDraw", handle, 4, 8));
+    processor.WriteBlockEnd(CreateCommandData("vkCmdDraw", handle, 5, 9));
+
+    // End Render Pass 2
+    processor.WriteBlockEnd(CreateCommandData("vkCmdEndRenderPass", handle, 0, 10));
+
+    processor.WriteBlockEnd(CreateCommandData("vkEndCommandBuffer", handle, 0, 11));
 
     // Single submit
     nlohmann::ordered_json args_submit_1 = {
         { "submitCount", 1 },
         { "pSubmits", { { { "commandBufferCount", 1 }, { "pCommandBuffers", { 1001 } } } } }
     };
-    gfxrecon::util::DiveFunctionData submit_data_1("vkQueueSubmit", 0, 9, args_submit_1);
+    gfxrecon::util::DiveFunctionData submit_data_1("vkQueueSubmit", 0, 12, args_submit_1);
     processor.WriteBlockEnd(submit_data_1);
 
     auto draw_counts_map = processor.TakeDrawCallMap();
     ASSERT_THAT(draw_counts_map, SizeIs(1));
     ASSERT_TRUE(draw_counts_map.count(handle));
 
-    // The total draw call count for the command buffer should be 4.
-    EXPECT_THAT(draw_counts_map.at(handle).begin_command_buffer_draw_call_count, 4);
-
-    // The draw call count for the render pass should be 2.
-    EXPECT_THAT(draw_counts_map.at(handle).render_pass_draw_call_counts, testing::ElementsAre(2));
+    EXPECT_THAT(draw_counts_map.at(handle).begin_command_buffer_draw_call_count, 5);
+    EXPECT_THAT(draw_counts_map.at(handle).render_pass_draw_call_counts,
+                testing::ElementsAre(2, 3));
 }
 
 }  // namespace


gfxr_ext/decode/dive_annotation_processor.cpp
```diff
@@ -61,9 +61,18 @@ void DiveAnnotationProcessor::WriteBlockEnd(const gfxrecon::util::DiveFunctionDa
             if (vkCmd.name == "vkBeginCommandBuffer")
             {
                 m_cmd_vk_commands_cache[cmd_handle].clear();
+                m_cmd_buffer_draw_call_counts_map[cmd_handle].push_back(0);
             }
 
             m_cmd_vk_commands_cache[cmd_handle].push_back(vkCmd);
+
+            if (function_name.find("vkCmdDraw") != std::string::npos)
```
wangra-google, line 69: there are more vk events that we need consider as drawcall:
vkCmdDrawIndexed
vkCmdDrawIndirect
vkCmdDrawIndexedIndirect
vkCmdDrawIndirectCount
vkCmdDrawIndexedIndirectCount

dive_core/gfxr_capture_data.h
```diff
@@ -57,7 +58,8 @@ class GfxrCaptureData : public CaptureData
     // Vector of SubmitInfo objects used to add the GFXR vulkan commands to the UI.
     std::vector<std::unique_ptr<DiveAnnotationProcessor::SubmitInfo>> m_gfxr_submits;
     std::unordered_map<uint64_t, std::vector<DiveAnnotationProcessor::VulkanCommandInfo>>
-    m_gfxr_command_buffers;
+                                                        m_gfxr_command_buffers;
+    std::unordered_map<uint64_t, std::vector<uint64_t>> m_gfxr_command_buffer_draw_counts;
```
wangra-google, line 62: is this m_gfxr_command_buffer_draw_counts used?

gfxr_ext/decode/dive_annotation_processor.h
```diff
@@ -78,11 +78,16 @@ class DiveAnnotationProcessor : public gfxrecon::decode::AnnotationHandler
...
+        return std::move(m_cmd_buffer_draw_call_counts_map);
+    }
 
 private:
     // This is a per submit cache that keeps all vk commands that are not in any command buffer
     std::vector<VulkanCommandInfo> m_none_cmd_vk_commands_per_submit_cache = {};
     // Use command buffer handle as the key to accociate with vk commands
     std::unordered_map<uint64_t, std::vector<VulkanCommandInfo>> m_cmd_vk_commands_cache = {};
-    std::vector<std::unique_ptr<SubmitInfo>>                     m_submits = {};
+    std::unordered_map<uint64_t, std::vector<uint64_t>> m_cmd_buffer_draw_call_counts_map = {};
```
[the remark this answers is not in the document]
wangra-google: or is this related later change for renderpass drawcall count?

gfxr_ext/decode/dive_annotation_processor_test.cpp
```diff
@@ -169,5 +169,255 @@ TEST(WriteBlockEndTest, MultipleSubmitsWithCommandsCreatesSubmitsWithCorrectComm
...
                 VulkanCommandInfoEqual(cmd_data_7.GetFunctionName(), 0, args_cmd_7));
 }
+
+TEST(WriteBlockEndTest,
+     SingleSubmitWithSingleCommandBufferHasCorrectNumberOfDrawCallsPerCommandBufferRecord)
+{
+    DiveAnnotationProcessor processor;
+
+    // Command Buffer 1 (handle 1001)
+    nlohmann::ordered_json           args_cmd_1 = { { "commandBuffer", 1001 } };
```
wangra-google, line 179: i feel it could be cleaner if we have a helper function like:
```
gfxrecon::util::DiveFunctionData CreateCommandData(const std::string& name,
                                                   uint64_t command_buffer_handle,
                                                   uint32_t cmd_buffer_index,
                                                   uint32_t block_index)
{
    nlohmann::ordered_json args = { { "commandBuffer", command_buffer_handle } };
    return gfxrecon::util::DiveFunctionData(name,
                                            cmd_buffer_index,
                                            block_index,
                                            args);
}
```

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -133,12 +148,17 @@ const DiveAnnotationProcessor::VulkanCommandInfo &vk_cmd_info)
 
 //--------------------------------------------------------------------------------------------------
 bool GfxrVulkanCommandHierarchyCreator::ProcessVkCmds(
-const std::vector<DiveAnnotationProcessor::VulkanCommandInfo> &vkCmds)
+const std::vector<DiveAnnotationProcessor::VulkanCommandInfo> &vkCmds,
+const std::vector<uint64_t>                                   &draw_call_counts,
+const std::vector<uint64_t>                                   &render_pass_draw_call_counts)
 {
+    std::vector<uint64_t> mutable_draw_call_counts = draw_call_counts;
```
wangra-google, line 155: why not just pass in std::vector<uint64_t>& as arg directly?

gfxr_ext/decode/dive_annotation_processor.cpp
```diff
@@ -56,14 +56,44 @@ void DiveAnnotationProcessor::WriteBlockEnd(const gfxrecon::util::DiveFunctionDa
...
+            if (vkCmd.name.find("vkCmdDraw") != std::string::npos)
+            {
+                if (!m_draw_call_counts_map[cmd_handle].first.empty())
+                {
+                    m_draw_call_counts_map[cmd_handle].first.back()++;
+                }
+
+                if (!m_render_pass_draw_call_counts_stack[cmd_handle].empty())
+                {
+                    std::cout << "RP Draw Count: "
```
wangra-google, line 91: is this for debugging only? i notice that you dont have this above for draw count in cmd buffer

gfxr_ext/decode/dive_annotation_processor.h
```diff
@@ -78,11 +79,19 @@ class DiveAnnotationProcessor : public gfxrecon::decode::AnnotationHandler
...
+        return std::move(m_draw_call_counts_map);
+    }
 
 private:
     // This is a per submit cache that keeps all vk commands that are not in any command buffer
     std::vector<VulkanCommandInfo> m_none_cmd_vk_commands_per_submit_cache = {};
     // Use command buffer handle as the key to accociate with vk commands
     std::unordered_map<uint64_t, std::vector<VulkanCommandInfo>> m_cmd_vk_commands_cache = {};
-    std::vector<std::unique_ptr<SubmitInfo>>                     m_submits = {};
+    std::unordered_map<uint64_t, std::pair<std::vector<uint64_t>, std::vector<uint64_t>>>
```
wangra-google, line 93: better to use struct instead of pair

gfxr_ext/decode/dive_annotation_processor.cpp
GrantComm: Removed.

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: I created the mutable counts since the const vectors belong to the const GfxrCaptureData. I could make GfxrCaptureData passed to ProcessGfxrSubmits() non-const which would allow me to pass in std::vector<uint64_t>& as arg directly

gfxr_ext/decode/dive_annotation_processor_test.cpp
GrantComm: Added.

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -48,14 +48,17 @@ void GfxrVulkanCommandHierarchyCreator::ConditionallyAddChild(uint64_t node_inde
...
+std::vector<uint64_t>                             draw_call_counts)
 {
     const std::string            &vulkan_cmd_name = vk_cmd_info.name;
     const nlohmann::ordered_json &vulkan_cmd_args = vk_cmd_info.args;
     std::ostringstream            vk_cmd_string_stream;
     vk_cmd_string_stream << vulkan_cmd_name;
     if (vulkan_cmd_name == "vkBeginCommandBuffer")
     {
+        vk_cmd_string_stream << ", Draw Call Count: " << std::to_string(draw_call_counts.front());
+        draw_call_counts.erase(draw_call_counts.begin());
```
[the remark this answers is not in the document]
GrantComm: This removes the top most element in the vector. This is so the count is always the most updated for the vkBeginRenderPass and vkBeginRenderPass nodes

gfxr_ext/decode/dive_annotation_processor.h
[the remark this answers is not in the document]
GrantComm: Updated the pr to include the count in the renderpass node as well.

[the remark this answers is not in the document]
wangra-google: cool, thanks.

gfxr_ext/decode/dive_annotation_processor.h
```diff
@@ -59,6 +60,12 @@ class DiveAnnotationProcessor : public gfxrecon::decode::AnnotationHandler
         std::string           name = "";
     };
 
+    struct DrawCallCounts
+    {
+        std::vector<uint64_t> begin_command_buffer_draw_call_counts = {};
```
wangra-google, line 65: sorry i still dont quite get why begin_command_buffer_draw_call_counts is a vector. m_draw_call_counts_map is a hashmap with cmd_handle as the key, so DrawCallCounts is corresponding to 1 cmd buffer. and if this is the case, each cmd buffer should have only 1 count?

GrantComm: Updated. Originally had as vector due to previous iteration where there was a bug in the tree view.

gfxr_ext/decode/dive_annotation_processor.cpp
```diff
@@ -56,14 +56,38 @@ void DiveAnnotationProcessor::WriteBlockEnd(const gfxrecon::util::DiveFunctionDa
...
             uint64_t cmd_handle = args["commandBuffer"];
 
-            // There could be cases where vkBeginCommandBuffer is called after vkEndCommandBuffer
-            // without having the command buffer submitted
-            if (vkCmd.name == "vkBeginCommandBuffer")
+            if (vkCmd.name.find("vkBeginCommandBuffer") != std::string::npos)
             {
                 m_cmd_vk_commands_cache[cmd_handle].clear();
+
+                // Total draw count for this command buffer
```
wangra-google, line 63: nit: i feel this comment should be put where we declare `begin_command_buffer_draw_call_count` rather than here. Here is more like to clear the count because we have encountered cases where vkBeginCommandBuffer is called multiple times for the same cmd within a frame, so we need to clear the count here whenever vkBeginCommandBuffer is called.

gfxr_ext/decode/dive_annotation_processor.cpp
```diff
@@ -56,14 +56,38 @@ void DiveAnnotationProcessor::WriteBlockEnd(const gfxrecon::util::DiveFunctionDa
...
+            }
+            else if (vkCmd.name.find("vkCmdBeginRenderPass") != std::string::npos)
+            {
+                m_render_pass_draw_call_counts_stack[cmd_handle].push(0);
+            }
+            else if (vkCmd.name.find("vkCmdEndRenderPass") != std::string::npos)
+            {
+                if (!m_render_pass_draw_call_counts_stack[cmd_handle].empty())
+                {
+                    uint64_t rp_draw_count = m_render_pass_draw_call_counts_stack[cmd_handle].top();
```
wangra-google, line 74: out of curiosity, why do you need a stack here? renderpass cannot be nested.

GrantComm: Ah ok, removed the use of the stack and use the renderpass draw count vector directly instead.

wangra-google (approved):

angela28chen (approved):
Thanks for writing additional unit tests for annotation processor write block end!

PR #356, opened 2025-09-05
Process metrics data to get the average across loops

RenfengLiu:


dive_core/perf_metrics_data.cpp
```diff
@@ -21,17 +21,21 @@
...
 
 namespace Dive
 {
+
 namespace
 {
 constexpr std::array kFixedHeaders = { "ContextID",   "ProcessID", "FrameID",
                                        "CmdBufferID", "DrawID",    "DrawType",
                                        "DrawLabel",   "ProgramID", "LRZState" };
+const std::string    kEmptyString;
```
footballhead, line 38: Use of static objects that aren't `constexpr` could introduce subtle bugs down the road since their initialization and destruction is not well defined (see [Static initialization fiasco](https://en.cppreference.com/w/cpp/language/siof.html), https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables). I'd consider alternatives like returning a `std::string_view`  from `GetMetricsDescription` (which removes the need for `kEmptyString` altogether).

wangra-google (approved):

wangra-google (approved):

dive_core/perf_metrics_data.cpp
```diff
@@ -185,20 +194,208 @@ const AvailableMetrics&      available_metrics)
...
+    return metrics_info[metric_index]->m_description;
+}
+
+size_t PerfMetricsDataProvider::GetDrawCountForCommandBuffer(size_t command_buffer_index) const
+{
+    if (command_buffer_index >= m_cmd_buffer_list.size())
+        return 0;
+
+    uint64_t cmd_id = m_cmd_buffer_list[command_buffer_index];
+    return m_cmd_buffer_to_draw_id.at(cmd_id).size();
```
footballhead, line 399: FYI: that `at()` will have different behavior based on whether exceptions are enabled or not (could throw, could crash) so (at least internally we prefer that) other options could be considered like `find` with an explicit check for `find() == end()`. This is the guidance for vector since (unfortunately) the other tips for maps are internal only: https://abseil.io/tips/224. However you should be able to find them

PR #357, opened 2025-09-08
Dive communication protocol: Unit tests for client and unix domain server

elviscapiaq:


network/tcp_client_test.cc
```diff
@@ -0,0 +1,286 @@
...
+{
+    return (std::filesystem::temp_directory_path() / "test_client_file.tmp").string();
+}
+
+// The tests assume the client and server are running on a PC (Linux), as we are not using
+// an Android device. However, this creates a limitation: we cannot directly test the TcpClient
+// against the UnixDomainServer because they use incompatible communication protocols. The TcpClient
+// uses TCP/IP sockets, while the UnixDomainServer uses Unix domain sockets (UDS). To test the
+// TcpClient, we create a minimal TCP server.
+class TestTcpServer
```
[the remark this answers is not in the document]
elviscapiaq: Thanks for all the feedback, I created a fake `TestTcpServer`.

elviscapiaq:
> At a high-level, I think this is a good approach since we should strive to write tests using the real implementation. ([Google SWE Handbook, chapter 18.e.i](https://abseil.io/resources/swe-book/html/ch13.html#real_implementations)) The main arguments against mocking from that chapter are:
> 
> * Using the real implementation provides us with more confidence that the code we are shipping to users is reliable.
> * When the real implementation changes, so too can the mocks. This means mocks add additional maintenance overhead. This can also make refactoring harder.
> 
> [Chapter 18: Test Doubles](https://abseil.io/resources/swe-book/html/ch13.html#real_implementations) goes into options that you can consider if testing the real implementation is not feasible. In general, they encourage looking at other options first instead of mocks, e.g. fakes. `TestTcpServer` is a good example of a fake.
> 
> I'd encourage reading Chapters 16-19 of the [SWE handbook](https://abseil.io/resources/swe-book/html/toc.html)

Thanks for poiting it out. I have read all these good practices and made a commit based on that.
Now we are faking the `TestTcpServer` as this is the recommended approach for this case.

elviscapiaq:
I am restarting this commit applying all the recommendations and after reading the tips about how to use a fake test. I found it very interesting and useful

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,180 @@
...
+
+const std::string kTestServerAddress = "test_server_socket";
+
+// The tests assume the client and server are running on a PC (Linux), as we are not using
+// an Android device. However, this creates a limitation: we cannot directly test the TcpClient
+// against the UnixDomainServer because they use incompatible communication protocols. The TcpClient
+// uses TCP/IP sockets, while the UnixDomainServer uses Unix domain sockets (UDS). To test the
+// UnixDomainServer, we will start an instance of it and use a basic, raw socket client to verify
+// its behavior, such as handling connections and basic messages.
+class UnixDomainServerTest : public ::testing::Test
```
footballhead: Avoid test fixtures when reasonable. Prefer to put all test logic into the test body and use helper free functions for things that are complicated or don’t contribute to the code that is being tested. This makes it clearer to read and understand since everything in the test is explicit.

In this case, I'd recommend putting `auto server = std::make_unique<UnixDomainServer>();` into the test body, along with the call to `Stop`. `ConnectClient()` can easily be turned into a free function since it doesn't use `server`. This makes sense too since we're not testing `SocketConnection` in this file, so hiding the details doesn't hamper test understandability.

https://abseil.io/tips/122

network/CMakeLists.txt
```diff
@@ -60,4 +60,21 @@ if(NOT ANDROID)
             absl::status_matchers
     )
     gtest_discover_tests(messages_test)
+
+    # The UnixDomainServer listens only on UNIX OS (Android/Linux).
+    if(NOT WIN32 AND NOT APPLE)
```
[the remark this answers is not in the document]
footballhead: Go it. Thanks for the thorough explanation

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,190 @@
...
+#include "unix_domain_server.h"
+
+#include <gtest/gtest.h>
+
+namespace Network
+{
+namespace
+{
+
+const std::string kTestServerAddress = "test_server_socket";
```
footballhead: Avoid static variables with non-trivial constructors and destructors. For strings, prefer to use `constexpr std::string_view` or `constexpr char[]`
```suggestion
constexpr std::string_view kTestServerAddress = "test_server_socket";
```

The order of initialization and destruction is not well defined which could lead to subtle bugs: accessing a variable before it has been initialized, or after it has been destroyed. See the [Google C++ Style guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables) and [Tip of the Week #140: Constants: Safe Idioms](https://abseil.io/tips/140)

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,190 @@
...
+    absl::StatusOr<std::unique_ptr<ISerializable>> response_msg = ReceiveSocketMessage(
+    (*client_conn).get());
+    ASSERT_TRUE(response_msg.ok());
+    ASSERT_NE(*response_msg, nullptr);
+    ASSERT_EQ((*response_msg)->GetMessageType(), MessageType::PONG_MESSAGE);
+
+    server->Stop();
+}
+
+TEST_F(UnixDomainServerTest, PingPongFails)
```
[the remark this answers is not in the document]
elviscapiaq: Updated.

network/tcp_client_test.cc
```diff
@@ -0,0 +1,267 @@
...
+    EXPECT_EQ(*result, "/var/log/capture.pm4");
+}
+
+TEST_F(TcpClientFakeTest, DisconnectStopsKeepAlive)
+{
+    EXPECT_TRUE(m_client->IsConnected());
+    m_client->Disconnect();
+    EXPECT_FALSE(m_client->IsConnected());
+    // Wait a bit to ensure no crashes from background threads
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
footballhead: Please don't sleep in tests. Try to replace the sleep with a more explicit wait.

If necessary, prefer to use abseil time library
```suggestion
    absl::SleepFor(absl::Milliseconds(100));
```

network/tcp_client_test.cc
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    std::shared_ptr<FakeSocketConnection> m_conn;
+    std::atomic<bool>                     m_running{ false };
+    std::thread                           m_thread;
+    RequestHandler                        m_handler;
+    std::mutex                            m_handler_mutex;
+};
+
+class TcpClientFakeTest : public ::testing::Test
```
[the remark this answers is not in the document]
elviscapiaq: Updated.

network/tcp_client_test.cc
```diff
@@ -0,0 +1,267 @@
...
+protected:
+    void SetUp() override
+    {
+        m_client_fake = new FakeSocketConnection();
+        m_server_fake = std::make_shared<FakeSocketConnection>();
+        m_client_fake->PairWith(m_server_fake.get());
+
+        // Inject the pre-created fake into the client
+        m_client = std::make_unique<TcpClient>(
+        [this]() { return std::unique_ptr<FakeSocketConnection>(m_client_fake); });
```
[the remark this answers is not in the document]
elviscapiaq: Updated.

wangra-google (approved):

network/tcp_client_test.cc
```diff
@@ -0,0 +1,289 @@
...
+    constexpr size_t kExpectedSize = 12345;
+    MessageType      received_type = MessageType::UNKNOWN;
+
+    ctx.server.SetHandler([&](FakeSocketConnection* conn, std::unique_ptr<ISerializable> req) {
+        received_type = req->GetMessageType();
+
+        FileSizeResponse resp;
+        resp.SetFound(true);
+        resp.SetFileSizeStr(std::to_string(kExpectedSize));
+        (void)SendSocketMessage(conn, resp);
```
footballhead: nit: For absl::Status, the preferred way marking that the return type is unused this is with [IgnoreError()](https://abseil.io/docs/cpp/guides/status#ignoring-status-results)
```suggestion
        SendSocketMessage(conn, resp).IgnoreError();
```

wangra-google (dismissed):

wangra-google (dismissed):

wangra-google (approved):

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,209 @@
...
+{
+    SocketType client_socket = socket(AF_UNIX, SOCK_STREAM, 0);
+    if (client_socket < 0)
+    {
+        return absl::InternalError("Client: socket creation failed.");
+    }
+    /// Connects to an Unix (Local) Domain with abstract namespace.
+    sockaddr_un addr = {};
+    addr.sun_family = AF_UNIX;
+    // first char is '\0'
```
footballhead: [Let the code speak for itself](https://testing.googleblog.com/2023/12/let-code-speak-for-itself.html). Comments should focus on the why, not the what. The what should be obvious from reading the code. If it's not then that's a good sign that it should be refactored
```suggestion
```

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,209 @@
...
+    server->Wait();
+}
+
+TEST(UnixDomainServerTest, ClientConnectAndDisconnect)
+{
+    auto server = std::make_unique<UnixDomainServer>();
+    ASSERT_TRUE(server->Start(kTestServerAddress).ok());
+
+    absl::StatusOr<std::unique_ptr<SocketConnection>> client_conn = ConnectClient();
+    ASSERT_TRUE(client_conn.ok());
```
footballhead: Prefer using the `absl_testing::IsOk()` matcher since it will print more information if something fails
```suggestion
#include "absl/status/status_matchers.h"
using ::absl_testing::IsOk;
// ...
    ASSERT_THAT(client_conn.ok(), IsOk());
```

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,209 @@
...
+    (*client_conn)->Close();
+    EXPECT_FALSE((*client_conn)->IsOpen());
+
+    server->Stop();
+}
+
+TEST(UnixDomainServerTest, HandShakeSuccess)
+{
+    auto server = std::make_unique<UnixDomainServer>();
+    ASSERT_TRUE(server->Start(kTestServerAddress).ok());
```
footballhead: Prefer using the absl_testing::IsOk() matcher since it will print more information if something fails

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,209 @@
...
+
+    absl::StatusOr<std::unique_ptr<SocketConnection>> client_conn = ConnectClient();
+    ASSERT_TRUE(client_conn.ok());
```
footballhead: Prefer using the absl_testing::IsOk() matcher since it will print more information if something fails

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,209 @@
...
+
+    // Client sends request.
+    HandshakeRequest request;
+    request.SetMajorVersion(5);
+    request.SetMinorVersion(2);
+    ASSERT_TRUE(SendSocketMessage((*client_conn).get(), request).ok());
```
footballhead: Prefer using the absl_testing::IsOk() matcher since it will print more information if something fails

network/unix_domain_server_test.cc
```diff
@@ -0,0 +1,209 @@
...
+
+    // Client receives response.
+    absl::StatusOr<std::unique_ptr<ISerializable>> response_msg = ReceiveSocketMessage(
+    (*client_conn).get());
+    ASSERT_TRUE(response_msg.ok());
```
footballhead: Prefer using the absl_testing::IsOk() matcher since it will print more information if something fails

PR #360, opened 2025-09-09
Updates download location of replay artifacts 

GrantComm:
Download files are now saved to location of replayed capture file.

### DIFF
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -595,6 +595,7 @@ absl::Status DeviceManager::RunReplayApk(const std::string &capture_path,
             std::this_thread::sleep_for(std::chrono::seconds(1));
         } while (m_device->IsProcessRunning(kGfxrReplayAppName));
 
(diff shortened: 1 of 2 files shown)


wangra-google (dismissed):

wangra-google (approved):

angela28chen (commented):
Is the issue with the gpu_time.csv collision addressed here? (file overwriting with timing files produced from different gfxr captures being put in the same original capture directory)

GrantComm:
> > > Is the issue with the gpu_time.csv collision addressed here? (file overwriting with timing files produced from different gfxr captures being put in the same original capture directory)
> > 
> > 
> > No this pr just ensures the artifacts are saved to the original capture directory.
> 
> OK, in that case can you file a P2 for this and add a reference to that in the code? I want to have a reminder that we need to address the collision issue in the future (I can see this being an issue if any users store gfxr captures in the same folder, the gpu_time csv will be ambiguous to which it belongs)

Sure thing, added todos.

wangra-google (approved):

PR #370, opened 2025-09-14
Forcing all pools to use VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT

wangra-google:
This is needed because without this flag, vkBeginCommandBuffer will NOT perform an implicit reset if the command buffer is not in the Initial state. Some of the command buffers are not created with this flag, but for the case of looping frames, calling vkBeginCommandBuffer on those cmds would cause VK_ERROR_INITIALIZATION_FAILED.

Note that there might be performance impact on this which needs more investigation (b/444976541)

### DIFF
--- a/capture_service/constants.h
+++ b/capture_service/constants.h
@@ -43,6 +43,9 @@ inline constexpr char kDefaultCaptureFolderName[] = "gfxr_capture";
 inline constexpr char kDefaultReplayFolderName[] = "gfxr_replay_downloads";
 inline constexpr char kProfilingPluginFolderName[] = "dive_profiling_plugin";
 inline constexpr char kProfilingPluginName[] = "dive_drawcall_metrics";
+// This file will be created by replay when it has completed trim state loading. /sdcard/Download/
+// is the base path since GFXR can reliably write there.
+inline constexpr char kReplayStateLoadedSignalFile[] = "/sdcard/Download/replay_state_loaded";
 
 inline constexpr int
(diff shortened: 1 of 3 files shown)


gfxr_ext/decode/dive_vulkan_replay_consumer.cpp
```diff
@@ -196,6 +196,31 @@ void DiveVulkanReplayConsumer::Process_vkResetCommandBuffer(const ApiCallInfo&
...
 
+void DiveVulkanReplayConsumer::Process_vkCreateCommandPool(
+const ApiCallInfo&                                     call_info,
+VkResult                                               returnValue,
+format::HandleId                                       device,
+StructPointerDecoder<Decoded_VkCommandPoolCreateInfo>* pCreateInfo,
+StructPointerDecoder<Decoded_VkAllocationCallbacks>*   pAllocator,
+HandlePointerDecoder<VkCommandPool>*                   pCommandPool)
+{
+    // Forcing all pools to use VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT.
```
footballhead, line 207: Thank you for the detailed comments!

footballhead (approved):
Happy to see that we've been able to fix some issues without touching GFXR code

angela28chen (approved):

gfxr_ext/decode/dive_vulkan_replay_consumer.cpp
wangra-google: np, glad that helps :)

PR #372, opened 2025-09-15
Send node index for DrawDispatchNode.

hysw:


hysw:
See #373 apparently there's a related bug, so prefer more comprehensive solution.

PR #376, opened 2025-09-16
Add a vkDeviceWaitIdle at the frame boundary

wangra-google:
This is needed since when we loop the frame, if the cpu is too fast, it might start to write to cmd while gpu is using it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,8 @@ cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 
 project(dive)
 
+option(DIVE_UI_WIN32 "(Windows-only) Build Dive UI as WIN32 GUI application instead of a console application. This will supress logging to stdout." ON)
+
 # Global properties
 set_property(GLOBAL PROPERTY USE_FOLDERS ON) # Defaults to ON with cmake 3.26+
 
--- a/capture_service/constants.h
+++ b/capture_service/constants.h
@@ -46,6 +46,9 @@ inline constexpr char kProfilingPluginName[] = "dive_drawcall_metrics";
 // This file will be created by replay when it has completed trim state loading. /sdcard/Download/
 // is the base path since GFXR can reliably write there.
 inline constexpr char kReplayStateLoadedSignalFile[] = "/sdcard/Download/replay_state_loaded";
+inline constexpr char kGpuTimingFile[] = "gpu_time.csv";  // produced by GFXR replay
+inline constexpr char kProfilingMetricsCsvSuffix[] = "_profiling_metrics.csv";
+inline constexpr char kGpuTimingCsvSuffix[] = "_gpu_time.csv";
 
 inline constexpr int
 kFirstPort = 49391;  // A port number within the dynamic port range (49152 to 65535)
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -508,7 +508,7 @@ absl::Status DeviceManager::DeployReplayApk(const std::string &serial)
 absl::Status DeviceManager::RunReplayApk(const std::string &capture_path,
                                          const std::string &replay_args,
                                          bool               dump_pm4,
-                                         const std::string &download_path)
+                                         const std::string &local_download_dir)
 {
     LOGD("RunReplayApk(): starting\n");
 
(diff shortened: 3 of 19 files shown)


footballhead:
For metrics collection, I'm still seeing crashes due to asserts with this PR applied. I don't see asserts if I launch replay with `--enable-gpu-time`

gfxr_ext/decode/dive_vulkan_replay_consumer.cpp
```diff
@@ -323,6 +325,15 @@ format::HandleId                            fence)
                                                  ResetQueryPool,
                                                  GetQueryPoolResults);
 
+    // vkDeviceWaitIdle is needed since when we loop the frame, we do not double/triple buffer cmds.
+    // If the CPU is too fast, it might start to write to cmd while GPU is using it which would
+    // cause random crashes. (No need to add device wait idle when gpu time is enabled since it is
+    // called in GPUTime::OnQueueSubmit)
+    if (!gpu_time_.IsEnabled() && submit_status.contains_frame_boundary)
```
footballhead, line 332: If GPUTime isn't enabled then we never get a frame boundary.

```c++
GPUTime::SubmitStatus GPUTime::OnQueueSubmit(uint32_t                  submit_count,
                                             const VkSubmitInfo*       submits_ptr,
                                             PFN_vkDeviceWaitIdle      pfn_device_wait_idle,
                                             PFN_vkResetQueryPool      pfn_reset_query_pool,
                                             PFN_vkGetQueryPoolResults pfn_get_query_pool_results)
{
    if (!m_enable)
    {
        return { GPUTime::GpuTimeStatus(), false };
    }

```

gfxr_ext/decode/dive_vulkan_replay_consumer.cpp
```diff
@@ -323,6 +325,40 @@ format::HandleId                            fence)
...
+
+                    if (command_buffer_info->is_frame_boundary)
+                    {
+                        is_frame_boundary = true;
+                        break;
+                    }
+                }
+            }
+        }
+    }
```
footballhead, line 351: Three things that stand out:

0. Only need to check for `submit_info_data != nullptr` once. Currently, this checks twice. Maybe the second is supposed to be a check for `submit_info_data[i] != nullptr` ?
1. Consider [reducing nesting to reduce complexity](https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html). Checking for invalid inputs using guard clauses tends to be the approach here. Some complexity metrics like clang-tidy's [readability-function-cognitive-complexity](https://clang.llvm.org/extra/clang-tidy/checks/readability/function-cognitive-complexity.html) check use nesting as an input for their code complexity algorithms
2. Consider refactoring this into a helper function. It's more self-documenting, and replacing the `break` with a `return true` should improve performance (as written, all submits will be processed even if the frame boundary has already been found)

Something like:

```c++
// If we had c++20, this would be a good use for a std::span. Could consider an absl::Span
bool IsFrameBoundary(Decoded_VkSubmitInfo* submit_info_data, uint32_t submit_count, CommonObjectInfoTable& object_info_table)
{
    // Use a guard clause to reduce nesting
    if (submit_info_data == nullptr)
    {
        return false;
    }

    for (uint32_t i = 0; i < submit_count; ++i)
    {
        // Fix check and use a guard clause to reduce nesting
        if (submit_info_data[i] == nullptr)
        {
            continue;
        }
            
        size_t     command_buffer_count = submit_info_data[i].pCommandBuffers.GetLength();
        const auto command_buffer_ids = submit_info_data[i].pCommandBuffers.GetPointer();
        for (uint32_t j = 0; j < command_buffer_count; ++j)
        {
            auto command_buffer_info = object_info_table.GetVkCommandBufferInfo(
            command_buffer_ids[j]);

            if (command_buffer_info->is_frame_boundary)
            {
                // Using return here will break from both for loops, not just the inner one
                return true;
            }
        }
    }

    return false;
}
```

Then this code becomes

```suggestion
    const bool is_frame_boundary = IsFrameBoundary(pSubmits->GetMetaStructPointer(), submitCount, GetObjectInfoTable());
```

wangra-google: Thanks for the suggestions. updated

angela28chen (approved):

PR #378, opened 2025-09-16
Adds frame node for mixed hierarchy

GrantComm:
Adds frame node for mixed hierarchy. Correlation for gpu timing in mixed hierarchy will be functional after UI reorganization. It is currently tied to the command hierarchy view.


angela28chen (approved):
Thanks for making this change!

wangra-google (approved):

PR #388, opened 2025-09-18
Remove temporary artifacts from previous GFXR replays

angela28chen:
Since we moved to using the GFXR file load folder as the output folder for replay artifacts, currently there is an issue where subsequent replays will not have the proper artifacts displayed in the UI because we have improper cleanup and reload.

This PR addresses the issue for three files (profiling metrics CSV, gpu timing CSV, and PM4 capture) and adds a warning in the UI for the user to manually save any temporary files that they don't want cleared on the next replay. In the future when we have .dive files, more of this will be hidden from the user for ease of use.

Changes:
- Deleting temp files from earlier replay runs upon initiation of new replay
- Force reload of GFXR (+rd if present) after replay has completed.
- Adding support for perf counter model reset with no CSV present
- Added replay warning text in UI Analyze Capture menu
image

Minor changes:
- Refactoring for updating tab views and getting path names
- Extra debug logging around file handling

### DIFF
--- a/ui/analyze_window.cpp
+++ b/ui/analyze_window.cpp
@@ -577,13 +577,12 @@ void AnalyzeDialog::SetReplayButton(const std::string &message, bool is_enabled)
 std::filesystem::path AnalyzeDialog::GetFullLocalPath(const std::string &gfxr_stem,
                                                       const std::string &suffix) const
 {
-    if (!m_local_capture_file_directory.ok())
+    if (m_local_capture_file_directory.empty())
     {
         return "";
     }
 
-    std::filesystem::path full_path = m_local_capture_file_directory.value();
-    full_path /= (gfxr_stem + suffix);
+    std::filesystem::path full_path = m_local_capture_file_directory / (gfxr_stem + suffix);
     return full_path;
 }
 
@@ -606,7 +605,7 @@ absl::Status AnalyzeDialog::Pm4Replay(Dive::DeviceManager &device_manager,
     return device_manager.RunReplayApk(remote_gfxr_file,
                                        replay_args,
                                        m_dump_pm4_enabled,
-                                       m_local_capture_file_directory.value());
+                                       m_local_capture_file_directory.string());
 }
 
 //--------------------------------------------------------------------------------------------------
@@ -617,7 +616,7 @@ absl::Status AnalyzeDialog::PerfCounterReplay(Dive::DeviceManager &device_manage
 
     return device_manager.RunProfilingOnReplay(remote_gfxr_file,
                                                *m_enabled_settings_vector,
-                                               m_local_capture_file_directory.value());
+                                               m_local_capture_file_directory.string());
 }
 
 //--------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


angela28chen:
> > Notes:
> > 
> > 1. I think this should be done for PM4 as well, any thoughts on that?
> > 2. The UI text is ugly, suggestions on how to make it better are welcome
> 
> Maybe the warning could be a popup that appears when one of the artifact producing replay options is selected? Using the ShowErrorMessage(const std::string &err_msg); function. It could only popup once so the user isn't constantly dealing with popups if enabling more than one artifact producing replay option

The thing is the artifacts will be wiped if the replay options that don't produce them are selected. So every time you replay it would affect the temporary artifacts. That's why I'm thinking a popup isn't a good option unless there's some kind of "don't show me this again" checkbox? Is that the popup option you mentioned?

wangra-google:
> The thing is the artifacts will be wiped if the replay options that don't produce them are selected. So every time you replay it would affect the temporary artifacts. That's why I'm thinking a popup isn't a good option unless there's some kind of "don't show me this again" checkbox? Is that the popup option you mentioned?

Personally I feel that a persistent warning text is better than a popup. But I guess this is not that important since using files is just a temp solution, and will soon be replaced by Elvis' comm infrastructure, and with that we dont have this issue. So for now i think you can keep the warning message as it is (maybe different font/color or maybe with warning signs like ⚠️)

ui/analyze_window.cpp
```diff
@@ -740,12 +755,18 @@ void AnalyzeDialog::OnReplay()
...
+        emit OnDisplayPerfCounterResults(QString::fromStdString(perf_counter_csv.string()));
+    }
+    else
+    {
+        qDebug() << "Cleared perf counter data";
+        emit OnDisplayPerfCounterResults("");
     }
 
     // Run the gpu_time replay
+    std::string gpu_time_csv_stem = "";
```
wangra-google, line 769: do we use gpu_time_csv_stem some where?

ui/analyze_window.cpp
```diff
@@ -757,8 +778,14 @@ void AnalyzeDialog::OnReplay()
             SetReplayButton(kDefaultReplayButtonText, true);
             return;
         }
-        UpdateGpuTimingTabView(remote_file.value());
         WaitForReplay(*device);
+        qDebug() << "Loading gpu timing data: " << perf_counter_csv.string().c_str();
```
wangra-google, line 782: a bit confused here: are we loading gpu time or perf counter here?

ui/analyze_window.h
```diff
@@ -139,7 +142,7 @@ private slots:
     const int                   kDataRole = Qt::UserRole + 1;
     const int                   kDefaultFrameCount = 3;
     const std::string           kDefaultReplayButtonText = "Replay";
-    absl::StatusOr<std::string> m_capture_file_directory = "";
+    absl::StatusOr<std::string> m_local_capture_file_directory = "";
```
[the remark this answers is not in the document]
angela28chen: Good point, made it into std::filesystem::path instead

GrantComm (approved):

PR #391, opened 2025-09-19
Add unit test for AndroidTraceManager frame trace

footballhead:
This adds a unit test for #385. Without the fix, the test fails. With the fix, the test passes.

This additionally required:

- removing an unused #include
- splitting trace_mgr into its own target so that it could be more granularly included in the test


capture_service/android_trace_mgr_test.cc
```diff
@@ -0,0 +1,61 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "trace_mgr.h"
+
+#include "gtest/gtest.h"
+
+// AndroidTraceManager uses these funtions to talk with libwrap. They must be defined at link time.
```
footballhead, line 21: typo
```suggestion
// AndroidTraceManager uses these functions to talk with libwrap. They must be defined at link time.
```

capture_service/android_trace_mgr_test.cc
```diff
@@ -0,0 +1,61 @@
...
+    void SetCaptureState(int state) {}
+    void SetCaptureName(const char* name, const char* frame_num) {}
+}
+
+namespace Dive
+{
+namespace
+{
+
+TEST(AndroidTraceManagerTest, OnFrameBoundaryDetectedTraceByFrameForOneFrame)
```
elviscapiaq, line 34: I think it is quite descriptive but a bit long. Could we use `FrameTraceStopsAfterOneCapturedFrame` instead, given the purpose of this test if I am correct?

capture_service/CMakeLists.txt
```diff
@@ -28,11 +28,23 @@ else()
...
 endif()
 
+# AndroidTraceManager can be compiled on the host. This allows us to unit test it.
+# Group both cc files in the same target since the public header trace_mgr.h declares both.
+add_library(trace_mgr
+  android_trace_mgr.cc
+  trace_mgr.cc
+)
+target_link_libraries(trace_mgr
+  PUBLIC
```
elviscapiaq, line 38: I think in general we don't have a clear format we are using for CMakeLists.txt files. However, seems like we should at least start having the same indentation here?

angela28chen (dismissed):

capture_service/android_trace_mgr_test.cc
footballhead: The longer the name the better! This function is never called directly by the user and frequently read in the test report so the extra verboseness helps. This approach is recommended by the [SWE handbook](https://abseil.io/resources/swe-book/html/ch12.html#example_onetwo-onefourdot_some_sample_m)

capture_service/trace_mgr.h
```diff
@@ -65,6 +65,12 @@ class AndroidTraceManager : public TraceManager
     virtual void OnNewFrame() override;
     virtual void WaitForTraceDone() override;
 
+    TraceState GetState() ABSL_LOCKS_EXCLUDED(m_state_lock)
```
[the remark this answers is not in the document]
footballhead: This is done since `m_state` is accessed by 2 threads (not processes) inside the process being traced:

1. The thread triggering the trace, which writes to m_state to kick off tracing. This is currently the network server message handler thread.
2. The thread calling xrEndFrame, which reads and writes to m_state to perform work and state transitions. This could be the any application thread, e.g. the main rendering loop.

In order to prevent races, the thread accessing m_state needs to acquire the mutex beforehand and release it afterwards.

If it helps, absl::Mutex is very similar to std::mutex and absl::MutexLock is very similar to std::lock_guard.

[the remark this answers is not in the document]
angela28chen: Thank you for the explanation

PR #392, opened 2025-09-20
Aggregate perf counter metrics.

hysw:
Use pattern matching based correlation for metrics aggregation.


dive_core/perf_metrics_data.cpp
```diff
@@ -223,18 +235,269 @@ PerfMetricsData::PerfMetricsData(std::vector<std::string>          metric_names,
...
 
-std::unique_ptr<PerfMetricsDataProvider> PerfMetricsDataProvider::Create(
-std::unique_ptr<PerfMetricsData> data)
+class PerfMetricsDataProvider::Correlator
 {
-    if (!data)
-        return nullptr;
+public:
+    using SubmitRef = uint64_t;
+    using CommandBufferRef = uint64_t;
```
wangra-google, line 242: is this used?

dive_core/perf_metrics_data.cpp
```diff
@@ -293,51 +540,33 @@ PerfMetricsDataProvider::PerfMetricsDataProvider(std::unique_ptr<PerfMetricsData
...
+        for (size_t i = 0; i < num_metrics; ++i)
         {
-            PerfMetricsKey key = { cmd_buffer_id, draw_id };
-            const auto     it_sums = sums_and_counts.find(key);
-            if (it_sums == sums_and_counts.end())
+            const auto* metric_info = m_raw_data->GetMetricInfos()[i];
+            if (metric_info && metric_info->m_metric_type == MetricType::kCount)
             {
-                continue;
+                const int64_t average = static_cast<int64_t>(sums[i] / count + 0.5);
```
wangra-google, line 563: std::llround?

dive_core/perf_metrics_data.cpp
```diff
@@ -223,18 +235,269 @@ PerfMetricsData::PerfMetricsData(std::vector<std::string>          metric_names,
...
-    return std::unique_ptr<PerfMetricsDataProvider>(new PerfMetricsDataProvider(std::move(data)));
+    m_pattern_size = template_frame_size;
+    m_record_to_index = std::move(record_to_index);
 }
 
-PerfMetricsDataProvider::PerfMetricsDataProvider(std::unique_ptr<PerfMetricsData> data) :
-    m_raw_data(std::move(data))
+std::unique_ptr<PerfMetricsDataProvider> PerfMetricsDataProvider::Create()
 {
+    return std::unique_ptr<PerfMetricsDataProvider>(new PerfMetricsDataProvider);
```
wangra-google, line 473: std::make_unique?

ui/analyze_window.cpp
```diff
@@ -727,7 +727,7 @@ void AnalyzeDialog::OnReplay()
     }
 
     // Run the perf counter replay
-    if (!m_enabled_settings_vector->empty())
+    if (!m_enabled_settings_vector->empty() || true)
```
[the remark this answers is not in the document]
hysw: Ignore ui/*, it will be removed when it's not draft anymore.

dive_core/perf_metrics_data.cpp
hysw: Private constructor :(

dive_core/perf_metrics_data.cpp
```diff
@@ -223,18 +235,269 @@ PerfMetricsData::PerfMetricsData(std::vector<std::string>          metric_names,
...
+}
+
+void PerfMetricsDataProvider::Correlator::ExtractDraws(
+const CommandHierarchy&                 command_hierarchy,
+std::vector<uint64_t>&                  out_draws,
+std::unordered_map<uint64_t, uint64_t>& out_mapping)
+{
+    std::unordered_map<uint64_t, uint64_t> draw_to_index;
+
+    std::vector<uint64_t> actual_draws;
```
wangra-google, line 312: if i understand correctly, the actual_draws contains all drawcalls in the binning pass, and alias_draws contains all drawcalls in the tiled passes? and this func builds a map that is used to map the drawcalls in the PM4 treeview drawcalls to the perf counter?

Could you add some comments to explain this? it might be a bit tricky to understand the logic.

dive_core/perf_metrics_data.cpp
```diff
@@ -223,18 +235,269 @@ PerfMetricsData::PerfMetricsData(std::vector<std::string>          metric_names,
...
+    m_pattern_size = 0;
+    m_record_to_index.clear();
+}
+
+void PerfMetricsDataProvider::Correlator::AnalyzeCommands(const CommandHierarchy& command_hierarchy)
+{
+    ExtractDraws(command_hierarchy, m_draws, m_draw_to_index);
+}
+
+bool PerfMetricsDataProvider::Correlator::IsSimilar(size_t                   n,
```
wangra-google, line 377: i feel that the name IsSimilar is a bit confusing, maybe something like `AreDrawSignaturesMatching` ?

dive_core/perf_metrics_data.cpp
```diff
@@ -15,15 +15,19 @@
 
 #include <cctype>
 #include <cerrno>
+#include <cstdint>
```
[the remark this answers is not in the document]
hysw: It's added by clangd's include cleaner. Having those headers here should be fine?

dive_core/perf_metrics_data.cpp
hysw: Updated.

dive_core/perf_metrics_data.cpp
hysw: That's basically the idea. If the commands are known, it also being used for validating the number of draw call in a frame. There's more information in the command streams we can use to matching draws better, but current approach should work for a large portion of traces.

wangra-google (dismissed):

dive_core/perf_metrics_data.cpp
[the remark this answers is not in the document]
wangra-google: oh, good to know. thanks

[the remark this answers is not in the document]
wangra-google: it is very interesting since in gaming industry, we prefer PCH so that it is much faster to compile the engine/game code.

PR #402, opened 2025-09-24
Perf counter metrics integration fix.

hysw:
COUNT is a double precision floating point type.
It happens to look like integer for some counter type.

### DIFF
--- a/dive_core/perf_metrics_data.cpp
+++ b/dive_core/perf_metrics_data.cpp
@@ -99,18 +99,6 @@ std::optional<PerfMetricsRecord> ParseRecordFixedFields(const std::vector<std::s
     return record;
 }
 
-template<typename T>
-bool ParseAndEmplaceMetric(const std::string& s, std::vector<std::variant<int64_t, float>>& values)
-{
-    T val{};
-    if (StringUtils::SafeConvertFromString(s, val))
-    {
-        values.emplace_back(val);
-        return true;
-    }
-    return false;
-}
-
 bool ParseMetrics(const std::vector<std::string>&       fields,
                   const std::vector<const MetricInfo*>& metric_infos,
                   PerfMetricsRecord&                    record)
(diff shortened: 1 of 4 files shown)


hysw:
> Can we force Qt to put only one digits after the decimal point without using scientific notation instead relying on the default behavior? My worry is that it would be hard to read otherwise.

It's possible to do that, although that's an UI issue, and we may want to customize per counter.

wangra-google (approved):

GrantComm (dismissed):

dive_core/perf_metrics_data.cpp
```diff
@@ -189,6 +154,24 @@ std::unique_ptr<AvailableMetrics> available_metrics)
...
+            return nullptr;
+        }
+        switch (info->m_metric_type)
+        {
+        case MetricType::kCount:
+        case MetricType::kPercent:
+            break;
+        default:
+            // kUnknown or other types are not supported.
+            return nullptr;
```
wangra-google, line 171: better assert in this case?

dive_core/perf_metrics_data.cpp
```diff
@@ -189,6 +154,24 @@ std::unique_ptr<AvailableMetrics> available_metrics)
     auto& metric_names = headers_opt->metric_names;
     auto& metric_infos = headers_opt->metric_infos;
 
+    for (size_t i = 0; i < metric_infos.size(); ++i)
+    {
+        const MetricInfo* info = metric_infos[i];
+        if (info == nullptr)
+        {
+            return nullptr;
```
wangra-google, line 162: better assert in this case?

dive_core/perf_metrics_data.cpp
hysw: Same as above.

dive_core/perf_metrics_data.cpp
```diff
@@ -490,6 +471,17 @@ std::unique_ptr<PerfMetricsData> data)
     result->Update(std::move(data));
     return result;
 }
+
+std::unique_ptr<PerfMetricsDataProvider> PerfMetricsDataProvider::CreateForTest(
+std::unique_ptr<PerfMetricsData>  data,
+std::unique_ptr<AvailableMetrics> desc)
+{
+    auto result = std::unique_ptr<PerfMetricsDataProvider>(new PerfMetricsDataProvider);
```
wangra-google, line 479: could call `auto result = Create(data);` just in case we may add code in `PerfMetricsDataProvider::Create` in the future.

hysw: Updated.

PR #404, opened 2025-09-24
libwrap: Truncate collected trace file

footballhead:
The current code tracks separate trace files for each FD. This was deemed ["useful because they separate traces from OGLES and Vulkan contexts into different files"](https://gitlab.freedesktop.org/freedreno/freedreno/-/merge_requests/23#note_1856011)

As a result, collect_trace_file() effectively replaces rd_end() in order to concatenate all open trace files into a single trace file; this is easier for the host program to download and analyze.

However, if the target file for concatentation already existed from a previous trace then it would be appended to. This increases file transfer times and ruins analysis. To that end, truncate the collected log file at the start of collect_trace_file().

For simplicity of implementation, this also involved removing the optimization at the start of append_file.

Note that this is only currently a problem for the PM4 capture UI feature (`--command=capture`), not the Analyze UI feature (`--command=profile_replay`). However, this approach should be more robust to a variety of error conditions compared to some simpler alternatives.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,8 +18,6 @@ cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 
 project(dive)
 
-option(DIVE_UI_WIN32 "(Windows-only) Build Dive UI as WIN32 GUI application instead of a console application. This will supress logging to stdout." ON)
-
 # Global properties
 set_property(GLOBAL PROPERTY USE_FOLDERS ON) # Defaults to ON with cmake 3.26+
 
--- a/capture_service/constants.h
+++ b/capture_service/constants.h
@@ -59,4 +59,10 @@ kFirstPort = 49391;  // A port number within the dynamic port range (49152 to 65
 inline constexpr int
 kPortRange = 7;  // A small range of ports should be enough to find an available one
 
+// GPU clock
+inline constexpr uint32_t kPinGpuClockMHz = 545;
+inline constexpr char     kDeviceGpuMinClockPath[] = "/sys/kernel/gpu/gpu_min_clock";
+inline constexpr char     kDeviceGpuMaxClockPath[] = "/sys/kernel/gpu/gpu_max_clock";
+inline constexpr char     kDeviceCurFreqPath[] = "/sys/class/kgsl/kgsl-3d0/devfreq/cur_freq";
+
 }  // namespace Dive
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -257,6 +257,10 @@ absl::Status AndroidDevice::SetupDevice()
 absl::Status AndroidDevice::CleanupDevice()
 {
     LOGD("Cleanup device %s\n", m_serial.c_str());
+
+    UnpinGpuClock().IgnoreError();
+    Adb().Run("shell setprop compositor.high_priority 1").IgnoreError();
+
     if (m_original_state.m_root_access_requested)
     {
         const auto &enforce = m_original_state.m_enforce;
@@ -512,6 +516,25 @@ absl::Status DeviceManager::RunReplayApk(const std::string &capture_path,
 {
     LOGD("RunReplayApk(): starting\n");
 
+    bool trouble_pinning_clock = false;
+    auto ret = m_device->Adb().Run("shell setprop compositor.high_priority 0");
+    if (!ret.ok())
+    {
+        LOGW("WARNING: Could not disable the compositor preemption: %s\n",
+             std::string(ret.message()).c_str());
+        trouble_pinning_clock = true;
+    }
+
+    if (!trouble_pinning_clock)
+    {
+        ret = m_device->PinGpuClock(kPinGpuClockMHz);
+        if (!ret.ok())
+        {
+            LOGW("WARNING: Could not pin GPU clock: %s\n", std::string(ret.message()).c_str());
+            trouble_pinning_clock = true;
+        }
+    }
+
     std::string updated_replay_args = replay_args;
 
     // Enable pm4 capture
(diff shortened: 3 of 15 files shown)


wangra-google (approved):

third_party/freedreno/wrap/wrap-util.c
```diff
@@ -543,6 +537,14 @@ void collect_trace_file(const char* capture_file_path)
 #endif
 	LOGD("full_path is %s", full_path);
 
+	// Could remove(full_path). However, that is susceptible to TOCTOU problems.
+	// Prefer to open the target file once and keep it open for the duration of collection.
+	int concatenated_log_fd = open(full_path, O_CREAT | O_TRUNC | O_WRONLY);
+	if (concatenated_log_fd < 0) {
+		LOGI("Failed to open `%s` to collect traces. Can't collect traces. Error: %s", full_path, strerror(concatenated_log_fd));
```
wangra-google, line 544: concatenated_log_fd would be -1 if the `open` fails? i guess we should use `errno` to get the actual error number?

third_party/freedreno/wrap/wrap-util.c
```diff
@@ -558,14 +560,19 @@ void collect_trace_file(const char* capture_file_path)
...
 				LOGI("Failed to append file %s to target file %s\n", df->file_name, full_path);
 			}
 			// delete the file that has been concatenated. 
 			remove(df->file_name);
         }
 	}
+
+	int close_result = close(concatenated_log_fd);
+	if (close_result < 0) {
+		LOGI("Failed to close concatenated trace file `%s`. There might be some traces missing. Error: %s", full_path, strerror(close_result));
```
wangra-google, line 574: same as above, close_result would be -1 if `close` failed.

third_party/freedreno/wrap/wrap-util.c
```diff
@@ -543,6 +537,14 @@ void collect_trace_file(const char* capture_file_path)
 #endif
```
wangra-google, line 537: not your change, but it is strange to see both path have the same code

footballhead: I agree, fixed in #407

third_party/freedreno/wrap/wrap-util.c
```diff
@@ -558,14 +560,19 @@ void collect_trace_file(const char* capture_file_path)
 			LOG_CLOSE_FILE(df->log_fd);
 			df->log_fd = LOG_NULL_FILE;
 			pthread_mutex_unlock(&write_lock);
-			if(-1 == append_file(full_path, df->file_name))
+			if(-1 == append_file(concatenated_log_fd, df->file_name))
 			{
 				LOGI("Failed to append file %s to target file %s\n", df->file_name, full_path);
 			}
 			// delete the file that has been concatenated. 
 			remove(df->file_name);
```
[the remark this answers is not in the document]
wangra-google: Thanks for detailed explains. interesting. i think for long-term, it is better to keep the .inprogress for debugging purpose (maybe we could even rename the extension to something like .failed) and cleanup later. But for short-term, since we have other higher priority tasks, we could leave it as it is for cleanliness for now.

wangra-google (dismissed):

third_party/freedreno/wrap/wrap-util.c
```diff
@@ -543,6 +537,14 @@ void collect_trace_file(const char* capture_file_path)
 #endif
 	LOGD("full_path is %s", full_path);
 
+	// Could remove(full_path). However, that is susceptible to TOCTOU problems.
+	// Prefer to open the target file once and keep it open for the duration of collection.
+	int concatenated_log_fd = creat(full_path, 0644);
```
footballhead, line 542: For `open`, mode is a vararg and should be set when using O_CREAT. I chose to replace the open call with the equivalent `creat` since the flags are the same but creat checks that `mode` is preset at compile-time.

wangra-google: make sense. thanks

third_party/freedreno/wrap/wrap-util.c
```diff
@@ -528,15 +521,24 @@ int append_file(const char *target, const char *src)
...
+// In the process, all log files are closed and all inprogress files are deleted.
 void collect_trace_file(const char* capture_file_path)
 {
 	LOGD("capture_file_path is %s", capture_file_path);
 
+	// Could remove(full_path). However, that is susceptible to TOCTOU problems.
+	// Prefer to open the target file once and keep it open for the duration of collection.
+	int concatenated_log_fd = creat(full_path, 0644);
+	if (concatenated_log_fd == -1) {
+		LOGI("Failed to open `%s` to collect traces. Can't collect traces. Error: %s", full_path, strerror(errno));
```
footballhead, line 538:
```suggestion
		LOGI("Failed to open `%s` to collect traces. Can't collect traces. Error: %s", capture_file_path, strerror(errno));
```

third_party/freedreno/wrap/wrap-util.c
```diff
@@ -552,12 +554,18 @@ void collect_trace_file(const char* capture_file_path)
...
-		// delete the file that has been concatenated. 
-		remove(df->file_name);
+		else
+		{
+			remove(df->file_name);
+		}
+	}
+
+	if (close(concatenated_log_fd) == -1) {
+		LOGI("Failed to close concatenated trace file `%s`. There might be some traces missing. Error: %s", full_path, strerror(errno));
```
footballhead, line 568:
```suggestion
		LOGI("Failed to close concatenated trace file `%s`. There might be some traces missing. Error: %s", capture_file_path, strerror(errno));
```

wangra-google (approved):

PR #406, opened 2025-09-24
libwrap: Fix capture filename synchronization

footballhead:
libwrap was ignoring the values that AndroidTraceManager was setting. This meant that duration-based PM4 captures wouldn't work correctly. Additionally, it made modifying the code difficult. E.g. I want to use debug.dive.replay.capture_pm4_file_name in more contexts for testing purposes but the current implementation made that frustratingly difficult.

With this change, libwrap becomes the source of truth for the capture file name when PM4 capture (--command=capture) is used. This does not affect the Analyze (--command=profile_replay) flow since DivePM4Capture is used instead.

### DIFF
--- a/capture_service/CMakeLists.txt
+++ b/capture_service/CMakeLists.txt
@@ -48,7 +48,8 @@ target_include_directories(service PRIVATE
 target_link_libraries(service PRIVATE
                       network
                       absl::flags
-                      absl::flags_parse)
+                      absl::flags_parse
+                      absl::strings)
 set_property(TARGET service PROPERTY POSITION_INDEPENDENT_CODE 1)
 
 add_executable(server server.cc
--- a/capture_service/android_trace_mgr.cc
+++ b/capture_service/android_trace_mgr.cc
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#include "absl/strings/str_cat.h"
+#include "absl/strings/str_format.h"
 #include "trace_mgr.h"
 
 #include <string>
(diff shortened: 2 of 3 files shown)


wangra-google (approved):

elviscapiaq (approved):

footballhead:
There's still a bug here but I think I went too far with the assumptions that I was making

1. I thought that the full path in given `TESTNAME` was redundant. It's actually mandatory. This is the code that picks the file:

```c
// rd_start
// `name` in this context is likely to be the value of TESTNAME env var
	testnum = getenv("TESTNUM");
	if (testnum) {
		n = strtol(testnum, NULL, 0);
		sprintf(buf, "%s-%04u.rd.inprogress", name, n);
	} else {
		if (device_fd == -1) {
			if (df->open_count == 0)
				sprintf(buf, "%s/%s.rd.inprogress", base_path, name);
			else
				sprintf(buf, "%s/%s-%d.rd.inprogress", base_path, name, df->open_count);
		}
		else {
			if (df->open_count == 0)
				sprintf(buf, "%s/%s-fd%d.rd.inprogress", base_path, name, device_fd);
			else
				sprintf(buf, "%s/%s-fd%d-%d.rd.inprogress", base_path, name, device_fd, df->open_count);
		}
	}
```

Note that the TESTNUM branch, unlike every other branch, does not include `base_path` (/sdcard/Download) in the filename. Since we always set TESTNUM, the other branches are never taken.

2. I couldn't force libwrap to use a specific filename. This is not true: I could force libwrap to use a specific file name for testing with:

```sh
adb shell setprop debug.dive.replay.capture_pm4 1
adb shell setprop debug.dive.replay.capture_pm4_file_name hitchens.rd
# libwrap will create /sdcard/Download/hitchens.rd
```

This, however, is not communicated to AndroidTraceManager. So PM4 capture can't actually download the file.

I'm going to close this and take another stab at it later now that I have a better idea of the nuances.

PR #408, opened 2025-09-25
Fix parsing of resolve event

wangra-google:
- the 1st 2 bits of RB_BLIT_INFO has been updated in main mesa (https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/freedreno/registers/adreno/a6xx.xml?ref_type=heads#L1763), but it is a massive change to integrate (which will be done in a separate PR), this change interpret those bits based on the updated a6xx xml file
- also update the trace_stats tool to have more detailed sync events

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,8 +18,6 @@ cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 
 project(dive)
 
-option(DIVE_UI_WIN32 "(Windows-only) Build Dive UI as WIN32 GUI application instead of a console application. This will supress logging to stdout." ON)
-
 # Global properties
 set_property(GLOBAL PROPERTY USE_FOLDERS ON) # Defaults to ON with cmake 3.26+
 
@@ -112,7 +110,7 @@ if(NOT ANDROID)
     add_subdirectory(plugins)
 else()
     if(NOT CMAKE_HOST_WIN32)
-        set(DIVE_GFXR_GRADLE_CONSOLE "verbose" CACHE LIST "Value of --console to use when building GFXR with Gradle")
+        set(DIVE_GFXR_GRADLE_CONSOLE "verbose" CACHE STRING "Value of --console to use when building GFXR with Gradle")
         message(CHECK_START "Build gfxr for android")
         execute_process(
             COMMAND ./gradlew layer:assemble${CMAKE_BUILD_TYPE} replay:assemble${CMAKE_BUILD_TYPE} --console=${DIVE_GFXR_GRADLE_CONSOLE} -P${ANDROID_ABI}
--- a/dive_core/perf_metrics_data.cpp
+++ b/dive_core/perf_metrics_data.cpp
@@ -99,18 +99,6 @@ std::optional<PerfMetricsRecord> ParseRecordFixedFields(const std::vector<std::s
     return record;
 }
 
-template<typename T>
-bool ParseAndEmplaceMetric(const std::string& s, std::vector<std::variant<int64_t, float>>& values)
-{
-    T val{};
-    if (StringUtils::SafeConvertFromString(s, val))
-    {
-        values.emplace_back(val);
-        return true;
-    }
-    return false;
-}
-
 bool ParseMetrics(const std::vector<std::string>&       fields,
                   const std::vector<const MetricInfo*>& metric_infos,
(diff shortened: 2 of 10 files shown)


trace_stats/main.cpp
```diff
@@ -152,54 +155,63 @@ void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &o
...
-        "\tMin GPRs in a single shader",
-        "\tMax GPRs in a single shader",
-        "\tMedian GPRs in a single shader",
-        "Total resolves",
-        "\tSysMem to Gmem Resolves",
-        "\tGmem to SysMem Resolves",
-        "\tGmem Clears",
+
+    constexpr std::array kStatMap = {
+        std::pair(kBinningDraws, "Num Draws (BINNING)"),
```
footballhead, line 166: Usually I'd ask for consideration of a struct instead of a pair since it's more self documenting. However, the context is limited enough that I can glean what's going on without difficulty.

trace_stats/main.cpp
```diff
@@ -152,54 +155,63 @@ void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &o
...
+
+    static_assert(kStatMap.size() == kNumStats,
+                  "ERROR: The 'Stat' enum and the 'kStatMap' descriptions are out of sync!");
+
+    constexpr std::array<const char *, kNumStats> kStatDescriptions = [&] {
+        std::array<const char *, kNumStats> arr{};
+        for (const auto &pair : kStatMap)
+        {
+            arr[pair.first] = pair.second;
+        }
```
footballhead, line 212: prefer structured bindings here to give more descriptive names (rather than `first` and `second`
```suggestion
        for (const auto &[stat, description] : kStatMap)
        {
            arr[stat] = description;
        }
```

trace_stats/main.cpp
```diff
@@ -120,13 +121,15 @@ void GatherAndPrintStats(const Dive::CaptureMetadata &meta_data, std::ostream &o
     // Number of CpWaitForIdle()
     // Number of prefetches
     // Number of loads/stores to GMEM
-    enum Stats
+    enum Stats : uint32_t
```
footballhead, line 124: consider enum classes since they're scoped (so don't pollute the current scope).
```suggestion
    enum class Stats : uint32_t
```

dive_core/data_core.cpp
```diff
@@ -369,15 +369,39 @@ bool CaptureMetadataCreator::OnPacket(const IMemoryManager &mem_manager,
                                                    type7_header->opcode,
                                                    m_state_tracker);
 
-            DIVE_ASSERT(sync_type != SyncType::kNone);  // Sanity check
-            if (sync_type == SyncType::kSysMemToGmemResolve)
-                event_info.m_type = EventInfo::EventType::kSysmemToGmemResolve;
-            else if (sync_type == SyncType::kGmemToSysMemResolve)
+            switch (sync_type)
```
footballhead, line 372: thanks for using a switch 👍

dive_core/data_core.cpp
```diff
@@ -369,15 +369,39 @@ bool CaptureMetadataCreator::OnPacket(const IMemoryManager &mem_manager,
...
+                event_info.m_type = EventInfo::EventType::kWaitForMe;
+                break;
+            case SyncType::kEventWriteStart:
+                event_info.m_type = EventInfo::EventType::kEventWriteStart;
+                break;
+            case SyncType::kEventWriteEnd:
+                event_info.m_type = EventInfo::EventType::kEventWriteEnd;
+                break;
+            default:
+                DIVE_ASSERT(false);  // Sanity check
```
footballhead, line 402: Prefer to use more inclusive language. https://www.khronos.org/about/inclusive-language#sanity

```suggestion
                DIVE_ASSERT(false);  // Unexpected SyncType could cause problems later
```

trace_stats/main.cpp
wangra-google: this is a bit painful to add `class` in this case. with this change, every where the enum is used as uint32_t needs to be explicitly cased with static_cast<> which makes the code harder to read (there are quite a lot in this file).

wangra-google: actually i found a better solution by using a struct instead, this way i could add the scope without forcing the static_cast

shanminchao (approved):

PR #409, opened 2025-09-25
Pin Android device's GPU clock frequency during GFXR replay

angela28chen:
- If the clock pinning does not work as expected, print some warnings but do not halt execution.
- If the clock pinning works as expected, do not give additional logs (though this can still be noticed from the logs that indicate which commands on the device were run)

### DIFF
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -259,6 +259,7 @@ absl::Status AndroidDevice::CleanupDevice()
     LOGD("Cleanup device %s\n", m_serial.c_str());
 
     UnpinGpuClock().IgnoreError();
+    Adb().Run("shell setprop compositor.high_priority 1").IgnoreError();
 
     if (m_original_state.m_root_access_requested)
     {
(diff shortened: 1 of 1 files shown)


angela28chen:
Looking at if there's a way to improve the logging situation

capture_service/device_mgr.cc
```diff
@@ -512,6 +515,14 @@ absl::Status DeviceManager::RunReplayApk(const std::string &capture_path,
 {
     LOGD("RunReplayApk(): starting\n");
 
+    bool pinned_clock = true;
```
wangra-google, line 518: i am wondering if you could create a helper class and put this code block to the constructor, and the next code block to the destructor so that we can avoid duplicating the code in both RunReplayApk and RunProfilingOnReplay

wangra-google: sounds good

wangra-google (approved):

elviscapiaq (approved):

PR #412, opened 2025-09-29
Moves trace stats collection and printing

GrantComm:
Moves trace stats collection and printing to dive_core so that the statistics can also be displayed in ui. A follow up pr will add that display. Attached is a screenshot of the statistics used on a cubexr capture.

Screenshot 2025-09-29 at 11 23 46 PM


dive_core/data_core.h
```diff
@@ -30,6 +30,73 @@
 namespace Dive
 {
 
+#define CHECK_AND_TRACK_STATE_1(stats_enum, state)                   \
```
wangra-google, line 33: i feel that it would be clearer if we separate these changes into separate trace_stats.h/cpp. data_core.h/cpp becomes too large with those changes.

trace_stats/trace_stats.h
```diff
@@ -0,0 +1,172 @@
...
+    void GatherTraceStats(const Dive::CaptureMetadata &meta_data);
+
+    // Print the capture statistics to the output stream
+    void PrintTraceStats(std::ostream &ostream);
+
+    // Get the capture statistics
+    const CaptureStats &GetCaptureStats() const { return m_capture_stats; };
+
+private:
+    CaptureStats m_capture_stats;
```
shanminchao, line 169: **Discuss**: Is it better to pass in a CaptureStats as a parameter to GatherTraceStats()?
i.e. void GatherTraceStats(const Dive::CaptureMetadata &meta_data, CaptureStats *capture_stats);

That way TraceStats as a class doesn't hold any state. Arguably cleaner.

trace_stats/trace_stats.h
```diff
@@ -0,0 +1,172 @@
...
+    }
+};
+
+// ---------------------------------------------------------------------
+// Statistics Container
+// ---------------------------------------------------------------------
+
+struct CaptureStats
+{
+    std::array<uint64_t, Dive::Stats::kNumStats> stats_list = {};
```
shanminchao, line 141: nit: To be consistent with other naming conventions, perhaps we should add a m_ prefix to member variables of structs and classes

trace_stats/trace_stats.cpp
```diff
@@ -0,0 +1,238 @@
...
+    for (uint32_t i = 0; i < Dive::Stats::kNumStats; ++i)
+    {
+        ostream << kStatDescriptions[i] << ": " << stats_list[i] << "\n";
+    }
+
+    ostream << viewport_stats_desc[kViewport] << ":\n";
+
+    for (const Viewport &vp : m_capture_stats.viewports)
+    {
+        ostream << viewport_stats_desc[kViewport_x] << ": " << (int)vp.m_vk_viewport.x << ", "
```
shanminchao, line 213: Add a "\t" to indent each line? The original had an indentation.

trace_stats/trace_stats.cpp
```diff
@@ -0,0 +1,238 @@
...
+                << ": " << std::fixed << std::setprecision(1) << vp.m_vk_viewport.minDepth << ", "
+                << viewport_stats_desc[kViewport_maxDepth] << ": " << std::fixed
+                << std::setprecision(1) << vp.m_vk_viewport.maxDepth << "\n";
+    }
+
+    ostream << window_scissor_stats_desc[kWindowScissors] << ":\n";
+
+    for (const WindowScissor &ws : m_capture_stats.window_scissors)
+    {
+        ostream << window_scissor_stats_desc[kWindowScissors_tl_x] << ": " << ws.m_tl_x << ", "
```
shanminchao, line 227: Add a "\t" to indent each line? The original had an indentation.

shanminchao: I don't see an updated screenshot in the PR description? Still doesn't have the indent.

trace_stats/trace_stats.h
shanminchao: Thanks!

trace_stats/trace_stats.cpp
shanminchao: I don't see an updated screenshot in the PR description? Still doesn't have the indent.

dive_core/common/gpudefs.h
```diff
@@ -80,4 +81,143 @@ inline uint16_t DecodeSGPRs(uint32_t pgm_rsrc1)
     return sgprs;
 }
 
+//--------------------------------------------------------------------------------------------------
```
wangra-google, line 84: is there a reason you prefer to put this in gpudefs.h? is it better to put it in trace_stats.h?

trace_stats/trace_stats.cpp
GrantComm: Ah, uploaded the wrong one. It is updated now.

shanminchao (approved):

PR #413, opened 2025-09-29
Adds Trace stats display to overview tab

GrantComm:
Adds trace stats display to the overview tab. This display is populated when pm4 capture data is loaded (only pm4 command hierarchy or mixed command hierarchy). 

Users can also click a button to copy the output of the trace stats tool and paste to another text file if desired.

copy_button
overview_tab


GrantComm:
> > Adds trace stats display to the overview tab. This display is populated when pm4 capture data is loaded (only pm4 command hierarchy or mixed command hierarchy)
> > pipeline_stats_scrsht mapping_stats_scrnsht
> 
> looks nice! I like you split the stats into multiple tabs. some comments:
> 
> * you can remove or hide the existing unused tabs (Problems & Most Expensive Events), my worry is that leaving them there without any info may confuse people
> * better to rename `pipeline stats` to `Draw/Dispatch Stats` and put all draw/dispatch stats in this tab (including shader related stats)
> * better to rename `Output mapping stats` to `Tile Stats`, and put anything related to Tiling/Binning here (including Viewports, scissors)
> * you can have another tab `Misc stats` and put anything that does not belong to any of the above 2 to here. (Num WaitMemWrite, Num WaitForIdle, Num WaitForMe, Total resolves)

Ok, will update this.

GrantComm:
> > > > Adds trace stats display to the overview tab. This display is populated when pm4 capture data is loaded (only pm4 command hierarchy or mixed command hierarchy)
> > > > pipeline_stats_scrsht mapping_stats_scrnsht
> > > 
> > > 
> > > looks nice! I like you split the stats into multiple tabs. some comments:
> > > 
> > > * you can remove or hide the existing unused tabs (Problems & Most Expensive Events), my worry is that leaving them there without any info may confuse people
> > > * better to rename `pipeline stats` to `Draw/Dispatch Stats` and put all draw/dispatch stats in this tab (including shader related stats)
> > > * better to rename `Output mapping stats` to `Tile Stats`, and put anything related to Tiling/Binning here (including Viewports, scissors)
> > > * you can have another tab `Misc stats` and put anything that does not belong to any of the above 2 to here. (Num WaitMemWrite, Num WaitForIdle, Num WaitForMe, Total resolves)
> > 
> > 
> > Ok, will update this.
> 
> Oh, one more comment on this: I really like this UI, but please keep in mind that there is a following task to save the trace_stats result into a txt so that it would be easier to compare across captures. do you have any thoughts on where we could put a `save` button?

Do we need to have a save button? Could the trace_stats txt file just be saved to the folder if a .rd file is loaded and then subsequent opens would just check to see if that file exists

wangra-google:
> > > > > Adds trace stats display to the overview tab. This display is populated when pm4 capture data is loaded (only pm4 command hierarchy or mixed command hierarchy)
> > > > > pipeline_stats_scrsht mapping_stats_scrnsht
> > > > 
> > > > 
> > > > looks nice! I like you split the stats into multiple tabs. some comments:
> > > > 
> > > > * you can remove or hide the existing unused tabs (Problems & Most Expensive Events), my worry is that leaving them there without any info may confuse people
> > > > * better to rename `pipeline stats` to `Draw/Dispatch Stats` and put all draw/dispatch stats in this tab (including shader related stats)
> > > > * better to rename `Output mapping stats` to `Tile Stats`, and put anything related to Tiling/Binning here (including Viewports, scissors)
> > > > * you can have another tab `Misc stats` and put anything that does not belong to any of the above 2 to here. (Num WaitMemWrite, Num WaitForIdle, Num WaitForMe, Total resolves)
> > > 
> > > 
> > > Ok, will update this.
> > 
> > 
> > Oh, one more comment on this: I really like this UI, but please keep in mind that there is a following task to save the trace_stats result into a txt so that it would be easier to compare across captures. do you have any thoughts on where we could put a `save` button?
> 
> Do we need to have a save button? Could the trace_stats txt file just be saved to the folder if a .rd file is loaded and then subsequent opens would just check to see if that file exists

good point, the only concern i have is that eventually we would merge all files into 1 .dive file, introduce a separate file might be confusing? or maybe that could be part of the .dive file?

PR #423, opened 2025-10-01
Adds vulkan command tooltip summaries

GrantComm:
Adds vulkan command tooltip summaries that are displayed on hover of vulkan commands. These summaries were Gemini generated and are retrieved using the vulkan command name.

tooltip

### DIFF
--- a/ui/tool_tip_summaries.cpp
+++ b/ui/tool_tip_summaries.cpp
@@ -19,7 +19,9 @@
 
 const std::unordered_map<std::string, const char*>& GetVulkanCommandToolTipSummaries()
 {
-    // Brief descriptions of Vulkan API commands used for tooltips in the UI.
+    // Brief descriptions of Vulkan API commands used for tooltips in the UI. These current as of
+    // 10/6/2025 and are collected from https://github.com/KhronosGroup/Vulkan-Docs.git (Commit
+    // 7fe2b62)
     static const std::unordered_map<std::string, const char*> vulkan_command_tool_tip_summaries = {
         { "vkAcquireDrmDisplayEXT", "Acquire access to a VkDisplayKHR using DRM" },
         { "vkAcquireFullScreenExclusiveModeEXT",


angela28chen (changes requested):
In the interest of transparency I think that we should make it clearer that these tooltips are AI-generated summaries. I have some ideas on how we can do this:

- This PR title should include "gemini/genAI"
- We should add disclosure info in the top level README.md of Dive project, including information such as which tool was used and the date of when the results were generated.
- We should add an icon in the tooltip that hints to the user that the accompanying text is AI-generated, such as the commonly used diamond icon for Gemini

In addition to the above, has anyone with Vulkan knowledge read through these tooltips and made sure they are coherent? @wangra-google If you could confirm that would be great. Thank you.

Alternative ideas to consider:
- Adding a link to the official vulkan documentation for more detail
```
https://registry.khronos.org/vulkan/specs/latest/man/html/ + <vulkan command name> + .html
```
- Obtaining summaries directly from scraping the official documentation (these seem to be pretty short and potentially less descriptive, but more reliable)
- Add "copy to clipboard" feature so users can search up the Vulkan command name easily

wangra-google:
> In the interest of transparency I think that we should make it clearer that these tooltips are AI-generated summaries. I have some ideas on how we can do this:
> 
> * This PR title should include "gemini/genAI"
> * We should add disclosure info in the top level README.md of Dive project, including information such as which tool was used and the date of when the results were generated.
> * We should add an icon in the tooltip that hints to the user that the accompanying text is AI-generated, such as the commonly used diamond icon for Gemini
> 
> In addition to the above, has anyone with Vulkan knowledge read through these tooltips and made sure they are coherent? @wangra-google If you could confirm that would be great. Thank you.
> 
> Alternative ideas to consider:
> 
> * Adding a link to the official vulkan documentation for more detail
> 
> ```
> https://registry.khronos.org/vulkan/specs/latest/man/html/ + <vulkan command name> + .html
> ```
> 
> * Obtaining summaries directly from scraping the official documentation (these seem to be pretty short and potentially less descriptive, but more reliable)
> * Add "copy to clipboard" feature so users can search up the Vulkan command name easily

I have checked the AI generated content, they are correct. 

I am not sure about if we need to mention this as AI generated in either readme or UI. For me, this is more like to use AI to pre-generate descriptions which is different from using AI to analyze based on dynamic content. And the users might get confused by seeing the gemini icon to think that we are using gemini to dynamically generate content which is not the case here. Here is just to give a general idea of what each function does, people who use the gpu profiler would already have some knowledge of each function, having a link for each function would introduce too much text in the tool tip, especially considering they are not clickable.

hysw:
fyi, you can regex through https://github.com/KhronosGroup/Vulkan-Docs (or ask gemini to do it for you)

e.g. https://gist.github.com/hysw/23fa3efc469c3858937d432b5bd4b0ef

ui/tool_tip_summaries.h
```diff
@@ -0,0 +1,467 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+#include <unordered_map>
+#include <string>
+
+// Generated summaries for Vulkan API commands, used for tooltips in the UI.
+const std::unordered_map<std::string, const char*> kVulkanCommandTooltipSummaries = {
```
hysw, line 21: Please don't instantiate variable in header file.
Also global non-constexpr variables are bad idea.

See https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables or go/totw/110

file.h
```
const std::unordered_map<std::string_view, std::string_view>& GetVulkanCommandTooltipSummaries();
```
file.cpp
```
const std::unordered_map<std::string_view, std::string_view>& GetVulkanCommandTooltipSummaries() {
  static const auto* impl  =
    new std::unordered_map<std::string_view, std::string_view>{
    // ...
  };
  return *impl;
}
```

hysw:
> Yes, I have reviewed the AI generated content, they are accurate.

I'd argue otherwise, there's quite a bit confabulation going on, e.g.

```
    { "vkQueueSubmit2KHR",
      "KHR extension for V1.3 submission: Submits command buffers with extended synchronization "
      "info." },
```

It's part of VK_KHR_synchronization2 for Vulkan 1.1, or Vulkan 1.3 core

wangra-google (dismissed):

ui/tool_tip_summaries.h
```diff
@@ -0,0 +1,20 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+#include <unordered_map>
+#include <string>
+
+const std::unordered_map<std::string, const char*>& GetVulkanCommandToolTipSummaries();
```
hysw, line 20: nit: please add a newline at the end of file.

GrantComm:
> Did you change it to use the Vulkan Spec description?

Yes, I updated it to use the descriptions in https://gist.github.com/hysw/23fa3efc469c3858937d432b5bd4b0ef

angela28chen:
> Thanks for these thoughtful suggestions.
> 
> Yes, I have reviewed the AI generated content, they are accurate.
> 
> My perspective is that this is more like to use AI to pre-generate descriptions (which is static, and has been human-reviewed, it functions more like standard documentation.) which is different from using AI to analyze based on dynamic content. And the users might get confused by seeing the gemini icon to think that we are using gemini to dynamically generate content which is not the case here. Here is just to give a general idea of what each function does, people who use the gpu profiler would already have some knowledge of each function, having a link for each function would introduce too much text in the tool tip, especially considering they are not clickable.
> 
> Considering our team's current priorities, my preference would be to keep the tooltips simple (also not mention the genAI in either UI or Readme for now). Perhaps we can add these other suggestions to our backlog as potential enhancements for a future iteration.

First of all, I can see that Grant has changed the tooltips from the first review and I'm not sure the source of it right now (@GrantComm is it the tooltips Peter provided?). My following comments will be about the tooltips produced by Gemini in the first PR review.

I think as Peter has pointed out & as I've seen myself, the content is not necessarily reliable in quality. I have done a brief look over of it and I can easily spot some minor issues. A handful of examples:
- Some repeated: vkCmdBindTileMemoryQCOM, vkCmdBindVertexBuffers2, vkCmdBindVertexBuffers2EXT, vkCmdDrawClusterHUAWEI...
- Some misrepresention of if it's available in Vulkan 1.3 core vs older Vulkan with extensions
- Misleading descriptions: "vkCmdBindDescriptorSets: Binds descriptor sets (containing resource links) to the current pipeline layout."
- etc.

Vulkan is a complicated spec that's hard for newcomers (the people most likely to rely on the tooltips) to learn. It's better to not provide information than to provide wrong information authoritatively. I think this is an issue that can damage the public perception of the reliability of a graphics tool. 

It's not a good use of our time to manually verify all the Gemini output: We should just rely on scraping freely available Vulkan docs if we want a brief description that accurately represents the Vulkan commands. 

But if we do go forward with using generative AI produced tooltips then we should make it clear what it is, especially so that users can understand where the information is from and treat it with the appropriate level of trust. Rather than omitting that info, I think we can take other routes to make sure users understand it's pre-generated and not freshly generated each time. It can be as simple as "(pre-generated by Gemini ⟡)" prefix to the tooltips.

hysw:
> > > Did you change it to use the Vulkan Spec description?
> > 
> > 
> > Yes, I updated it to use the descriptions in https://gist.github.com/hysw/23fa3efc469c3858937d432b5bd4b0ef
> 
> I'm not sure where Peter got these descriptions but if they're just scraped from the Vulkan documentation and not involved with Gemini then I think the PR title and description should change to reflect that?

Get Vulkan doc `git clone https://github.com/KhronosGroup/Vulkan-Docs.git`

Then use the following script
```
import os
import re
import sys

def extract_info_from_adocs_recursive(directory="."):
    """
    Recursively searches all .adoc files in the given directory and its
    subdirectories for lines matching a specific pattern and extracts
    'refpage' and 'desc' attributes.

    The pattern a line must match is:
    [open,refpage='<VALUE>',desc='<VALUE>',type='protos']

    Args:v
        directory (str): The root directory to start the scan. Defaults to the
                         current directory.
    """
    # Regular expression to capture 'refpage' and 'desc' values
    pattern = re.compile(r".*?refpage='(.*?)',.*?desc='(.*?)',.*")
    pattern_alias = re.compile(r".*?alias='(.*?)'.*")

    found_any = False
    exists = set()
    entries = []
    try:
        for root, dirs, files in os.walk(directory):
            for filename in files:
                if filename.endswith(".adoc"):
                    filepath = os.path.join(root, filename)
                    try:
                        with open(filepath, 'r', encoding='utf-8') as file:
                            for line_num, line in enumerate(file, 1):
                                match = pattern.search(line)
                                if match:
                                    refpage = match.group(1)
                                    if refpage in exists:
                                        continue
                                    exists.add(refpage)
                                    desc = match.group(2)
                                    if desc.startswith("Stub description of "):
                                        continue
                                    entries.append((refpage, desc))

                                    alias_match = pattern_alias.search(line)
                                    if alias_match:
                                      alias = alias_match.group(1)
                                      assert "," not in alias
                                      if alias in exists:
                                          continue
                                      exists.add(alias)
                                      entries.append((alias, desc))
                                    found_any = True
                    except Exception as e:
                        print(f"Error reading file {filepath}: {e}", file=sys.stderr)
    except Exception as e:
        print(f"Error walking through directory {directory}: {e}", file=sys.stderr)
    for refpage, desc in sorted(entries):
        if not refpage.startswith("vk"):
            continue
        print(f'{{ "{refpage}": "{desc}" }},')

if __name__ == "__main__":
    extract_info_from_adocs_recursive()
```

Note: I did ask Gemini to give a boilerplate recursive regex scan script, but the generated descriptions are from Vulkan Spec.

ui/tool_tip_summaries.cpp
```diff
@@ -0,0 +1,1052 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+#include "tool_tip_summaries.h"
+#include <unordered_map>
+
+const std::unordered_map<std::string, const char*>& GetVulkanCommandToolTipSummaries()
+{
+    // Generated summaries for Vulkan API commands, used for tooltips in the UI.
```
hysw, line 22: nits: this line should be updated.

wangra-google (dismissed):

hysw (dismissed):

ui/tool_tip_summaries.cpp
```diff
@@ -0,0 +1,1052 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+#include "tool_tip_summaries.h"
+#include <unordered_map>
+
+const std::unordered_map<std::string, const char*>& GetVulkanCommandToolTipSummaries()
+{
+    // Brief descriptions of Vulkan API commands used for tooltips in the UI.
```
angela28chen, line 22: In a comment, add some basic info here about where the tooltips are from (https://github.com/KhronosGroup/Vulkan-Docs.git or website link for easy reference) and when they were collected (the date or commit). Useful context particularly if we need to update these tooltips in the future.

angela28chen (changes requested):
LGTM, just added a comment because I think we should still be saying where the tooltips are sourced from

ui/tool_tip_summaries.cpp
angela28chen: Thanks!

angela28chen (approved):

hysw (approved):

PR #427, opened 2025-10-02
State tracker fix

shanminchao:
Fix bug with IsRegSet/GetRegValue

Previous implementation uses an enable_mask and grabs the value of the register for the first enabled mask (i.e. BINNING, SYSMEM, or GMEM). This is not correct. Modified instead to grab the register value of the current ShaderEnableBit, as determined by the CP_SET_MARKER packet

### DIFF
--- a/dive_core/common/emulate_pm4.cpp
+++ b/dive_core/common/emulate_pm4.cpp
@@ -48,6 +48,55 @@ bool EmulateStateTracker::OnPacket(const IMemoryManager &mem_manager,
                                    uint64_t              va_addr,
                                    Pm4Header             header)
 {
+    if (header.type == 7 && header.type7.opcode == CP_SET_MARKER)
+    {
+        PM4_CP_SET_MARKER packet;
+        DIVE_VERIFY(mem_manager.RetrieveMemoryData(&packet, submit_index, va_addr, sizeof(packet)));
+        // as mentioned in adreno_pm4.xml, only b0-b3 are considered when b8 is not set
+        DIVE_ASSERT((packet.u32All0 & 0x100) == 0);
+        a6xx_marker marker = static_cast<a6xx_marker>(packet.u32All0 & 0xf);
+        switch (marker)
+        {
+            // This is emitted at the beginning of the render pass if tiled rendering mode is
+            // disabled
+        case RM6_BYPASS:
+            m_shader_enable_bit = ShaderEnableBit::kSYSMEM;
+            break;
+            // This is emitted at the beginning of the binning pass, although the binning pass
+            // could be missing even in tiled rendering mode
+        case RM6_BINNING:
+            m_shader_enable_bit = ShaderEnableBit::kBINNING;
+            break;
+            // This is emitted at the beginning of the tiled rendering pass
+        case RM6_GMEM:
+            m_shader_enable_bit = ShaderEnableBit::kGMEM;
+            break;
+            // This is emitted at the end of the tiled rendering pass
+        case RM6_ENDVIS:
+            // should be paired with RM6_BIN_RENDER_START only if RM6_BIN_VISIBILITY exist, end of
+            // tiled mode
+            m_shader_enable_bit = std::nullopt;
+            break;
+            // This is emitted at the beginning of the resolve pass
+        case RM6_RESOLVE:
+            m_shader_enable_bit = ShaderEnableBit::kGMEM;
+            break;
+            // This is emitted for each dispatch
+        case RM6_COMPUTE:
+            break;
+            // This seems to be the end of Resolve Pass
+        case RM6_YIELD:
+            // should be paired with RM6_BIN_RESOLVE, end of resolve pass
+            m_shader_enable_bit = std::nullopt;
+            break;
+        case RM6_BLIT2DSCALE:
+        case RM6_IB1LIST_START:
+        case RM6_IB1LIST_END:
+        default:
+            break;
+        }
+    }
+
     if (header.type == 7 && header.type7.opcode == CP_CONTEXT_REG_BUNCH)
     {
         uint32_t dword = 0;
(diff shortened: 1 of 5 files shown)


dive_core/common/emulate_pm4.cpp
```diff
@@ -231,41 +243,32 @@ void EmulateStateTracker::PopEnableMask()
...
 {
-    for (unsigned int i = 0; i < kShaderEnableBitCount; ++i)
-    {
-        if (enable_mask & (1u << i))
-        {
-            return m_reg[i][offset];
-        }
-    }
-    return 0;
+    uint32_t i = (uint32_t)shader_enable_bit;
```
wangra-google, line 248: static_cast

dive_core/common/emulate_pm4.cpp
```diff
@@ -231,41 +243,32 @@ void EmulateStateTracker::PopEnableMask()
...
-    for (unsigned int i = 0; i < kShaderEnableBitCount; ++i)
-    {
-        if (enable_mask & (1u << i))
-        {
-            return ((uint64_t)m_reg[i][offset]) | (((uint64_t)m_reg[i][offset + 1]) << 32);
-        }
-    }
-    return 0;
+    uint32_t i = (uint32_t)shader_enable_bit;
+    return ((uint64_t)m_reg[i][offset]) | (((uint64_t)m_reg[i][offset + 1]) << 32);
```
wangra-google, line 264: static_cast

dive_core/common/emulate_pm4.cpp
```diff
@@ -284,38 +287,16 @@ void EmulateStateTracker::SetReg(uint32_t offset, uint32_t value)
...
 {
-    for (unsigned int i = 0; i < kShaderEnableBitCount; ++i)
-    {
-        if (enable_mask & (1u << i))
-        {
-            return !!(m_reg_is_set[i][offset / 8] & (1 << (offset % 8)));
-        }
-    }
-    return false;
+    return !!(m_reg_is_set[(uint32_t)shader_enable_bit][offset / 8] & (1 << (offset % 8)));
```
wangra-google, line 297: static_cast

wangra-google: also maybe better use static_cast<bool> for !!

ui/shader_view.cpp
```diff
@@ -215,7 +215,7 @@ void ShaderView::paintEvent(QPaintEvent *event)
                     {
                         std::ostringstream ostr;
                         bool               empty = true;
-                        if (reference.m_enable_mask & (uint32_t)Dive::ShaderEnableBit::kBINNING)
+                        if (reference.m_enable_mask & (uint32_t)Dive::ShaderEnableBitMask::kBINNING)
```
wangra-google, line 218: static_cast

shanminchao:
> > One example symptom of this bug:
> > On a particular trace (trace-frame-5381.rd), the event state window claims blend state was never set, when in fact it was set.
> 
> sorry still a bit confused, why is the affecting the color blending state?

When the register was first set, it was set with its enable_mask set to 6 (i.e. BINNING & GMEM).
But when we end up calling IsRegSet() when we encounter a draw call, the enable_mask then is set to 7.

So the old code checked bit by bit of the enable_mask. In this case it checks the first bit of the enable_mask during time of draw (I.e. 7), see that it's set, then checks the m_reg_is_set[0], which was never set to true because the enable_mask was 6 when that register was set. Note that bit 0 is referring to SYSMEM.

dive_core/common/emulate_pm4.cpp
shanminchao: Not sure I understand. You mean static_cast to bool instead of using the double exclamation?

wangra-google: Yes

dive_core/common/emulate_pm4.h
```diff
@@ -85,9 +85,16 @@ union Pm4Header
 //--------------------------------------------------------------------------------------------------
 enum class ShaderEnableBit : uint32_t
```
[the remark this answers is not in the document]
shanminchao: I like your trick and I tried using it. But realized there's some limitations to this method.
For example, this isn't allowed:

ShaderEnableBit shader_enable_bit;
(uint32_t)shader_enable_bit

or this:

ShaderEnableBit m_shader_enable_bit = ShaderEnableBit::kBINNING;


In both cases, you need to use the ShaderEnableBit::Type type, and not ShaderEnableBit directly, which sucks.
But the static cast sucks too. Hmmm...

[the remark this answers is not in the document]
shanminchao: ... and it looks like even in this case:

            ShaderEnableBit::Type shader_enable_bit = enable_index;

This is still invalid without a static_cast. Since the whole point was to avoid static_cast, I guess this method isn't working too well. So I guess I'll just opt for doing static_cast everywhere instead. Sigh...

shanminchao:
> > > > > > One example symptom of this bug:
> > > > > > On a particular trace (trace-frame-5381.rd), the event state window claims blend state was never set, when in fact it was set.
> > > > > 
> > > > > 
> > > > > sorry still a bit confused, why is the affecting the color blending state?
> > > > 
> > > > 
> > > > When the register was first set, it was set with its enable_mask set to 6 (i.e. BINNING & GMEM). But when we end up calling IsRegSet() when we encounter a draw call, the enable_mask then is set to 7.
> > > > So the old code checked bit by bit of the enable_mask. In this case it checks the first bit of the enable_mask during time of draw (I.e. 7), see that it's set, then checks the m_reg_is_set[0], which was never set to true because the enable_mask was 6 when that register was set. Note that bit 0 is referring to SYSMEM.
> > > 
> > > 
> > > oh, i think i was confused with ShaderEnableBit: so we have separate registers for different rendering mode? (i guess it should be called something like RenderMode or RenderPass?) Why do we need to have separate registers?
> > > For kBINNING, i guess it is for Binning pass. kGMEM is for Tiled rendering pass, and kSYSTEM, it is for Direct rendering pass. Correct? but if that is the case, why is following code?
> > > ```
> > > case RM6_RESOLVE:
> > >             m_shader_enable_bit = ShaderEnableBit::kGMEM;
> > >             break;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > We dont have any drawcall in resolve pass, i guess we don't need state register in this case?
> > 
> > 
> > Yes it looks like separate registers for different rendering mode. I didn't write this code, but that bit definitely exists in the set draw state pm4 packet, and the shader view uses this to determine the shader type (I.e. binning pass vs tiling pass VS, for example). And I guess that's why it's called ShaderEnableBit. So yeah, it's a little confusing and tbh I don't fully understand why the hardware works this way. Is there really 3 sets of registers in the hardware? Not sure...
> > I set RM6_RESOLVE to GMEM since I figure resolves can only happen with tiled rendering.
> 
> Ok, thanks. That makes sense. Maybe we could add some comments on this ShaderEnableBit to explain what it means, it is a bit confusing.

I'll leave that exercise for the future, since we should (low priority) figure out how this stuff really works before determining what to rename this to. Right now my primary concern is to get the state tracker working.

wangra-google (approved):

dive_core/common/emulate_pm4.h
```diff
@@ -134,26 +141,28 @@ class EmulateStateTracker
...
+    bool IsRegSet(uint32_t offset, ShaderEnableBit shader_enable_bit) const;
 
 private:
     static constexpr size_t kNumRegs = 0xffff + 1;
     uint32_t                m_reg[kShaderEnableBitCount][kNumRegs];
     uint8_t                 m_reg_is_set[kShaderEnableBitCount][(kNumRegs / 8) + 1];
     uint32_t                m_enable_mask = (1u << kShaderEnableBitCount) - 1;
     DiveVector<uint32_t>    m_enable_mask_stack;
+    bool                    m_shader_enable_bit_valid = false;
+    ShaderEnableBit         m_shader_enable_bit = ShaderEnableBit::kBINNING;
```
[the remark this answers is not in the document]
shanminchao: Good point! Done as requested!

wangra-google (approved):

dive_core/common/emulate_pm4.cpp
```diff
@@ -48,6 +48,53 @@ bool EmulateStateTracker::OnPacket(const IMemoryManager &mem_manager,
...
+            // should be paired with RM6_BIN_RENDER_START only if RM6_BIN_VISIBILITY exist, end of
+            // tiled mode
+            break;
+            // This is emitted at the begining of the resolve pass
+        case RM6_RESOLVE:
+            m_shader_enable_bit = ShaderEnableBit::kGMEM;
+            break;
+            // This is emitted for each dispatch
+        case RM6_COMPUTE:
+            break;
```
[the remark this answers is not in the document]
shanminchao: Oh oops, you're right.
I did remove it from RM6_COMPUTE though, since I actually don't believe that state reset needed to be there. But restored the 2 other ones. Thanks for catching it!

dive_core/common/emulate_pm4.cpp
```diff
@@ -48,6 +48,57 @@ bool EmulateStateTracker::OnPacket(const IMemoryManager &mem_manager,
...
+        a6xx_marker marker = static_cast<a6xx_marker>(packet.u32All0 & 0xf);
+        m_shader_enable_bit_valid = true;
+        switch (marker)
+        {
+            // This is emitted at the begining of the render pass if tiled rendering mode is
+            // disabled
+        case RM6_BYPASS:
+            m_shader_enable_bit = ShaderEnableBit::kSYSMEM;
+            break;
+            // This is emitted at the begining of the binning pass, although the binning pass
```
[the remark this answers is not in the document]
shanminchao: Fixed

footballhead (approved):

PR #430, opened 2025-10-02
GFXR: Allocate memory that RenderDoc expects

footballhead:
RenderDoc replaces external memory with non-external memory when replaying a .RDC file. If the memory type for the external memory does not match the non-external memory then RenderDoc throws an error. To avoid this, if we detect external memory, allocate what we think RenderDoc wants instead of what's in the .GFXR file. While this allows RenderDoc to open .RDC files, anything that uses external memory will be black since RenderDoc doesn't support it.

Based on testing, imported AHardwareBuffers use memoryTypeIndex 1 but RenderDoc wants it to be 0. We can't call vkGetImageMemoryRequirements since (apart from dedicated memory) we don't know which image this memory will be bound to.

Changing memoryTypeIndex in DiveVulkanReplayConsumer is brittle since the superclass implementation could overwrite memoryTypeIndex. However, it reduces the number of modifications made to GFXR; this should reduce the number of merge conflicts when we perform the next subtree pull.

This is protected by a system property until we use the RenderDoc in-application API to detect if RenderDoc is capturing. This allows the developer to toggle this behavior if it's causing issues.

### DIFF
--- a/gfxr_ext/decode/dive_vulkan_replay_consumer.cpp
+++ b/gfxr_ext/decode/dive_vulkan_replay_consumer.cpp
@@ -23,6 +23,10 @@ limitations under the License.
 #include <numeric>
 #include <limits>
 
+#include "graphics/vulkan_struct_get_pnext.h"
+#include "util/logging.h"
+#include "util/platform.h"
+
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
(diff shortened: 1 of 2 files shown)


gfxr_ext/decode/dive_vulkan_replay_consumer.cpp
```diff
@@ -590,6 +594,50 @@ StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator)
...
+void DiveVulkanReplayConsumer::Process_vkAllocateMemory(
+const ApiCallInfo&                                   call_info,
+VkResult                                             returnValue,
+format::HandleId                                     device,
+StructPointerDecoder<Decoded_VkMemoryAllocateInfo>*  pAllocateInfo,
+StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator,
+HandlePointerDecoder<VkDeviceMemory>*                pMemory)
+{
+    // TODO: b/448083729 - Use RenderDoc in-application API to detect if RenderDoc is capturing
+    if (util::platform::GetEnv("debug.gfxrecon.renderdoc_hack") == "1")
```
wangra-google, line 606: curious who sets `debug.gfxrecon.renderdoc_hack` to 1?

footballhead: Sure thing, it looks something like this:

1. `adb shell setprop debug.gfxrecon.renderdoc_hack 1` (required for OpenXR captures; optional for Vulkan captures that don't use external memory)
2. Push their GFXR/GFXA to the device
3. Open RenderDoc
4. Select replay context
5. Select Launch Application tab
    1. Executable Path: `com.lunarg.gfxreconstruct.replay/android.app.NativeActivity`
    2. Intent arguments: `--es "args" '"--loop-single-frame --loop-single-frame-count 0 /sdcard/Download/my_capture.gfxr"'`
6. Launch
7. Capture Frame(s) Immediately
8. Double-click the item in the Captures collected box

footballhead: That assumes replay is already built and installed. Otherwise you also need:

1. `scripts\build_android.bat Debug`
2. `python third_party\gfxreconstruct\android\scripts\gfxrecon.py install-apk install\gfxr-replay.apk`
3. `adb shell appops set --uid com.lunarg.gfxreconstruct.replay MANAGE_EXTERNAL_STORAGE allow`

wangra-google (approved):

gfxr_ext/decode/dive_vulkan_replay_consumer.cpp
wangra-google: cool! thanks :)

PR #431, opened 2025-10-02
Fixes Trace Stats outputs

GrantComm:
Fixes bug in trace stats outputs caused by using vectors instead of sets for viewports, window scissors, and shader references

### DIFF
--- a/trace_stats/trace_stats.cpp
+++ b/trace_stats/trace_stats.cpp
@@ -210,7 +210,7 @@ void TraceStats::GatherTraceStats(const Dive::CaptureMetadata &meta_data,
                 capture_stats.m_shader_ref_set.insert(info.m_shader_references[ref]);
     }
 
-    stats_list[Dive::Stats::kNumBinnigPasses] = capture_stats.m_num_binning_passes;
+    stats_list[Dive::Stats::kNumBinningPasses] = capture_stats.m_num_binning_passes;
     stats_list[Dive::Stats::kNumTilingPasses] = capture_stats.m_num_tiling_passes;
 
     if (!capture_stats.m_event_num_indices.empty())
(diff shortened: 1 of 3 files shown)


ui/window_scissors_stats_model.cpp
```diff
@@ -33,11 +33,12 @@ WindowScissorsStatsModel::WindowScissorsStatsModel(QObject *parent) :
 }
 
 //--------------------------------------------------------------------------------------------------
-void WindowScissorsStatsModel::LoadData(std::vector<Dive::WindowScissor> window_scissors)
+void WindowScissorsStatsModel::LoadData(std::set<Dive::WindowScissor> window_scissors)
```
wangra-google, line 36: const std::set<Dive::WindowScissor>&

trace_stats/trace_stats.cpp
```diff
@@ -87,6 +87,18 @@ namespace Dive
         stats_list[Dive::Stats::k##type##Resolves]++; \
     } while (0)
 
+#define PRINT_FIELD(name, value, last_item)                                           \
```
wangra-google, line 90: nit: use `is_last_item` to indicate that it is a bool

wangra-google (approved):

trace_stats/trace_stats.cpp
```diff
@@ -271,36 +271,59 @@ void TraceStats::PrintTraceStats(const CaptureStats &capture_stats, std::ostream
 
     for (uint32_t i = 0; i < Dive::Stats::kNumStats; ++i)
     {
-        ostream << kStatDescriptions[i] << ": " << stats_list[i] << "\n";
+        if (i != Stats::kNumBinnigPasses && i != Stats::kNumTilingPasses)
```
shanminchao, line 274: I don't think this check was in the original trace_stats code.
Why are we making an exception of these 2 stats, and how does this additional code change the printout? Do you have an example?

GrantComm: I originally included those 2 stats at the top of the print output:

image


This change puts those two back under the Window Scissors section:
https://github.com/google/dive/blob/2be2df6e60d31a72427e631a8d09faf0db449d65/trace_stats/main.cpp#L431

wangra-google (approved):

trace_stats/trace_stats.cpp
GrantComm: Fixed the typo.

I have those in the enum since I use it to populate the ui display as well.

PR #433, opened 2025-10-07
Only capture non-passthrough mode for compositor

RenfengLiu:


wangra-google (dismissed):

scripts/dive_compositor_capture.sh
```diff
@@ -18,62 +18,86 @@ set -eux
...
+    # Don't know when the capture has finished so defer to the user.
+    # Ideally, they have `adb logcat -s gfxrecon` running in another terminal.
+    echo "Press any key to retrieve capture"
+    read
+    adb shell setprop debug.gfxrecon.capture_android_trigger false
+    # Easiest to just pull the entire directory.
+    adb pull ${COMPOSITOR_CAPTURE_DIR} ./captures
+    # Clean up so that we're not constantly pulling the same files.
+    # Also causes names to be printed to output so the user knows.
+    for file in $(adb shell ls "${COMPOSITOR_CAPTURE_DIR}" | tr -d '\r'); do
```
footballhead, line 73: where does the carriage return come from and why do we need to remove it?

scripts/dive_compositor_capture.sh
```diff
@@ -18,62 +18,86 @@ set -eux
...
-done
+}
+
+cleanup() {
+  adb shell rm -rf ${COMPOSITOR_CAPTURE_DIR}
+  adb shell rm /data/local/debug/vulkan/libVkLayer_gfxreconstruct.so
+  adb shell setprop cpm.gfxr_layer 0
+  adb shell setprop persist.compositor.protected_context "\"\""
+  adb shell setprop persist.compositor.tiled_rendering "\"\""
+  adb shell setprop compositor.lazy_depth_buffer "\"\""
```
footballhead, line 85: optional: you can avoid escaping by combining single and double quotes. I think it looks cleaner but that's personal choice. I can see what it may not be more readable since the reader could get visually confused with the number of quotes
```suggestion
  adb shell setprop persist.compositor.protected_context '""'
  adb shell setprop persist.compositor.tiled_rendering '""'
  adb shell setprop compositor.lazy_depth_buffer '""'
```

footballhead (dismissed):

PR #460, opened 2025-10-20
Cleanup for some AnalyzeDialog signals, structs, functions related to loading

angela28chen:
Changes:
- Rename `AnalyzeDialog` signals to not start with prefix On* which is conventionally used for slots.
- Combine `OnNewFileOpened` and `ReloadCapture` signals into `LoadCapture`
- Add GFXA files to ReplayArtifactsPaths
- Added some log messages

### DIFF
--- a/capture_service/CMakeLists.txt
+++ b/capture_service/CMakeLists.txt
@@ -28,6 +28,15 @@ else()
   set(COMMAND_UTILS_SRC "command_utils.cc")
 endif()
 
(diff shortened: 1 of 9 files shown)


angela28chen:
I think ReplayArtifactsPaths probably shouldn't live under ui/ but I'm not sure where else to put it for now

angela28chen:
Manual Testing:
- Windows : TODO
- Linux: TODO

PR #463, opened 2025-10-20
Always clean up for RunReplayGfxrScript

RenfengLiu:
This is the first step on cleanup. Probably need to apply RAII for all the functions that execute the adb commands.

### DIFF
--- a/ui/main_window.cpp
+++ b/ui/main_window.cpp
@@ -1479,7 +1479,7 @@ void MainWindow::OnFileLoaded()
 void MainWindow::OnOpenFile()
 {
     QString supported_files = QStringLiteral(
-    "Dive files (*.rd);;GFXR files (*.gfxr);;All files (*.*)");
+    "Supported Files (*.rd *.gfxr);;Dive files (*.rd);;GFXR files (*.gfxr);;All files (*.*)");
     QString file_name = QFileDialog::getOpenFileName(this,
                                                      "Open Document",
                                                      Settings::Get()->ReadLastFilePath(),


wangra-google:
> I think this needs a larger refactor like classes for the different variants of replay to do their custom setup/cleanup stuff. I'm not sure if a partial refactor and the new struct here is that useful? RAII for functions that execute adb commands, we've sort of been doing a lot of that in AndroidDevice::CleanupDevice() I believe.
> 
> My suggestion is that for now we should:
> 
> 1. Add these PM4 property cleanup stuff to AndroidDevice::CleanupDevice().
> 2. Add other replay cleanup stuff like `RunReplayProfilingBinary(): CLEANUP` to `AndroidDevice::CleanupDevice()` as well
> 3. Add UI menu option for manual trigger device cleanup as discussed here: b/453201471. This is intended to be useful after crashes.
> 
> If we do decide to keep this, I suggest renaming to be clearer it's cleanup for PM4 replay specifically, also addressing the cleanup for the other replay: RunReplayProfilingBinary()

Those are valid suggestions. for 3), it would be really useful, but @RenfengLiu: how big do you think the change would it be? if it is a big one, maybe it is better to put it in a separate PR, and depending on the size of the change, we may not include it in the binaries that we are going to share just to avoid introducing more bugs. 

@GrantComm where do you think would be the best location to put the UI element of 3) from UI perspective?

RenfengLiu:
> > I think this needs a larger refactor like classes for the different variants of replay to do their custom setup/cleanup stuff. I'm not sure if a partial refactor and the new struct here is that useful? RAII for functions that execute adb commands, we've sort of been doing a lot of that in AndroidDevice::CleanupDevice() I believe.
> > My suggestion is that for now we should:
> > 
> > 1. Add these PM4 property cleanup stuff to AndroidDevice::CleanupDevice().
> > 2. Add other replay cleanup stuff like `RunReplayProfilingBinary(): CLEANUP` to `AndroidDevice::CleanupDevice()` as well
> > 3. Add UI menu option for manual trigger device cleanup as discussed here: b/453201471. This is intended to be useful after crashes.
> > 
> > If we do decide to keep this, I suggest renaming to be clearer it's cleanup for PM4 replay specifically, also addressing the cleanup for the other replay: RunReplayProfilingBinary()
> 
> Those are valid suggestions. for 3), it would be really useful, but @RenfengLiu: how big do you think the change would it be? if it is a big one, maybe it is better to put it in a separate PR, and depending on the size of the change, we may not include it in the binaries that we are going to share just to avoid introducing more bugs.
> 
> @GrantComm where do you think would be the best location to put the UI element of 3) from UI perspective?

I've updated to do clean up for the function, and for the device level as well

adding the button would be better in anther PR, I think.

wangra-google:
> > > I think this needs a larger refactor like classes for the different variants of replay to do their custom setup/cleanup stuff. I'm not sure if a partial refactor and the new struct here is that useful? RAII for functions that execute adb commands, we've sort of been doing a lot of that in AndroidDevice::CleanupDevice() I believe.
> > > My suggestion is that for now we should:
> > > 
> > > 1. Add these PM4 property cleanup stuff to AndroidDevice::CleanupDevice().
> > > 2. Add other replay cleanup stuff like `RunReplayProfilingBinary(): CLEANUP` to `AndroidDevice::CleanupDevice()` as well
> > > 3. Add UI menu option for manual trigger device cleanup as discussed here: b/453201471. This is intended to be useful after crashes.
> > > 
> > > If we do decide to keep this, I suggest renaming to be clearer it's cleanup for PM4 replay specifically, also addressing the cleanup for the other replay: RunReplayProfilingBinary()
> > 
> > 
> > Those are valid suggestions. for 3), it would be really useful, but @RenfengLiu: how big do you think the change would it be? if it is a big one, maybe it is better to put it in a separate PR, and depending on the size of the change, we may not include it in the binaries that we are going to share just to avoid introducing more bugs.
> > @GrantComm where do you think would be the best location to put the UI element of 3) from UI perspective?
> 
> I've updated to do clean up for the function, and for the device level as well
> 
> adding the button would be better in anther PR, I think.

agree, that should be a separate PR.

capture_service/device_mgr.cc
```diff
@@ -536,8 +539,23 @@ absl::Status AndroidDevice::CleanupDevice()
...
     Adb().Run("shell settings delete global gpu_debug_layers_gles").IgnoreError();
 
+    // clean up for gfxr renderdoc capture
     UnsetSystemProperty(Adb(), kReplayCreateRenderDocCapture).IgnoreError();
 
+    // clean up for gfxr replay app
+    Adb()
+    .Run(absl::StrFormat("shell appops set %s MANAGE_EXTERNAL_STORAGE default", kGfxrReplayAppName))
+    .IgnoreError();
+    Adb().Run(absl::StrFormat("uninstall %s", kGfxrReplayAppName)).IgnoreError();
```
angela28chen: Thanks for adding the uninstall here!

capture_service/device_mgr.cc
```diff
@@ -893,6 +912,11 @@ absl::Status DeviceManager::RunReplayProfilingBinary(const GfxrReplaySettings &s
     std::string remote_profiling_dir = absl::StrFormat("%s/%s",
                                                        kTargetPath,
                                                        kProfilingPluginFolderName);
+    Defer       cleanup([&]() {
+        LOGD("RunReplayProfilingBinary(): CLEANUP\n");
+        std::string clean_cmd = absl::StrFormat("shell rm -rf -- %s", remote_profiling_dir);
```
angela28chen: Do we want this in AndroidDevice::CleanupDevice() too?

capture_service/device_mgr.cc
```diff
@@ -493,6 +494,8 @@ absl::Status AndroidDevice::CleanupDevice()
 
     UnpinGpuClock().IgnoreError();
     Adb().Run("shell setprop compositor.high_priority 1").IgnoreError();
+    Adb().Run("shell setprop openxr.enable_frame_delimiter false").IgnoreError();
```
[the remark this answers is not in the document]
RenfengLiu: done

PR #465, opened 2025-10-21
Background load of trace stats.

hysw:
Better trace stats loading.

- Mutex protect data_core.
- Running trace stats on worker thread.
- Use a thread pool for disassembly.
- Allow trace stats collection to be cancelled.

### DIFF
--- a/dive_core/context.h
+++ b/dive_core/context.h
@@ -1,0 +1,152 @@
+/*
+ Copyright 2025 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+#include <atomic>
+#include <memory>
+
+#pragma once
+
+namespace Dive
+{
+
+// Usage:
+//
+// Async operations:
+//     absl::Status LongOperation(const Dive::Context& context)
+//     {
+//         if (context.Cancelled())
+//         {
+//             return absl::Status(absl::StatusCode::kCancelled, "cancelled");
+//         }
+//         ...
+//     }
+// CLI:
+//     LongOperation(Dive::Context::Background());
+// UI:
+//     Dive::SimpleContext op_context;
+//     void Start() // qt slot
+//     {
+//         op_context = Dive::SimpleContext::Create();
+//         run_async([context = op_context](){LongOperation(context)});
+//     }
+//     void Stop()
+//     {
+//         if(!op_context.IsNull())
+//         {
+//             op_context->Cancel();
+//         }
+//     }
+class Context;
+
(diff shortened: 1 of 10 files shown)


shanminchao:
If I'm understanding correctly: You have a worker thread that handles the trace_stats load. And within that thread, you create individual thread in the threadpool for the disassembly of each shader.

So the question is: Is the worker thread even necessary now that the majority of the load time (i.e. disassembly) is offloaded to threads? You mentioned in chat that the loading of trace_stats has been reduced from 16s to 0.8s (amazing!), and we can probably optimize the remaining 0.8s a little further. So wondering if the complexity and overhead with managing a worker thread here is worth it.

trace_stats/trace_stats.cpp
```diff
@@ -223,8 +328,27 @@ void TraceStats::GatherTraceStats(const Dive::CaptureMetadata &meta_data,
...
+    auto       task_count = static_cast<unsigned int>(meta_data.m_shaders.size());
+    thread_pool.Start(thread_pool.SuggestedNumberOfWorkers(task_count));
+    for (const Dive::Disassembly &disassembly : meta_data.m_shaders)
+    {
+        thread_pool.Run([&context, &disassembly]() {
+            if (context.Cancelled())
+            {
+                return;
+            }
+            disassembly.GetNumInstructions();
```
shanminchao: Perhaps we should just have an explicit Disassemble function to make it more clear what this is doing, rather than rely on a side effect

hysw: Done.

dive_core/context.h
```diff
@@ -0,0 +1,152 @@
+/*
+ Copyright 2019 Google LLC
```
shanminchao: nit: Should probably update this to 2025

dive_core/context.h
```diff
@@ -0,0 +1,152 @@
...
+
+// A cancelable context implementation.
+class SimpleContextToken : public ContextToken
+{
+public:
+    bool Cancelled() const override { return m_cancelled.load(); }
+    void Cancel() { m_cancelled.store(true); }
+
+private:
+    std::atomic_bool m_cancelled = false;
```
shanminchao: Dumb question: Instead of creating this infrastructure involving ContextHolder, ContextToken, etc, why not just pass around an atomic_bool as the "context". This whole context infrastructure exists essentially just to pass around a thread-safe bool, isn't it? Can you explain its purpose, or better yet add some comments in the code describing its intention?

Also, not saying we don't need it; just glancing through the code and trying to understand the intention

hysw: The intentional is to have somewhat an analog of https://pkg.go.dev/context , although there's no goroutine / channel in C++ . e.g.
```c++
class Context {
public:
  std::optional<std::chrono::time_point<std::chrono::steady_clock>> Deadline() const;
  bool Cancelled() const; //  Channel does not exist :(
  absl::Status Error() const;
};
```

`std::shared_ptr<std::atomic_bool>` is the simplest implementation of a cancelable context, so I'm starting with that.

shanminchao: Ok fair enough, thanks for answering my question!

wangra-google (approved):

hysw:
Test: (linux)
- opening gfxr + rd file, look at trace stats
- opening rd file, look at trace stats
- opening gfxr file, run dump pm4, look at trace stats

wangra-google (approved):

shanminchao (approved):

PR #466, opened 2025-10-21
Make a button for deleting replay artifacts

angela28chen:
Rather than always deleting at the beginning of GFXR replay series, make the option to delete it a button

analyze capture dialog with updated buttons and warning


angela28chen:
Manual testing:
- Windows GFXR replay with cube + PM4 + GPU timing, pressing delete button 
- Linux WIP

PR #477, opened 2025-10-28
GFXR: Improve developer build times by hiding git SHA

footballhead:
Every time I changed a git branch or made a commit, it would take 1+ minute to build GFXR. This was caused by many large C++ files including a header file which contained the git SHA; every time the SHA changed, those files would need to be recompiled. Since they were large files, this took a long time.

Instead of consuming the SHA directly, hide it behind a function call and let the linker deal with it. This reduces the number of targets to recompile every time the git SHA changes. The only exception is version.rc which wants a string literal.


wangra-google:
Thanks for making this change. 
The only concern i have is that this is touching many files in the gfxr code which would make future integration more difficult?

wangra-google:
> FWIW I found that my laptop was on the "balanced" power profile. If I increase Processor Power Management > Minimum Processor State > Plugged in to 100%, it takes 46s without this fix and 6s after.
> 
> My test:
> 
> ```shell
> # Build the layer once to generate a baseline
> gradlew layer:assembleDebug -Parm64-v8a
> 
> # Generate a new SHA without code changes
> git commit --allow-empty -m "test"
> 
> # Rebuild layer with only the SHA change
> gradlew layer:assembleDebug -Parm64-v8a
> ```

that is great, thanks for verifying the actual gain.

footballhead:
I'm going to create a PR upstream to increase the impact of the fix as well as reduce the number of conflicts.

footballhead:
> I'm going to create a PR upstream to increase the impact of the fix as well as reduce the number of conflicts.

Given this plan, I'm going to close this PR. I'll reopen if something changes.

PR #478, opened 2025-10-28
Adding double quotes to adb push/pull commands

elviscapiaq:
- It resolves the issue with capture paths containing spaces and special characters.


RenfengLiu (approved):

elviscapiaq:
> Thanks for making this change, i am wondering if it would be more readable to use `R"(...)"`?
> 
> So
> 
> ```
> absl::StrFormat("push \"%s\" \"%s\"", local_gfxr_path, remote_gfxr_path)
> ```
> 
> becomes
> 
> ```
> absl::StrFormat(R"(push "%s" "%s")", local_gfxr_path, remote_gfxr_path)
> ```

I agree, will update with that. Also I verified this change on Linux and WIndows.

hysw:
For shell, prefer `'` quote, also you may want to escape all the quote in strings.

PR #479, opened 2025-10-28
Allow GFXR replay with validation layer

footballhead:
Add the --validation_layer CLI option which sets use_validation_layer. If set, push and enable the Vulkan Validation Layer. This only works for normal replay to keep things simple; this does not preclude expanding it to work with other replay modes in the future.

Includes prebuilt Validation Layers for 1.4.328.1. Downloaded from: https://github.com/KhronosGroup/Vulkan-ValidationLayers/releases/download/vulkan-sdk-1.4.328.1/android-binaries-1.4.328.1.tar.gz

The prebuilt was chosen since compiling from the repo required pulling in a lot of dependencies. Choosing the prebuilt doesn't preclude us from going down this path in the future.

### DIFF
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -611,7 +611,7 @@ absl::Status AndroidDevice::CleanupDevice()
     UnsetSystemProperty(Adb(), kReplayCreateRenderDocCapture).IgnoreError();
 
     // cleanup for gfxr replay with validation layer
-    UninstallVulkanLayer(adb, kGfxrReplayAppName, kVkValidationLayerName).IgnoreError();
+    UninstallVulkanLayer(Adb(), kGfxrReplayAppName, kVkValidationLayerName).IgnoreError();
 
     // clean up for gfxr replay app
     Adb()


angela28chen (approved):

capture_service/device_mgr.cc
```diff
@@ -87,8 +87,11 @@ absl::Status EnableVulkanLayer(const AdbSession &adb,
     RETURN_IF_ERROR(adb.Run(absl::StrFormat("shell settings put global gpu_debug_app %s", app)));
     RETURN_IF_ERROR(
     adb.Run(absl::StrFormat("shell settings put global gpu_debug_layers %s", layer)));
-    RETURN_IF_ERROR(
-    adb.Run(absl::StrFormat("shell settings put global gpu_debug_layer_app %s", layer_app)));
+    if (!layer_app.empty())
+    {
+        RETURN_IF_ERROR(
```
RenfengLiu: Not direct related to this PR, but if there's error happened during this adb commands, how should we revert the previous related commands, this might need to be addressed in a more systematic way.

capture_service/device_mgr.cc
```diff
@@ -97,6 +100,31 @@ absl::Status IsAppInstalled(const AdbSession &adb, std::string_view package)
...
+    RETURN_IF_ERROR(
+    adb.Run(absl::StrFormat(R"(push "%s" "%s")",
+                            ResolveAndroidLibPath(std::string(layer_filename)).generic_string(),
+                            kTargetPath)));
+    RETURN_IF_ERROR(adb.Run(
+    absl::StrFormat(R"(shell run-as %s cp "%s/%s" .)", package, kTargetPath, layer_filename)));
+    return absl::OkStatus();
+}
+
+absl::Status UninstallVulkanLayer(const AdbSession &adb,
```
RenfengLiu: @angela28chen  do you think should we also call this in the device cleanup, so that it has all the cleanup stuff.

RenfengLiu (dismissed):

capture_service/device_mgr.cc
footballhead: While possible, I think this adds complexity without major benefits. If this function fails then we can't feasibly continue with replay; I'd expect the cleanup logic to run and delete all these properties. The downside is we need to be diligent about keeping the cleanup up-to-date

RenfengLiu (dismissed):

angela28chen (approved):

PR #480, opened 2025-10-28
Upgrade CMake minimum to 3.22

footballhead:
While Debian stable provides 3.31, the Android Gradle Plugin (used by GFXR) is hardcoded to install and use 3.22 from the Android SDK. (See [CmakeLocator.kt](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/cxx/configure/CmakeLocator.kt?q=CmakeLocator.kt)). While we could specify a particular CMake version in the build.gradle file, it's unclear to me whether the Android SDK Manager would automatically download that version or if the user needs to have that version specifically installed.

3.22 was chosen since it is still greater than the 3.18 required for https://github.com/google/dive/pull/450 but works with GFXR without modification.

The new behavior of `find_package(Python3 COMPONENTS Interpreter)` introduced in CMake 3.15 was causing CMake to use the Python installed by jurplel/install-qt-action (we wanted a Python3_FIND_STRATEGY of VERSION but LOCATION was being used since CMP0094 is set to NEW by CMake >= 3.15). Since Mako was only installed in the system python, the build would fail. This was fixed by telling jurplel/install-qt-action to not install Python (use the system Python instead). Eliminating extraneous Python installations should improve maintainability by reducing complexity.

Remove cmake_minimum_required from subdirectories since it is redundant.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -79,6 +79,9 @@ jobs:
           submodules: recursive
       - name: Install Qt
         uses: jurplel/install-qt-action@v3
+        with:
+          # Minimize the number of Python installations so that CMake chooses the system Python with python3-mako.
+          setup-python: false
       - name: Install dependency
         run: |
           sudo apt-get update --yes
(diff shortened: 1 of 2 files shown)


footballhead:
The SDK manager does have newer CMake versions, not sure if we can get Gradle to download one of those 
image

footballhead:
The Linux builds are failing because `find_package (Python3 COMPONENTS Interpreter)` is choosing ` /opt/hostedtoolcache/Python/3.11.13/x64/bin/python3.11` instead of the system python, but we only install mako for the system python.

PR #482, opened 2025-10-29
Adding component_files library

angela28chen:
Planning to use this library to reduce some repetition in Dive code around getting associated file names

### DIFF
--- a/utils/component_files.cpp
+++ b/utils/component_files.cpp
@@ -16,16 +16,28 @@ limitations under the License.
 
 #include "component_files.h"
 
-namespace Dive::ComponentFiles
+#include "absl/status/status.h"
+#include "absl/strings/str_format.h"
+
+namespace Dive
 {
 
-absl::StatusOr<Dive::ComponentFiles::Paths> GetHostPaths(const std::filesystem::path &parent_dir,
-                                                         const std::string           &gfxr_stem)
+absl::StatusOr<Dive::ComponentFilePaths> GetComponentFilesHostPaths(
+const std::filesystem::path &parent_dir,
+const std::string           &gfxr_stem)
 {
-    assert(!gfxr_stem.empty());
-    assert(!parent_dir.empty());
+    if (gfxr_stem.empty())
+    {
+        return absl::FailedPreconditionError(absl::StrFormat("gfxr_stem cannot be empty"));
+    }
 
-    Dive::ComponentFiles::Paths artifacts = {};
+    if (parent_dir.empty())
+    {
+        return absl::FailedPreconditionError(absl::StrFormat("parent_dir cannot be empty"));
+    }
+
+    Dive::ComponentFileConstants constants;
+    Dive::ComponentFilePaths     artifacts = {};
 
     // check format of gfxr_stem
     // dots in filename are allowed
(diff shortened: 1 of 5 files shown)


utils/component_files.h
```diff
@@ -0,0 +1,45 @@
...
+limitations under the License.
+*/
+
+#include <filesystem>
+#include <string>
+
+#include "absl/status/statusor.h"
+#include "component_files_constants.h"
+
+namespace Dive::ComponentFiles
```
hysw: https://abseil.io/tips/130

"Don’t nest deeply: a single top-level namespace per project gets the same result without long/complicated names, with less exposure to accidents, without causing surprise for new engineers, and without the need to build any tooling."

utils/component_files.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include "component_files.h"
+
+namespace Dive::ComponentFiles
+{
+
+absl::StatusOr<Dive::ComponentFiles::Paths> GetHostPaths(const std::filesystem::path &parent_dir,
+                                                         const std::string           &gfxr_stem)
+{
+    assert(!gfxr_stem.empty());
```
wangra-google: would it be better to return absl::Status here? something like:
```
return absl::FailedPreconditionError("gfxr_stem cannot be empty");
```

utils/component_files_test.cpp
```diff
@@ -0,0 +1,135 @@
...
+    std::string gfxr_stem = "PLACEHOLDER_trim_trigger_ID";
+
+    Paths expected_res = {};
+    expected_res.gfxr = parent_dir / "PLACEHOLDER_trim_trigger_ID.gfxr";
+    expected_res.gfxa = parent_dir / "PLACEHOLDER_asset_file_ID.gfxa";
+    expected_res.perf_counter_csv = parent_dir /
+                                    "PLACEHOLDER_trim_trigger_ID_profiling_metrics.csv";
+    expected_res.gpu_timing_csv = parent_dir / "PLACEHOLDER_trim_trigger_ID_gpu_time.csv";
+    expected_res.pm4_rd = parent_dir / "PLACEHOLDER_trim_trigger_ID.rd";
+    expected_res.renderdoc_rdc = parent_dir / "PLACEHOLDER_trim_trigger_ID_capture.rdc";
```
wangra-google: do we need to test `expected_res.screenshot_png`?

utils/component_files_test.cpp
```diff
@@ -0,0 +1,135 @@
...
```
wangra-google: same as above, seems that `screenshot_png` is missing

utils/component_files.cpp
```diff
@@ -0,0 +1,67 @@
...
+    artifacts.perf_counter_csv = parent_dir /
+                                 (gfxr_stem + Dive::ComponentFiles::kProfilingMetricsHostSuffix +
+                                  Dive::ComponentFiles::kCsvExt);
+    artifacts.gpu_timing_csv = parent_dir /
+                               (gfxr_stem + Dive::ComponentFiles::kGpuTimingHostSuffix +
+                                Dive::ComponentFiles::kCsvExt);
+    artifacts.pm4_rd = parent_dir / (gfxr_stem + Dive::ComponentFiles::kRdExt);
+    artifacts.renderdoc_rdc = parent_dir / (gfxr_stem + Dive::ComponentFiles::kRenderDocHostSuffix +
+                                            Dive::ComponentFiles::kRdcExt);
+    return artifacts;
```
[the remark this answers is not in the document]
angela28chen: Oops thanks, added this + unit tests

utils/component_files.cpp
```diff
@@ -0,0 +1,67 @@
...
+    std::string gfxa_stem = gfxr_stem;
+    size_t      pos = gfxa_stem.find(Dive::ComponentFiles::kGfxrFileNameSubstr);
+    if (pos == std::string::npos)
+    {
+        return absl::FailedPreconditionError(
+        absl::StrFormat("unexpected name for gfxr file: %s, expecting name containing: %s",
+                        gfxr_stem,
+                        Dive::ComponentFiles::kGfxrFileNameSubstr));
+    }
+    int gfxr_string_length = sizeof(Dive::ComponentFiles::kGfxrFileNameSubstr) / sizeof(char);
```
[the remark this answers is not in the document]
wangra-google: make sense. thanks

PR #484, opened 2025-10-29
Fix shader disassembly race condition.

hysw:


wangra-google (approved):

hysw:
Test: (linux) open multiple traces, verify no crash with assert.

PR #489, opened 2025-10-30
Adds white icon for exit

GrantComm:
Screenshot 2025-10-30 at 10 12 21 AM


angela28chen (approved):
LGTM, do we have some kind of set dimensions for icons used in dive or is it as long as the aspect ratio works?

GrantComm:
> LGTM, do we have some kind of set dimensions for icons used in dive or is it as long as the aspect ratio works?

There aren't currently set dimensions. Do you think that should be done in a followup pr just so it's standard across all of them?

hysw:
Icon size is somewhat relevant when you got high dpi interfaces, e.g. you can test on Linux with `QT_SCALE_FACTOR=4 dive`

PR #490, opened 2025-10-30
Refactor for AnalyzeDialog opening file flow

angela28chen:
Changes:
- Removed some constants from `capture_service/constants.h` and used `utils/component_files_constants.h` instead, to consolidate logic around determining local file paths for the different components of a dive capture
- Added some utility functions to `component_files` library around identifying a file extension
- Some signals/slots renamed to have the "On" prefix for slots only.
- Added debug warning for if people try to open `.dive` files.
- Added warning message box for if an invalid state is detected when the user tries to analyze
- GFXR capture must be loaded before running "Analyze", and it's impossible for the user to change the loaded GFXR capture from within `AnalyzeDialog`. I think this sets us up better for the eventual save/load feature. Tweaked some things to align with this.

### DIFF
--- a/ui/main_window.cpp
+++ b/ui/main_window.cpp
@@ -1644,12 +1644,9 @@ void MainWindow::OnAnalyzeCapture()
 {
     if (!m_gfxr_capture_loaded && !m_correlated_capture_loaded)
     {
+        qDebug() << "Not launching AnalyzeDialog because GFXR file not succesfully loaded, instead "
+                    "prompting user to load a file";
         OnOpenFile();
-    }
-    bool gfxr_capture_loaded = (m_gfxr_capture_loaded || m_correlated_capture_loaded);
-    if (!gfxr_capture_loaded)
-    {
-        qDebug() << "Not launching AnalyzeDialog because GFXR file not succesfully loaded";
         return;
     }
     qDebug() << "Launching AnalyzeDialog with: " << m_capture_file;
--- a/ui/trace_window.cpp
+++ b/ui/trace_window.cpp
@@ -1012,6 +1012,13 @@ void GfxrCaptureWorker::run()
         {
             absl::StatusOr<Dive::ComponentFilePaths>
             ret = Dive::GetComponentFilesHostPaths(m_target_capture_dir, gfxr_stem);
+            if (!ret.ok())
+            {
+                std::string err_msg = absl::StrFormat("Failed to get component files: %s",
+                                                      ret.status().message());
+                qDebug() << err_msg.c_str();
+                return;
+            }
             component_files = *ret;
         }
 
--- a/utils/component_files.cpp
+++ b/utils/component_files.cpp
@@ -82,26 +82,22 @@ const std::string           &gfxr_stem)
 
 bool IsGfxrFile(std::filesystem::path file)
 {
-    ComponentFileConstants consts;
-    return file.extension() == consts.kGfxrExt;
+    return file.extension() == ComponentFileConstants::kGfxrExt;
 }
 
 bool IsPngFile(std::filesystem::path file)
 {
-    ComponentFileConstants consts;
-    return file.extension() == consts.kPngExt;
+    return file.extension() == ComponentFileConstants::kPngExt;
 }
 
 bool IsDiveFile(std::filesystem::path file)
 {
-    ComponentFileConstants consts;
-    return file.extension() == consts.kDiveExt;
+    return file.extension() == ComponentFileConstants::kDiveExt;
 }
 
 bool IsRdFile(std::filesystem::path file)
 {
-    ComponentFileConstants consts;
-    return file.extension() == consts.kRdExt;
+    return file.extension() == ComponentFileConstants::kRdExt;
 }
 
 }  // namespace Dive


hysw (approved):

angela28chen:
Manual Testing:
- Windows: Load GFXR with and without rd, launch analyze via icon and via menu option, gfxr replay and check that the loading still seems to work
- Linux: TODO

Note that the behaviour changes a little for if you try to analyze without a capture loaded. You'll have to click analyze again after the capture has been loaded.

ui/trace_window.cpp
```diff
@@ -1005,17 +1008,23 @@ void GfxrCaptureWorker::run()
 
     if (!original_screenshot_path.empty() && !gfxr_stem.empty())
     {
-        std::filesystem::path new_screenshot_filename = gfxr_stem + Dive::kPngSuffix;
-        std::filesystem::path png_new_path = m_target_capture_dir;
-        png_new_path /= new_screenshot_filename;
+        Dive::ComponentFilePaths component_files = {};
```
wangra-google: i guess if you dont care about the ret, you can 
```
Dive::ComponentFilePaths component_files =
    *Dive::GetComponentFilesHostPaths(m_target_capture_dir, gfxr_stem);
```

or if you do want to handle the ret
```
if (auto ret = Dive::GetComponentFilesHostPaths(m_target_capture_dir, gfxr_stem); ret.ok())
{
    component_files = *ret;
} else {
    // Handle the error (ret.status())
}
```

ui/analyze_window.cpp
```diff
@@ -445,108 +433,73 @@ void AnalyzeDialog::OnDeviceListRefresh()
...
-
-        // Check if the required asset file exists.
-        bool asset_file_exists = std::filesystem::exists(asset_file_path);
-
-        if (!asset_file_exists)
+        qDebug() << "OnAnalyzeCaptureStarted(): empty filename";
+        return;
+    }
+    std::filesystem::path    local_gfxr_parse = file_path.toStdString();
+    Dive::ComponentFilePaths component_paths = {};
```
wangra-google: if (auto ret = Dive::GetComponentFilesHostPaths(local_gfxr_parse.parent_path(),
                                               local_gfxr_parse.stem().string()); ret.ok())
{
    component_paths = *ret;
} else {
    // Handle the error (ret.status())
}

ui/main_window.cpp
```diff
@@ -1636,13 +1646,14 @@ void MainWindow::OnAnalyzeCapture()
     {
         OnOpenFile();
     }
```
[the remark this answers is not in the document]
angela28chen: Thanks, I will adjust this section to reflect this + I want to address the case where user has a .rd file loaded and tries to analyze

ui/analyze_window.cpp
wangra-google: Oh, thanks for bring this up. I completely missed that comment. I feel that this way uses the C++ language feature specifically designed for initialization and checking (if with initializer, available since C++17). It also keeps the logic for initializing, checking, and extracting the value together in one line, improving local coherence. 

But after thinking about it, i think it is just a personal preference and both are fine as long as we can keep a consistent style in the code, and i am ok if you prefer to keep it this way. (Still interesting to have this discussion :) )

ui/main_window.cpp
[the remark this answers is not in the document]
angela28chen: I tested with my current changes and rd file loaded + tries to analyze -> user is prompted to open a file

ui/trace_window.cpp
angela28chen: Added

wangra-google (approved):

GrantComm:
Can you briefly explain how the removal of the user's ability to change the loaded GFXR capture from within AnalyzeDialog sets us up better for the save/load feature? My thought was that it reduced to number of clicks required to replay multiple captures. Additionally, if the user accidentally opened the wrong capture, they would be able to quickly open the desired one without having to close the analyze window, load a different capture, and then reopen the window

angela28chen:
> Can you briefly explain how the removal of the user's ability to change the loaded GFXR capture from within AnalyzeDialog sets us up better for the save/load feature? My thought was that it reduced to number of clicks required to replay multiple captures. Additionally, if the user accidentally opened the wrong capture, they would be able to quickly open the desired one without having to close the analyze window, load a different capture, and then reopen the window

I was thinking that since capture loading isn't trivial, it's better to force the user to use the same flow when possible so that we have it better tested. Removing that functionality (to change gfxr capture from within analyzedialog) also makes the signals from analyzedialog->mainwindow simpler. Lastly, the user will be able to see the main window overlay (visual updates for file loading I think?) better with the analyze dialog window out of the way.

I think making things simpler around loading will make the save/load feature easier to implement and test. I also want to reinforce the idea of "the current session / opened dive capture" so I think it makes sense to have it top-level and not something that is easily changeable from within AnalyzeDialog.

To me the convenience/performance tradeoff in this PR is a really minor, that's +2 clicks with no wait time to close and re-open analyze dialog. Loading GFXR is unavoidable both ways, and will take comparatively much longer.

hysw (approved):

PR #491, opened 2025-10-31
Capture File Manager.

hysw:
- Manage loaded capture and async loading.

### DIFF
--- a/ui/capture_file_manager.h
+++ b/ui/capture_file_manager.h
@@ -1,0 +1,134 @@
+/*
+ Copyright 2025 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+#pragma once
+
+#include <memory>
+
+#include <QMetaType>
+#include <QObject>
+#include <QReadWriteLock>
+
+#include "context.h"
+#include "file_path.h"
+
+class QThread;
+namespace Dive
+{
+class DataCore;
+struct CaptureStats;
+struct ComponentFilePaths;
+}  // namespace Dive
+
+struct LoadFileResult
+{
+
+    enum class Status
+    {
+        kSuccess,
+        kUnknown,
+        kFileIoError,
+        kCorruptData,
+        kVersionError,
+        kParseFailure,
+        kUnsupportedFile,
+        kGfxaAssetMissing,
+    };
+
+    enum class FileType
+    {
+        kUnknown,  // Load failure
+        kCorrelatedFiles,
+        kRdFile,
+        kGfxrFile,
+    };
+
+    Status status = Status::kUnknown;
+
+    FileType                 file_type = FileType::kUnknown;
+    Dive::FilePath           reference = {};
+    Dive::ComponentFilePaths components = {};
+};
+
+class CaptureFileManager : public QObject
+{
+    Q_OBJECT
+public:
+    static void RegisterCustomMetaType();
+
+    explicit CaptureFileManager(QObject *parent);
+    ~CaptureFileManager();
+
+    CaptureFileManager(const CaptureFileManager &) = delete;
+    CaptureFileManager(CaptureFileManager &&) = delete;
+    CaptureFileManager &operator=(const CaptureFileManager &) = delete;
+    CaptureFileManager &operator=(CaptureFileManager &&) = delete;
+
+    void Start(Dive::DataCore &);
+
+    // Locking:
+    QReadWriteLock &GetDataCoreLock() { return m_data_core_lock; }
+
+    Dive::ComponentFilePaths ResolveComponents(const Dive::FilePath &reference);
+    void LoadFile(const Dive::FilePath &reference, const Dive::ComponentFilePaths &components);
+
+    void GatherTraceStats();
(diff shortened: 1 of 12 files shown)


ui/debug_utils.h
```diff
@@ -0,0 +1,91 @@
...
+private:
+    FuncT                m_func;
+    std::tuple<ArgsT...> m_args;
+
+    void Invoke() { std::apply(m_func, m_args); }
+};
+#endif
+
+template<typename FuncT, typename... ArgsT>
+DebugScoped(FuncT&&, ArgsT&&... args) -> DebugScoped<FuncT, ArgsT...>;
```
footballhead: This looks like a custom absl::Cleanup. Why not use absl::Cleanup?

ui/debug_utils.h
```diff
@@ -0,0 +1,91 @@
...
+
+template<typename FuncT> inline auto DebugScopedTimer(FuncT&& func)
```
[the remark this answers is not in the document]
hysw: Done.

Qt calls stop watch QElapsedTimer :(

ui/main_window.h
```diff
@@ -73,6 +75,7 @@ class QSortFilterProxyModel;
 class QAbstractProxyModel;
 class FrameTabView;
 class QScrollArea;
+class CaptureFileManager;
```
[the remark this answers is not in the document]
hysw: Removed the include.

wangra-google (approved):

utils/component_files.h
```diff
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+#pragma once
```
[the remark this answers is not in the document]
angela28chen: Oops, thanks peter for catching this. I went ahead and made a separate PR just because I think that capture manager is going to take longer to review (am reading through it now)

https://github.com/google/dive/pull/516

ui/debug_utils.h
angela28chen: Is there a reason this is under ui/? If it can be used more widely and it isn't using Qt headers, should this be under utils/?

ui/custom_metatypes.cpp
```diff
@@ -0,0 +1,37 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+#include "custom_metatypes.h"
+
+#include <QMetaType>
+
+#include "context.h"
```
angela28chen: nit: `ui/context.h` and etc?

ui/capture_file_manager.h
```diff
@@ -0,0 +1,138 @@
...
+#include "context.h"
+#include "file_path.h"
+
+class QThread;
+namespace Dive
+{
+class DataCore;
+struct CaptureStats;
+struct ComponentFilePaths;
+}  // namespace Dive
```
angela28chen: Just wondering, I'm assuming the purpose of having Dive namespace declarations here is for forward declaration? Is this just how it should be done, or is there something strange with how we have set things up that doesn't allow us to easily include header files?

ui/main_window.cpp
```diff
@@ -148,6 +152,43 @@ std::optional<std::filesystem::path> ResolveAssetPath(const std::string &name)
     return std::nullopt;
 }
 
+//--------------------------------------------------------------------------------------------------
+void OnLoadingFailure(QWidget               &parent,
+                      LoadFileResult::Status reason,
+                      const Dive::FilePath  &reference_path,
+                      const QString         &aux)
```
angela28chen: nit: rather than "aux", more descriptive name like "local_gfxa_file_path" or "aux_file_path"

ui/capture_file_manager.h
```diff
@@ -0,0 +1,138 @@
...
+        kVersionError,
+        kParseFailure,
+        kUnsupportedFile,
+        kGfxaAssetMissing,
+    };
+
+    enum class FileType
+    {
+        kUnknown,  // Load failure
+        kDiveFile,
```
angela28chen: I'm wondering if we should just mark this as unimplemented for now since I don't know if anyone has tried to load a .dive file since pre-GFXR days and if it's not a flow that is well-supported we should probably just disallow for the time being? My understanding is we don't have a functional save feature so .dive files aren't being produced either

hysw: This is actually .rd + .gfxr file case

ui/capture_file_manager.h
hysw: Usually you want to include a forward declaration headers, e.g. `<iosfwd>` for `<iostream>`, however we don't have those forward declaration headers.

ui/capture_file_manager.h
angela28chen: OK, wow that is confusing since don't we have files with extension `.dive` (that are basically .rd) too? I think adding a comment/todo here to clarify this is (.rd+.gfxr) would be helpful for future people skimming this part of the code.

ui/capture_file_manager.h
footballhead: Yes I believe that's a good rule of thumb. The two main pros are improved compile times and avoiding unnecessary recompilation. The Google C++ style has a good overview of forward declaration pros and cons. FWIW they don't recommend them https://google.github.io/styleguide/cppguide.html#Forward_Declarations

footballhead:
I'm going to be relying on other folks to review this PR since I don't have a lot of knowledge of this part of the code. As such, I won't be approving.

ui/capture_file_manager.h
hysw: There are enough hazard in the dive project, e.g. include path dependency. We are stuck with forward declare a bunch of stuff until we fix dependency stuff.

e.g. include ComponentFilePaths directly breaks plugin example, so there's #523

wangra-google (dismissed):

hysw:
Tested on Linux
- Open a file
- Open a file, then opening a different file
- Open a file, then capture PM4
- GFXR capture openxr/vulkan application

PR #493, opened 2025-11-03
Fix validation layer file cleanup

footballhead:
I noticed in the logs that we were trying to clean up the wrong file:

```
> adb -s [serial] shell rm -rf -- "/data/local/tmp/VK_LAYER_KHRONOS_validation"
```

UninstallVulkanLayer was being called with the layer name, not the
filename of the .so. This was leaving the validation layer on the
device. This has been fixed.


elviscapiaq (approved):
Thanks for fixing it.

GrantComm (approved):

PR #499, opened 2025-11-04
Right aligns perf counter table

GrantComm:
Right aligns perf counter table
Screenshot 2025-11-04 at 1 36 57 PM

### DIFF
--- a/ui/perf_counter_model.cpp
+++ b/ui/perf_counter_model.cpp
@@ -132,6 +132,11 @@ QVariant PerfCounterModel::data(const QModelIndex &index, int role) const
         return QVariant(Qt::AlignRight);
     }
 
+    if (role != Qt::DisplayRole)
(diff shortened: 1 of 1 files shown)


ui/perf_counter_model.cpp
```diff
@@ -159,12 +164,7 @@ QVariant PerfCounterModel::data(const QModelIndex &index, int role) const
     if (static_cast<size_t>(metric_col_index) < record.m_metric_values.size())
     {
         const auto &metric_value = record.m_metric_values.at(metric_col_index);
-
-        if (role == Qt::DisplayRole)
```
hysw: Why is this part removed? I'm pretty sure there are roles that we don't handle.

wangra-google (approved):

hysw (approved):

PR #510, opened 2025-11-05
Prevent crash when directory contains unexpected number of files

elviscapiaq:
- The 'areTimestampsCurrent' function previously assumed the presence of exactly three specific files (gfxr, gfxa, and rd). If the directory contained any other number of files, the application would crash.
- This change modifies the file count check to allow an arbitrary number of files in the directory. If Dive requires the specific three files and they are not all present, the dialog is now displayed instead of causing a crash.

### DIFF
--- a/ui/trace_window.cpp
+++ b/ui/trace_window.cpp
@@ -40,6 +40,8 @@
 #include <filesystem>
 #include <string>
 #include <vector>
+#include <QButtonGroup>
+#include <QRadioButton>
 
 #include "absl/status/status.h"
 #include "absl/strings/str_cat.h"
@@ -54,7 +56,9 @@
 namespace
 {
 const std::vector<std::string> kAppTypes{ "Vulkan APK", "OpenXR APK", "Command Line Application" };
-}
+const int                      kGfxrCaptureButtonId = 1;
+const int                      kPm4CaptureButtonId = 2;
+}  // namespace
 
 // =================================================================================================
 // TraceDialog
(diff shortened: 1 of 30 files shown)


ui/trace_window.cpp
```diff
@@ -811,31 +811,26 @@ bool GfxrCaptureWorker::areTimestampsCurrent(Dive::AndroidDevice     *device,
                                              std::vector<std::string> previous_timestamps)
```
wangra-google: should this be `const std::vector<std::string>&`?

angela28chen: I see that we're using vectors with timestamps whose elements correspond to files in `m_file_list`. Is there the chance that files in `m_file_list` could change order between prev/current timestamps (since it seems that it's possible for the number of them to change?) If so maybe we need to compare timestamps using a data structure that associates file names with the timestamps rather than just an ordering.

ui/trace_window.cpp
```diff
@@ -811,31 +811,26 @@ bool GfxrCaptureWorker::areTimestampsCurrent(Dive::AndroidDevice     *device,
...
-    absl::StatusOr<std::string> third_current_timestamp = device->Adb().RunAndGetResult(
-    get_third_current_timestamp_command);
-
-    current_time_stamps.push_back(first_current_timestamp->data());
-    current_time_stamps.push_back(second_current_timestamp->data());
-    current_time_stamps.push_back(third_current_timestamp->data());
-    return (current_time_stamps[0] == previous_timestamps[0] &&
-            current_time_stamps[1] == previous_timestamps[1] &&
-            current_time_stamps[2] == previous_timestamps[2]);
+
```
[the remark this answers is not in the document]
elviscapiaq: I don't think it is neccesary as we build previous_timestamps from m_file_list. To improve the checking process, I am using now a map to compare file sizes to avoid order issues as Angela suggested.

ui/trace_window.h
```diff
@@ -70,9 +70,10 @@ class GfxrCaptureWorker : public QThread
 
     // Appends/increments the numerical suffix "_#" to target_capture_path for a fresh directory, if
     // the directory already exists
-    void                    SetGfxrTargetCaptureDir(const std::string &target_capture_dir);
-    bool                    areTimestampsCurrent(Dive::AndroidDevice     *device,
-                                                 std::vector<std::string> previous_timestamps);
+    void SetGfxrTargetCaptureDir(const std::string &target_capture_dir);
+    bool areTimestampsCurrent(Dive::AndroidDevice                      *device,
```
angela28chen: nit: naming seems unconventional for this and the below function, maybe capitalize the first letter?

angela28chen (approved):

ui/trace_window.cpp
wangra-google: i guess this `current_timestamps.clear();` is not needed since current_timestamps is a local var anyways?

PR #511, opened 2025-11-05
Add MacOS support

RenfengLiu:


capture_service/constants.h
```diff
@@ -77,4 +77,10 @@ inline constexpr char kReplayCreateRenderDocCapture[] = "debug.gfxrecon.replay.r
 inline constexpr char kRenderDocCaptureLayerName[] = "VK_LAYER_RENDERDOC_Capture";
 inline constexpr char kRenderDocAppName[] = "org.renderdoc.renderdoccmd.arm64";
 
+#if defined(__APPLE__)
+inline constexpr char kPythonCommand[] = "python3";
```
hysw: We should be using `python3` name regardless platform.

dive_core/CMakeLists.txt
```diff
@@ -116,18 +116,18 @@ if(MSVC)
...
-endif()
+endif() # MSVC
 
 file(GLOB_RECURSE HDR_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.h")
 file(GLOB_RECURSE SRC_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
 list(FILTER SRC_FILES EXCLUDE REGEX "_tests.cpp$")
 list(FILTER SRC_FILES EXCLUDE REGEX "${CMAKE_CURRENT_SOURCE_DIR}/tests/.*")
 
-if(DIVE_NO_DISASSEMBLY)
+if(DIVE_NO_DISASSEMBLY OR APPLE)
```
wangra-google: curious, the shader lib doesn't compile for apple?

RenfengLiu: no, it doesn't not, and I tried to fix it, but it takes way longer than I expected. might be next steps

wangra-google (dismissed):

capture_service/device_mgr.cc
```diff
@@ -795,7 +796,7 @@ absl::Status DeviceManager::DeployReplayApk(const std::string &serial)
 
     std::string replay_apk_path = ResolveAndroidLibPath(kGfxrReplayApkName, "").generic_string();
     std::string recon_py_path = ResolveAndroidLibPath(kGfxrReconPyPath, "").generic_string();
-    std::string cmd = absl::StrFormat("python %s install-apk %s -s %s",
+    std::string cmd = absl::StrFormat("python3 %s install-apk %s -s %s",
```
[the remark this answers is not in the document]
hysw: imo, `python` command should not exist, Python 2 and Python 3 are two different languages. It's unfortunate that we need actually trying to find it

capture_service/device_mgr.cc
```diff
@@ -39,10 +39,64 @@ limitations under the License.
 namespace Dive
 {
 
-namespace
+namespace {
+
+std::string GetPythonPath() {
```
angela28chen: Maybe GetPythonPath() and ValidatePythonPath() could be put under `utils/`?

angela28chen (approved):

wangra-google (approved):

capture_service/device_mgr.cc
RenfengLiu: this is currently only used in this file, maybe we can move to there once it get more widely used.

PR #512, opened 2025-11-05
Use tab visibility instead of remove and add tabs.

hysw:


ui/main_window.cpp
```diff
@@ -2389,6 +2326,11 @@ void MainWindow::OnTabViewSearchBarVisibilityChange(bool isHidden)
 //--------------------------------------------------------------------------------------------------
 void MainWindow::OnTabViewChange()
 {
+    if (m_tabs_updating)
```
wangra-google: just curious, why do we need to check `m_tabs_updating` here?  isn't it set only in UpdateTabAvailability to true then back to false? and OnTabViewChange is only called at the end of that func?

ui/main_window.cpp
```diff
@@ -2336,16 +2256,33 @@ QString MainWindow::StrippedName(const QString &fullFileName)
...
+    SetTabAvailable(m_tab_widget,
+                    m_tabs.gfxr_vulkan_command_arguments,
+                    mask & TabMaskBits::kGfxrVulkanCommandArguments);
+    SetTabAvailable(m_tab_widget, m_tabs.perf_counter, mask & TabMaskBits::kPerfCounters);
+    SetTabAvailable(m_tab_widget, m_tabs.gpu_timing, mask & TabMaskBits::kGpuTiming);
+    SetTabAvailable(m_tab_widget, m_tabs.frame, mask & TabMaskBits::kFrame);
+#if defined(ENABLE_CAPTURE_BUFFERS)
+    SetTabAvailable(m_tab_widget, m_tabs.buffer, mask & TabMaskBits::kBuffer);
+#endif
 #ifndef NDEBUG
```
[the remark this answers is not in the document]
hysw: Probably in some later PR. I need to keep PR scope reasonable.

ui/main_window.cpp
hysw: Qt might call this function. Especially when we have a bunch of `processEvents()` floating around.

wangra-google: which means this func might be happening in a different thread?

wangra-google (approved):

ui/main_window.cpp
wangra-google: sorry still a bit confused, if m_tabs_updating is only changed to true then to false within UpdateTabAvailability, and OnTabViewChange will never be running in parallel with UpdateTabAvailability, what would trigger the case where m_tabs_updating is true here?

hysw: I could remove that if I can make assumption about `UpdateTabAvailability` won't be call `OnTabViewChange` before the end.

There's at least two paths:

Probably common case:
> `UpdateTabAvailability()` 
> => `QTabWidget::setTabVisible()` conditionally
> => `QTabWidget::setCurrentIndex()` 
> => `QTabBar::setCurrentIndex()`
> => `emit QTabBar::currentChanged()`
> => `QTabWidgetPrivate::showTab()`
> => `emit QTabWidget::currentChanged()`
> => `OnTabViewChange()`

Then there's the `processEvents` path in Dive code that may also trigger it
> `UpdateTabAvailability()` 
> => `QTabWidget::setTabVisible()`
> => there's a bunch of layout and widget attribute update
> => we might have connected to one of those signal
> => `processEvents()` shows up a lot in our code
> at this point, basically anything that responsive to user input might be called

wangra-google: oh, i see. thanks for the detailed explanation.

hysw:
Tested on windows.
Open rd file, gfxr file, and gfxr file with rd file in same directory.

PR #518, opened 2025-11-06
Enable capture vulkan command line application in gfxr

RenfengLiu:
currently the gfxr capture for command line vulkan application is implemented, but the UI still shows the option. this PR adds the ability to take gfxr capture for it.


RenfengLiu:
I think I messed up with rebase close

PR #522, opened 2025-11-06
Only show frame count when gpu time is enabled

GrantComm:
Screenshot 2025-11-06 at 1 30 58 PM

Screenshot 2025-11-06 at 1 31 06 PM


wangra-google (approved):

wangra-google:
> > The frame count also applies to the "normal" type of replay
> 
> @angela28chen @wangra-google Do you think it makes sense to have it shown when none of the options are selected or gpu time is selected but then hidden if just metrics are selected, dump pm4, or renderdoc are selected? Or would that be confusing?

@angela28chen has a valid point, and thanks for bring this up. I think the case when nothing is selected is mostly only for debugging purpose? do you both see any benefit of exposing this to the users (my worry is that exposing this option to the users might be confusing if it is not useful)? I am wondering if we should only allow that in debug build but not release?

PR #526, opened 2025-11-07
Some clang-format updates for a more consistent style

angela28chen:
- Enforce newline at EOF
- Enforce pointer alignment to the right (`int *a` rather than `int* a`)


angela28chen:
Had more discussion offline about formatting

https://github.com/google/dive/pull/545

PR #527, opened 2025-11-07
Removes unnecessary QApplication::processEvents calls

GrantComm:
Removes QApplication::processEvents calls.
**Manual Testing:**
Linux: Perform replay from analyze dialog and check that button text is being updated, correlate gfxr, pm4, gpu time, counter data.


wangra-google:
more general question: from what i understand, `QApplication::processEvents();` will force event loop to flush, what is the general rule of deciding if we should call this function or not?

wangra-google (approved):

PR #528, opened 2025-11-07
Don't depends on libopenxr_loader for the UI

RenfengLiu:
UI should not depends on libopenxr_loader
```
~/src/dive$ otool -L ./install/dive
./install/dive:
...
        @rpath/libopenxr_loader.dylib (compatibility version 0.0.0, current version 0.0.0)
...
```

### DIFF
--- a/dive_core/CMakeLists.txt
+++ b/dive_core/CMakeLists.txt
@@ -17,7 +17,7 @@
 project(dive_core)
 
 set(CMAKE_CXX_STANDARD 20)
-
+include(CheckFunctionExists)
 if(MSVC)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /MP")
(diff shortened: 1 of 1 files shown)


hysw (dismissed):

dive_core/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@
 project(dive_core)
 
 set(CMAKE_CXX_STANDARD 20)
-
+include(CheckFunctionExists)
```
wangra-google: do we use CHECK_FUNCTION_EXISTS_ADD_LIBRARIES in the proj somewhere? i couldn't find it. or is this for something else?

wangra-google: oh, i see. good to know. thanks

PR #529, opened 2025-11-07
clang-format *.cc in GitHub Action

footballhead:
.cc is accepted as a file extension for C++ source files (likely thanks to Google style) but the Action was only checking for .cpp files. This meant that files like capture_service/device_mgr.cc were not being checked.


RenfengLiu (approved):

angela28chen (approved):
Thanks for catching this!

PR #534, opened 2025-11-10
Allow installing the GFXR replay app

footballhead:
We are encountering issues installing the replay app. While we work on a better solution, toggle the developer option that allows us to reliably adb install.


capture_service/device_mgr.cc
hysw: We should replace these with `absl::Cleanup`, possibly in another PR.

footballhead: https://github.com/google/dive/pull/548

PR #539, opened 2025-11-11
Removes package filter, adds warning

GrantComm:
- Adds a warning above the package selector when advanced options are not displayed

**Manual Testing:**
Linux: 
- Open capture dialog with advanced options disabled
- Open capture dialog with advanced options enabled
- Open capture dialog with advanced options enabled and select "All" filter, close dialog, Open capture dialog with advanced options disabled (ensure only debuggable applications are show)

**Advanced Options Disabled**
Screenshot 2025-11-10 at 4 39 30 PM

**Advanced Options Enabled**
Screenshot 2025-11-19 at 3 34 28 PM

### DIFF
--- a/ui/analyze_window.h
+++ b/ui/analyze_window.h
@@ -18,7 +18,6 @@
 #include <future>
 #include <optional>
 #include "capture_service/device_mgr.h"
-#include "package_filter.h"
 #include "dive_core/available_metrics.h"
 #include "utils/component_files.h"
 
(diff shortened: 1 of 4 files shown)


wangra-google:
I verified, the non-debuggable apks are not working 
image

So probably it is a good idea to not have the filter shown to the users.

However, i am thinking since Peter added the option "Show Advanced Options", maybe would show the filter only when this option is on? (by default we hide the filter, and only allow users to choose debuggable apks)

PR #542, opened 2025-11-11
Update replay options.

hysw:
- Add "Show Advanced Options" in file menu.
- Change normal replay to custom replay.
- Hide custom replay by default.
- Make a checkbox for perf counter replay.


wangra-google (approved):

hysw:
Also when metrics is not available
image

angela28chen (approved):
Thanks for the images, it appears there is a bit of vertical compression that makes some fields look cropped, does that have to do with the default height for the analyze window or something?

PR #547, opened 2025-11-12
Using ccache to cache the build

RenfengLiu:


RenfengLiu:
> Great idea! Should we add this to the other builds as well?

Yes, it should be able to apply to other build, let's see how it works for Mac first.

PR #515, opened 2025-11-06
Updates Application types

GrantComm:
Updates Application types for pm4 and gfxr capture.

Manual Testing:

Linux: 
- Setup and perform GFXR capture on CubeXr
- Setup and perform PM4 capture on CubeXR and GLES-OpenXR application

**Options for Gfxr**
Screenshot 2025-11-13 at 3 03 33 PM

**Options for PM4**
Screenshot 2025-11-13 at 3 03 26 PM

PR #553, opened 2025-11-12
Refactoring dive_client_cli

elviscapiaq:
- Command Registry: Implemented a CommandRegistry that maps the Command enum to a struct containing the Name, Description, Validator, and Executor. We no longer need to update ParseFlag, UnparseFlag, main switch statements, and help text separately.

- Context & Options Struct: Instead of passing 8 different strings (serial, package, arch, etc.) into every function, I grouped them into a GlobalOptions struct. I created a CommandContext that holds the DeviceManager and the GlobalOptions. This makes the function signatures uniform (absl::Status(const CommandContext& ctx)), allowing for polymorphism.

- Decoupled Validation: Each command in the registry has a Validator function. Now the main logic is: Parse Flags -> Select Device -> Validate Command Requirements -> Execute.

- Adds a function to get target device.

- To improve the log output for errors, it was replaced bool functions with absl::Status.


capture_service/dive_client_cli.cc
```diff
@@ -683,263 +686,161 @@ void TriggerGfxrCapture(Dive::DeviceManager& mgr, const std::string& gfxr_captur
...
-    // loop_single_frame_count is parsed from --gfxr_replay_flags
 
-    Dive::DeviceManager mgr;
-    auto                list = mgr.ListDevice();
-    if (list.empty())
-    {
-        std::cout << "No device connected" << std::endl;
-        return 0;
-    }
+    GlobalOptions opts;
```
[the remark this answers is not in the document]
elviscapiaq: Done.

capture_service/dive_client_cli.h
```diff
@@ -0,0 +1,153 @@
...
+                                const std::string&   gfxr_capture_directory);
+
+// Command Executors.
+absl::Status CmdListDevice(const CommandContext& ctx);
+absl::Status CmdListPackage(const CommandContext& ctx);
+absl::Status CmdRunPackage(const CommandContext& ctx);
+absl::Status CmdRunAndCapture(const CommandContext& ctx);
+absl::Status CmdGfxrCapture(const CommandContext& ctx);
+absl::Status CmdGfxrReplay(const CommandContext& ctx);
+absl::Status CmdCleanup(const CommandContext& ctx);
```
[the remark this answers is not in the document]
elviscapiaq: Done.

capture_service/dive_client_cli.h
```diff
@@ -0,0 +1,154 @@
...
+{
+// Overload for parsing GfxrReplayOptions.
+bool AbslParseFlag(absl::string_view text, GfxrReplayOptions* run_type, std::string* error);
+
+// Overload for converting GfxrReplayOptions back to a string.
+std::string AbslUnparseFlag(GfxrReplayOptions run_type);
+}  // namespace Dive
+
+// Selects and sets up the target device based on the serial flag.
+absl::StatusOr<Dive::AndroidDevice*> GetTargetDevice(Dive::DeviceManager& mgr,
```
hysw: Most of those could be in `Dive::` namespace.

hysw (commented):
LGTM, although I'm not that familiar with cli code, so I'll defer to other to approve.

RenfengLiu (dismissed):

capture_service/dive_client_cli.h
```diff
@@ -0,0 +1,154 @@
...
+
+// Overload for converting the Command enum back to a string.
+std::string AbslUnparseFlag(Command command);
+
+// Abseil flags parsing uses ADL. GfxrReplayOptions is in the Dive namespace so AbslParseFlag
+// and AbslUnparseFlag need to be as well.
+namespace Dive
+{
+// Overload for parsing GfxrReplayOptions.
+bool AbslParseFlag(absl::string_view text, GfxrReplayOptions* run_type, std::string* error);
```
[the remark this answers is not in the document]
hysw: These overloads must be added to the same namespace where the type is defined, so that they can be discovered by Argument-Dependent Lookup (ADL).
https://abseil.io/docs/cpp/guides/flags#custom

Also [With few exceptions, place code in a namespace.](https://google.github.io/styleguide/cppguide.html#Namespaces)

Although if it's used in multiple places, it's probably better putting in a shared header somewhere.

angela28chen (commented):
Thanks for looking at refactoring dive_client_cli! Two structural concerns I have:
1. Should we be adding a header file for a CLI? My understanding is that header files are used when the code is a library that other code will depend on, but this is not the case here.
2. I feel like the CommandContext is useful to group together but it should be a refactor at the device_mgr level since most of those parameters are just passed down. And if it's refactored at that level, then UI/ can make use of it as well to simplify code there. Does that make sense?

capture_service/dive_client_cli.cc
```diff
@@ -26,103 +28,192 @@ limitations under the License.
...
-    if (text == "list_device")
+    if (options.package.empty() && options.vulkan_command.empty())
     {
-        *command = Command::kListDevice;
-        return true;
+        return absl::InvalidArgumentError("Missing required flag: --package or --vulkan_command");
     }
-    if (text == "list_package")
+
+    static const std::set<std::string> valid_types = { "openxr", "vulkan", "vulkan_cli" };
```
wangra-google: maybe better to be consistent with https://github.com/google/dive/pull/515/files#r2528100218

capture_service/dive_client_cli.cc
```diff
@@ -26,103 +28,192 @@ limitations under the License.
...
+        .name = "run",
+        .description = "Run an app for manual testing or external capture.",
+        .validator = ValidateRunOptions,
+        .executor = CmdRunPackage,
+        },
+        },
+        {
+        Command::kRunAndCapture,
+        {
+        .name = "capture",
```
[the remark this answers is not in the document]
wangra-google: agree, it is confusing, maybe `kRunAndCapture` should be renamed to something like `kPM4Capture`. description should adjusted as well.

capture_service/dive_client_cli.cc
```diff
@@ -26,103 +28,192 @@ limitations under the License.
...
+                                                       options.replay_settings.remote_capture_path,
+                                                       "'. File must have a .gfxr extension."));
     }
-    if (text == "gfxr_capture")
+    return absl::OkStatus();
+}
+
+const std::map<Command, CommandMetadata>& GetCommandRegistry()
+{
+    static const auto* registry = new std::map<Command, CommandMetadata>{
```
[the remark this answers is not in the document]
elviscapiaq: Yes, it works, but following Google style, I updated using Ran's recommendation.

capture_service/dive_client_cli.h
elviscapiaq: Done.

capture_service/dive_client_cli.h
```diff
@@ -0,0 +1,154 @@
...
+
+// Retrieves a GFXR capture from the device and downloads it.
+absl::Status RetrieveGfxrCapture(Dive::DeviceManager& mgr,
+                                 const std::string&   gfxr_capture_directory);
+
+// Triggers a GFXR capture on the device, allowing for multiple captures and screenshot.
+absl::Status TriggerGfxrCapture(Dive::DeviceManager& mgr,
+                                const std::string&   gfxr_capture_directory);
+
+// Command Executors.
```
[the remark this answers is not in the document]
elviscapiaq: I removed the header file as we are not reusing it.

capture_service/dive_client_cli.cc
wangra-google: sorry a bit confused here, aren't we supposed to support following 4 types?
`vulkan_openxr, vulkan_non_openxr, vulkan_cli_non_openxr, gles_openxr`

capture_service/dive_client_cli.cc
```diff
@@ -22,104 +22,300 @@ limitations under the License.
...
+
 #include "android_application.h"
 #include "command_utils.h"
 #include "constants.h"
 #include "device_mgr.h"
 #include "network/tcp_client.h"
-#include "absl/strings/str_cat.h"
 #include "utils/version_info.h"
 
 using namespace std::chrono_literals;
```
hysw: Unused?

Also per [go/cstyle#Operator_Overloading](https://google.github.io/styleguide/cppguide.html#Operator_Overloading)
> Do not overload operator"", i.e., do not introduce user-defined literals. Do not use any such literals provided by others (including the standard library).

elviscapiaq: Done.

wangra-google (dismissed):

capture_service/dive_client_cli.cc
```diff
@@ -22,421 +22,389 @@ limitations under the License.
...
+                               ctx.options.gfxr_capture_file_dir);
+        break;
+    case AppType::kVulkanCLI_Non_OpenXR:
+        ret = device->SetupApp(ctx.options.vulkan_command,
+                               ctx.options.vulkan_command_args,
+                               Dive::ApplicationType::VULKAN_CLI,
+                               ctx.options.device_architecture,
+                               ctx.options.gfxr_capture_file_dir);
+        break;
+    case AppType::kGLES_OpenXR:
```
wangra-google: https://github.com/google/dive/pull/561 is adding a new type that you probably need to check

capture_service/dive_client_cli.cc
```diff
@@ -701,266 +654,396 @@ void TriggerGfxrCapture(Dive::DeviceManager& mgr, const std::string& gfxr_captur
...
+{
+    switch (type)
+    {
+    case AppType::kVulkan_OpenXR:
+        return "vulkan_openxr";
+    case AppType::kVulkan_Non_OpenXR:
+        return "vulkan_non_openxr";
+    case AppType::kVulkanCLI_Non_OpenXR:
+        return "vulkan_cli_non_openxr";
+    case AppType::kGLES_OpenXR:
```
wangra-google: need to handle the new case introduced by https://github.com/google/dive/pull/561

capture_service/dive_client_cli.cc
```diff
@@ -22,421 +22,389 @@ limitations under the License.
...
+    std::function<absl::Status(const CommandContext&)> executor;
+};
+
+namespace
+{
+
+// Helper to validate options common to run/capture commands.
+absl::Status ValidateRunOptions(const GlobalOptions& options)
+{
+    if (options.package.empty() && options.vulkan_command.empty())
```
wangra-google: would it be better if we add checks like:
- if vulkan_command is used, the app_type must be kVulkanCLI_Non_OpenXR. 
- If the command is present but the type is kVulkan_OpenXR, it is likely wrong.

capture_service/dive_client_cli.cc
```diff
@@ -701,266 +654,396 @@ void TriggerGfxrCapture(Dive::DeviceManager& mgr, const std::string& gfxr_captur
...
-    return ret.ok();
 }
 
+}  // namespace Dive
+
+ABSL_FLAG(Command,
+          command,
+          Command::kNone,
+          "The operation to perform. Available commands:\n"
+          "\tlist_device  : List connected Android devices.\n"
```
wangra-google: the descriptions seem to be duplication of the ones defined in GetCommandRegistry?

maybe we could have something like this that can be used for both cases to avoid duplication:
```
struct CommandInfo {
    const char* name;
    const char* description;
};

namespace
{
    constexpr std::array<std::pair<Command, CommandInfo>, 7> kCommandInfoMap
    {
        {
            {Command::kListDevice,   {"list_device", "List connected Android devices."}},
            ...
        }
    };
}
```

hysw: imo, `ABSL_FLAG` is probably not a good choice for the usecase. There's no subcommand support and it's missing a bunch of feature supporting for a semi-interactive cli tool. Although rewriting it will be a bit painful.

elviscapiaq: Done.

elviscapiaq:
I fixed the conflicts and also applied the requested changes.

src/dive/common/app_types.h
```diff
@@ -0,0 +1,74 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#pragma once
+
+#include <algorithm>
+#include <array>
+#include <string_view>
+
```
hysw: `namespace Dive`

src/dive/common/app_types.h
```diff
@@ -0,0 +1,74 @@
...
+struct AppTypeInfo
+{
+    AppType          type;
+    std::string_view ui_name;
+    std::string_view cli_name;
+    bool             is_gfxr_capture_supported;
+    std::string_view description;
+};
+
+inline constexpr std::array<AppTypeInfo, 5> kAppTypeInfos = { {
```
hysw: nit: use [designated initializer](https://en.cppreference.com/w/cpp/language/aggregate_initialization.html#Designated_initializers) and put type names in when it makes sense
```cpp
AppTypeInfo{ .type = AppType::kVulkan_OpenXR,
             .ui_name = "Vulkan (OpenXR)",
             .cli_name = "vulkan_openxr",
             .is_gfxr_capture_supported = true,
             .description = "For Vulkan OpenXR applications(apk)" },
```

capture_service/dive_client_cli.cc
```diff
@@ -710,266 +664,365 @@ void TriggerGfxrCapture(Dive::DeviceManager& mgr, const std::string& gfxr_captur
...
+    absl::Status   ret = selected_def->executor(ctx);
+    if (!ret.ok())
     {
-        PrintUsage();
-        res = true;
-        break;
-    }
+        std::cout << "Error executing command '" << selected_def->name << "': " << ret.message()
+                  << std::endl;
+        return 1;
```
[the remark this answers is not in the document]
elviscapiaq: Done.

capture_service/dive_client_cli.cc
```diff
@@ -710,266 +664,365 @@ void TriggerGfxrCapture(Dive::DeviceManager& mgr, const std::string& gfxr_captur
...
     }
 
-    return res ? 0 : 1;
-}
+    return 0;
+}
```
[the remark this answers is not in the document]
elviscapiaq: Done.

src/dive/common/app_types.h
elviscapiaq: Done.

src/dive/common/app_types.h
```diff
@@ -0,0 +1,74 @@
...
+  false,
+  "For GLES OpenXR applications(apk)" },
+{ AppType::kGLES_Non_OpenXR,
+  "GLES (Non-OpenXR)",
+  "gles_non_openxr",
+  false,
+  "for GLES applications(apk)" },
+} };
+
+inline constexpr size_t kNumGfxrCaptureAppTypes = std::count_if(kAppTypeInfos.begin(),
```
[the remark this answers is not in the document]
elviscapiaq: Done.

ui/trace_window.cpp
```diff
@@ -52,40 +52,14 @@
 #include "capture_service/device_mgr.h"
 #include "network/tcp_client.h"
 #include "utils/component_files.h"
+#include "common/app_types.h"
```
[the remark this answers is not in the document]
elviscapiaq: Done.

elviscapiaq:
I applied the requested minor changes. Ready for review.

angela28chen:
Can you please add a comment here outlining the manual tests you've done for this CLI change? Thanks

wangra-google (approved):

PR #554, opened 2025-11-12
Enable cpplint.

hysw:


wangra-google (approved):

hysw:
Some decision that need some consensus
 
Consolidated all the lint into a single workflow file, and renamed  
* v_18_1_8_on_dive_sources -> lint_clang_format
* gersemi_v_0_23_1_on_dive_sources -> lint_cmake_format

utils/string_utils.h
```diff
@@ -69,7 +69,7 @@ template<typename T> bool SafeConvertFromString(const std::string &s, T &out)
     {
         if constexpr (std::is_signed_v<T>)
         {
-            long long val = std::strtoll(start, &end, 10);
+            long long val = std::strtoll(start, &end, 10);  // NOLINT
```
footballhead: Not for this PR but should we look into absl::simpleAtoi as a replacement?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,51 @@
+name: Dive Lint
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the main branch
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  lint_cpplint:
```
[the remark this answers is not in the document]
hysw: It's more or less equivalent in this case, with a different organization in UI.

image
image

image
image

ui/shader_view.cpp
```diff
@@ -253,18 +253,17 @@ void ShaderView::paintEvent(QPaintEvent *event)
                     }
 
                     // Column 2: Address
-                    const uint32_t buffer_size = 256;
-                    char           buffer[buffer_size];
-                    snprintf(buffer, buffer_size, "%p", (void *)shader_info.GetShaderAddr());
+                    constexpr uint32_t kBufferSize = 256;
+                    char               buffer[kBufferSize];
+                    snprintf(buffer, kBufferSize, "%p", (void *)shader_info.GetShaderAddr());
```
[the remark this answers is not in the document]
hysw: This is modified in-place since there's a single instance of it. absl::strFormat is definitely better.

utils/string_utils.h
hysw: Yes.

Note: NOLINT is for `long long`.

hysw:
> Is there a chance that cpplint will conflict with clang-format?

I don't think so.

> Can you add the manual invocation to CONTRIBUTING.md where we have other info about linting?

Probably later. This version is for CI, the manual check was to validate the filter works on all files. For actual manual run, we want a stricter version.

wangra-google (dismissed):

.github/workflows/cmake-format.yml
```diff
@@ -1,30 +0,0 @@
...
-# events but only for the main branch
-on:
-  push:
-    branches: [ main ]
-  pull_request:
-
-# Cancel previous runs if a more recent commit is pushed.
-concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref }}
-  cancel-in-progress: true
```
footballhead: Do we still need this? I noticed that it's missing from lint.yml

wangra-google (approved):

scripts/dive_cpplint_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def auto_file_list(base_commit, dive_root):
+    result = subprocess.run(["git", "diff", base_commit, "--name-only"],
+                            capture_output=True, text=True, check=True)
+    return get_file_list(result.stdout.splitlines(), dive_root)
+
+
+def main(args):
+    files = []
```
footballhead: FYI one weird quirk of python is that `if`s are not scoped. so this line is technically not necessary. Though I appreciate the consistency with other languages for readability

scripts/dive_cpplint_wrapper.py
```diff
@@ -0,0 +1,154 @@
...
+        print("WARNING: Found out of tree file:", path)
+        return False
+    rel_path = path.resolve().relative_to(dive_root_resolved)
+    if path.suffix not in DIVE_CPPLINT_FILE_SUFFIX:
+        print("Skip non-C++ file:", path)
+        return False
+    if not rel_path.parts or rel_path.parts[0] not in DIVE_SOURCE_DIRS_SET:
+        print("Skip non-Dive code:", path)
+        return False
+    return True, ""
```
footballhead: The second element of the returned tuple is unused and can be removed
```suggestion
    return True
```

footballhead (dismissed):

scripts/dive_cpplint_wrapper.py
hysw: Done

wangra-google (approved):

ui/debug_utils.h
```diff
@@ -26,7 +26,7 @@ template<typename FuncT> class DebugScoped
 {
 public:
     explicit DebugScoped(FuncT&&) {};
-    ~DebugScoped() { std::ignore = this; }
+    ~DebugScoped() { (void)this; }
```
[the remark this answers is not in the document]
hysw: Telling compiler not be overly smart and emit unused variable warning. Apparently g3 cpplint likes void cast better

PR #558, opened 2025-11-13
Stricter lint and format check for local development.

hysw:


hysw:
With .clang-tidy, it's probably sufficient for local development. Will revisit this when other lint issues are fixed (e.g. formatting, etc)

PR #560, opened 2025-11-14
Resolve stats

shanminchao:
Added resolve-specific stats
Also cleaned up the way command hierarchy dealt with the various events, by making them all "event nodes"

### DIFF
--- a/dive_core/capture_event_info.h
+++ b/dive_core/capture_event_info.h
@@ -24,13 +24,13 @@
 namespace Dive
 {
 
-
 //--------------------------------------------------------------------------------------------------
 // Helper functions
 namespace Util
 {
 enum class EventType : uint8_t
 {
+    kUnknown,
     kDraw,
     kDispatch,
     kBlit,
@@ -47,26 +47,26 @@ enum class EventType : uint8_t
     kWaitForIdle,
     kWaitForMe,
     kEventWriteStart,
-    kEventWriteEnd,
+    kEventWriteEnd
 };
 
-bool IsEvent(const IMemoryManager &mem_manager,
-             uint32_t              submit_index,
-             uint64_t              addr,
-             uint32_t              opcode,
-             EmulateStateTracker  &state_tracker);
+bool IsEvent(const IMemoryManager      &mem_manager,
+             uint32_t                   submit_index,
+             uint64_t                   addr,
+             uint32_t                   opcode,
+             const EmulateStateTracker &state_tracker);
 
-EventType GetEventType(const IMemoryManager &mem_manager,
-                       uint32_t              submit_index,
-                       uint64_t              va_addr,
-                       uint32_t              opcode,
-                       EmulateStateTracker  &state_tracker);
+EventType GetEventType(const IMemoryManager      &mem_manager,
+                       uint32_t                   submit_index,
+                       uint64_t                   va_addr,
+                       uint32_t                   opcode,
+                       const EmulateStateTracker &state_tracker);
 
-std::string GetEventString(const IMemoryManager &mem_manager,
-                           uint32_t              submit_index,
-                           uint64_t              va_addr,
-                           Pm4Type7Header        header,
-                           EmulateStateTracker  &state_tracker);
+std::string GetEventString(const IMemoryManager      &mem_manager,
+                           uint32_t                   submit_index,
+                           uint64_t                   va_addr,
+                           Pm4Type7Header             header,
+                           const EmulateStateTracker &state_tracker);
 uint32_t    GetIndexCount(const IMemoryManager &mem_manager,
                           uint32_t              submit_index,
                           uint64_t              va_addr,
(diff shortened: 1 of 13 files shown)


dive_core/event_state.h
```diff
@@ -8076,6 +8398,312 @@ template<typename CONFIG> class EventStateInfoT
                "attachment";
     }
 
+    //-----------------------------------------------
+    // FIELD ResolveScissor: Defines the rectangular bounds of the resolve operation
+
+    // `ResolveScissorPtr()` returns a shared pointer to an array of `size()` elements
+    inline const VkRect2D* ResolveScissorPtr() const
+    {
+        return reinterpret_cast<VkRect2D*>(reinterpret_cast<uint8_t*>(m_buffer.get()) +
```
hysw: nit: Use `std::byte` instead of `uint8_t`.

It might be worth while to have a 
```cpp
template<typename T>
T* ResolvePtrAt(size_t offset) {
    // Maybe do a pointer bound check for debug?
    return reinterpret_cast<T*>(reinterpret_cast<std::byte*>(m_buffer.get())+offset);
}
```

dive_core/capture_event_info.h
```diff
@@ -48,6 +47,7 @@ enum class EventType : uint8_t
     kWaitForMe,
     kEventWriteStart,
     kEventWriteEnd,
+    kUnknown
```
hysw: Any reason `kUnknown` is not the first? It makes enum zero init a bit less problematic.

dive_core/capture_event_info.h
```diff
@@ -138,6 +138,38 @@ struct EventInfo
 
     RenderModeType m_render_mode = RenderModeType::kUnknown;
     std::string    m_str;
+
+    static bool IsResolve(Util::EventType type)
```
hysw: nit: `constexpr`

dive_core/capture_event_info.h
```diff
@@ -24,6 +24,55 @@
...
+    kSysmemToGmemResolve,
+    kWaitMemWrites,
+    kWaitForIdle,
+    kWaitForMe,
+    kEventWriteStart,
+    kEventWriteEnd,
+    kUnknown
+};
+
+bool IsEvent(const IMemoryManager &mem_manager,
```
wangra-google: should `state_tracker` be `const EmulateStateTracker&` ? same for the `GetEventType` and `GetEventString` funcs

dive_core/data_core.cpp
```diff
@@ -570,6 +532,60 @@ void CaptureMetadataCreator::FillEventStateInfo(EventStateInfo::Iterator event_s
...
+EventStateInfo::Iterator event_state_it)
+{
+    uint32_t rb_resolve_gmem_buffer_base_reg_offset = GetRegOffsetByName(
+    "RB_RESOLVE_GMEM_BUFFER_BASE");
+    if (m_state_tracker.IsRegSet(rb_resolve_gmem_buffer_base_reg_offset))
+    {
+        RB_RESOLVE_GMEM_BUFFER_BASE rb_resolve_gmem_buffer_base_reg;
+        rb_resolve_gmem_buffer_base_reg.u32All = m_state_tracker.GetRegValue(
+        rb_resolve_gmem_buffer_base_reg_offset);
+        event_state_it->SetResolveBaseGmem(rb_resolve_gmem_buffer_base_reg.u32All);
```
wangra-google: just curious, how come the base address is a 32bits value? i thought all address are 64bits

wangra-google: oh, i just read your later change, it is an offset not a real address. Wondering if would be better to call it SetResolveBaseGmemOffset

dive_core/event_state.h
```diff
@@ -8612,6 +9240,51 @@ template<typename CONFIG> class EventStateInfoT
 #define PARTIAL_SIZE_EventStateInfo kUBWCLosslessEnabledOnDSOffset + kUBWCLosslessEnabledOnDSSize
 #undef PARTIAL_INDEX_EventStateInfo
 #define PARTIAL_INDEX_EventStateInfo kUBWCLosslessEnabledOnDSIndex + 1
+    static_assert(alignof(VkRect2D) <= kAlignment,
+                  "Field type aligment requirement cannot exceed kAlignment");
+    static constexpr uint32_t kResolveScissorIndex = PARTIAL_INDEX_EventStateInfo;
+    static constexpr size_t   kResolveScissorOffset = PARTIAL_SIZE_EventStateInfo;
```
wangra-google: do we need to always align up for all offset to be safe?

ui/hover_help_model.cpp
```diff
@@ -427,13 +427,13 @@ void HoverHelp::SetCommandHierarchyNodeItem(const Dive::CommandHierarchy& comman
...
         }
     }
-    else if (node_type == Dive::NodeType::kSyncNode)
-    {
-        hover_msg_sent = true;
-        Dive::SyncType sync_type = command_hierarchy.GetSyncNodeSyncType(node_index);
-        Dive::SyncInfo sync_info = command_hierarchy.GetSyncNodeSyncInfo(node_index);
-        SetCurItem(Item::kSyncNode, (uint32_t)sync_type, sync_info.m_u32All);
-    }
+    //     else if (node_type == Dive::NodeType::kSyncNode)
```
wangra-google: do we still need to keep this?

ui/main_window.cpp
```diff
@@ -3301,17 +3301,24 @@ CorrelationTarget            target)
         return std::nullopt;
     }
 
-    uint64_t       node_index = (uint64_t)(source_index.internalPointer());
-    Dive::NodeType node_type = m_data_core->GetCommandHierarchy().GetNodeType(node_index);
-    bool           type_is_valid = false;
+    uint64_t                      node_index = (uint64_t)(source_index.internalPointer());
```
wangra-google: static_cast just to be consistent with other code :)

dive_core/capture_event_info.h
shanminchao: Done

ui/main_window.cpp
wangra-google: more for consistent with using C++ style than C style. i dont really have a strong opinion on this though, your call :)

hysw: This should be `internalId()`

I don't know why are we using `internalPointer()`

PR #562, opened 2025-11-18
System application capture changes

GrantComm:
Modifications to the capture process so that plugins can be used.

**Manual Testing:**
Linux:
- Open capture dialog, gfxr capture of CubeXr, gfxr capture with plugin, pm4 capture of CubeXr, pm4 capture of CubeXr
- Open capture dialog, pm4 capture of CubeXr, pm4 capture with plugin, gfxr capture of CubeXr, gfxr capture of with plugin

### DIFF
--- a/ui/main_window.h
+++ b/ui/main_window.h
@@ -17,6 +17,13 @@
 #pragma once
 
 #include <QMainWindow>
+#if defined(_MSC_VER)
+#ifdef DIVE_UI_LIB_BUILD
+#pragma message("Compiling main_window.h with DIVE_UI_LIB_BUILD defined (dllexport context)")
+#else
+#pragma message("Compiling main_window.h without DIVE_UI_LIB_BUILD defined (dllimport context)")
+#endif
+#endif
 #include <array>
 #include <functional>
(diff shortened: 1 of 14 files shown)


capture_service/device_mgr.h
```diff
@@ -201,6 +201,8 @@ class DeviceManager
     DeviceManager &operator=(const DeviceManager &) = delete;
     DeviceManager(const DeviceManager &) = delete;
 
+    std::string                     GetPythonPath() const;
```
wangra-google: is there any specific reason why you would prefer exposing those 2 funcs?

ui/main_window.h
```diff
@@ -124,6 +124,9 @@ class MainWindow : public QMainWindow
     explicit MainWindow(ApplicationController &controller);
     ~MainWindow();
     bool LoadFile(const std::string &file_name, bool is_temp_file = false, bool async = true);
+    TraceDialog &GetTraceDialog();
+
+    ProgressTrackerCallback m_progress_tracker;
```
wangra-google: why is this public?

PR #564, opened 2025-11-18
Changes internalPointer() to internalId()

GrantComm:
Removes usage of (uint64_t)index.internalPointer()

**Manual Testing:**
Linux: Capture CubeXr, Click through tree, Search For text, Perform replay from analyze dialog, correlate gfxr and pm4 data.

### DIFF
--- a/ui/command_buffer_model.cpp
+++ b/ui/command_buffer_model.cpp
@@ -90,7 +90,7 @@ QVariant CommandBufferModel::data(const QModelIndex &index, int role) const
             return int(Qt::AlignLeft | Qt::AlignVCenter);
     }
 
-    uint64_t node_index = (uint64_t)(index.internalPointer());
+    uint64_t node_index = index.internalId();
     if (role == Qt::ForegroundRole && IsSelected(node_index))
         return QColor(255, 128, 128);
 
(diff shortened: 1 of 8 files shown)


ui/command_buffer_model.cpp
```diff
@@ -90,7 +90,7 @@ QVariant CommandBufferModel::data(const QModelIndex &index, int role) const
             return int(Qt::AlignLeft | Qt::AlignVCenter);
     }
 
-    uint64_t node_index = (uint64_t)(index.internalPointer());
+    uint64_t node_index = index.internalId();
```
hysw: For further refactoring, probably have
```cpp
inline uint64_t ToNodeIndex(const QModelIndex &index) { return index.internalId(); }
```
defined somewhere, then
```cpp
auto node_index = ToNodeIndex(index);
```
so eventually we can change `node_index` to a opaque type instead of `uint64_t` which could be implicitly casted.

ui/command_buffer_model.cpp
```diff
@@ -191,7 +191,7 @@ QModelIndex CommandBufferModel::index(int row, int column, const QModelIndex &pa
         return createIndex(row, column, (void *)node_index);
```
hysw: Also those `void*` cast are unnecessary.

hysw (approved):

wangra-google (approved):

PR #566, opened 2025-11-19
Move common and utils to src

hysw:
Split off from #486


wangra-google (dismissed):

src/dive/utils/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+    add_library(component_files component_files.h component_files.cpp)
+    target_link_libraries(
+        component_files
+        PUBLIC dive_src_includes absl::statusor
+    )
+endif()
+
+# === version_info =============================================================
+
+add_library(version_info version_info.h version_info.cpp)
```
angela28chen: I think version_info should also be NOT ANDROID because it's not designed to work on android devices

angela28chen (dismissed):

angela28chen (approved):

hysw:
> Please leave a comment describing manual tests

On windows, build and opening trace file fine.

This change only does folder structure change, which affect include path, and should otherwise be a no-op.

PR #577, opened 2025-11-21
Tweak ui element to avoid artifacts.

hysw:
- Add initial parent for two widgets.
  - Make sure qt does not create a window to show them for one frame.
- Reorder tab visibility/enable update.


wangra-google:
what is the bug that you are trying to fix with this pr?

hysw:
> what is the bug that you are trying to fix with this pr?

Small qt rendering issue.

A single frame of pop up window from non-parented widget
image

and

Tab widget have rendering artifacts when the last enabled tab become invisible.

image

PR #579, opened 2025-11-21
Fix missing VK_KHR_surface extension

footballhead:
OpenXR runtimes may not use WSI. Thus, the capture will be missing both VK_KHR_surface and the platform-specific surface extension.

GFXR Vulkan replay is architected to always need a WSI. It will always add the platform surface extension based on the WSI chosen with --wsi (except on Android where it's always VK_KHR_android_surface). Despite all the platform surface extensions depending on VK_KHR_surface, replay doesn't add it. This causes a validation error:

```
VUID-vkCreateInstance-ppEnabledExtensionNames-01388: vkCreateInstance(): pCreateInfo->ppEnabledExtensionNames[0] Missing extension required by the instance extension VK_KHR_android_surface: VK_KHR_surface.
The Vulkan spec states: All required extensions for each extension in the VkInstanceCreateInfo::ppEnabledExtensionNames list must also be present in that list (https://docs.vulkan.org/spec/latest/chapters/initialization.html#VUID-vkCreateInstance-ppEnabledExtensionNames-01388)
```

We can implement DiveVulkanReplayConsumer::Process_vkCreateInstance to add the missing extension before actually calling vkCreateInstance.


wangra-google (approved):

footballhead:
> > Is this a upstream issue that should also be fixed upstream?
> 
> We could probably create an issue, but i feel that they most likely wont have time to look at this. :(

Yes. I believe that this should ultimately be fixed in upstream. I'll create an internal issue just so I don't forget. There are several options. I think that the easiest would be to modify [VulkanReplayConsumerBase::ModifyCreateInstanceInfo](https://github.com/google/dive/blob/main/third_party/gfxreconstruct/framework/decode/vulkan_replay_consumer_base.cpp#L2684) to add the extension:

```c++
    for (const auto& itr : application_->GetWsiContexts())
    {
        // TODO : It's kinda ugly to be referencing Dx12 (even if just by name) in the Vulkan codepath, but
        // having a string associated with the WSI context isn't really something Dx12 has a concept of...this
        // should be able to be refactored away in another PR
        if (gfxrecon::util::platform::StringCompareNoCase(itr.first.c_str(), "Dx12WsiContext"))
        {
            modified_extensions.push_back(itr.first.c_str());
            // hitchens: We could also push_back VK_KHR_surface here
        }
    }
```

But maybe the WSI contexts should also declare dependent extensions. The code here is kind of scattered so it's hard to know the best place to modify. I would need to talk to one of their developers.

PR #590, opened 2025-11-25
Expandable color blend event state.

hysw:
Color blending is currently displayed as a long string.
Make it subitem improves readability.

Also update accent color to make it consistent with other part of UI.

Part of theme and appearance rework (#500).

Before
Screenshot 2025-11-24 223346

After
Screenshot 2025-11-24 224836

### DIFF
--- a/ui/event_state_view.cpp
+++ b/ui/event_state_view.cpp
@@ -20,6 +20,7 @@
 #include <QVBoxLayout>
 #include <map>
 #include <string>
+#include "color_utils.h"
 #include "dive_core/command_hierarchy.h"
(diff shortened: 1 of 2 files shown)


hysw:
> if you check your 1st screenshot, it is not per attachment

It's LogicOp*0* with the attachment number at the end

wangra-google:
> LogicOp_0_

oh, i see. yeah that seems to be a bug. ok, let's keep your change. Can you create a bug for that and assign it to me? Thanks

GrantComm (approved):

PR #592, opened 2025-11-25
Allow Android Studio debugging of replay

footballhead:
wrap.sh requires specific modifications to make the Android Studio debugger work. These modifications come from the Android Developer docs (with one small deviation): https://developer.android.com/ndk/guides/wrap-script

### DIFF
--- a/third_party/gfxreconstruct/android/tools/replay/src/main/resources/lib/arm64-v8a/wrap.sh
+++ b/third_party/gfxreconstruct/android/tools/replay/src/main/resources/lib/arm64-v8a/wrap.sh
@@ -23,6 +23,9 @@ shift
 capture_pm4_enabled=$(getprop debug.dive.replay.capture_pm4)
 os_version=$(getprop ro.build.version.sdk)
 
+# Up to and including SDK 27 (Android 8.1), -Xrunjdwp is the documented approach for using the debugger.
+# SDK 28 (Android 9) added -XjdwpProvider -XjdwpOptions as the preferred way of using the debugger.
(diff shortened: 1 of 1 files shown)


third_party/gfxreconstruct/android/tools/replay/src/main/resources/lib/arm64-v8a/wrap.sh
```diff
@@ -15,11 +15,26 @@
...
+os_version=$(getprop ro.build.version.sdk)
+
+if [ "$os_version" -eq "27" ]; then
+  cmd="$cmd -Xrunjdwp:transport=dt_android_adb,suspend=n,server=y -Xcompiler-option --debuggable $@"
+elif [ "$os_version" -eq "28" ]; then
+  cmd="$cmd -XjdwpProvider:adbconnection -XjdwpOptions:suspend=n,server=y -Xcompiler-option --debuggable $@"
+else
+  # Unlike what the docs claim, need -Xcompiler-option --debuggable here as well
+  cmd="$cmd -XjdwpProvider:adbconnection -XjdwpOptions:suspend=n,server=y -Xcompiler-option --debuggable $@"
+fi
```
footballhead: Since these branches are the same, consider collapsing these into a single statement
```suggestion
else
  # Despite what the Android docs claim, need -Xcompiler-option --debuggable here as well
  cmd="$cmd -XjdwpProvider:adbconnection -XjdwpOptions:suspend=n,server=y -Xcompiler-option --debuggable $@"
fi
```

RenfengLiu (approved):

PR #594, opened 2025-11-25
Add replay option to wait for debugger

footballhead:
`dive_client_cli --wait_for_debugger` will cause replay to launch but pause before android_main for a debugger to attach.

Android Studio needs to be configured correctly with the replay symbols.

This relies on #592 to work correctly.


angela28chen (approved):

capture_service/dive_client_cli.cc
```diff
@@ -244,6 +244,10 @@ ABSL_FLAG(Dive::GfxrReplayOptions,
           "\n\tperf_counters: Collect metrics\n\tgpu_timing: Collect GPU timing\n\trenderdoc: "
           "Create a RenderDoc capture");
 ABSL_FLAG(bool, validation_layer, false, "Run GFXR replay with the Vulkan Validation Layer");
+ABSL_FLAG(bool,
```
footballhead: hold off merging until #553 is merged

PR #600, opened 2025-11-26
Cleanup for only targeting one ANDROID_ABI when building GFXR library

angela28chen:
Since we only build GFXR library for one ABI at a time, change the code to reflect that build process:
- Instead of asking the user to specify ABI with a flag when using `dive_client_cli`, just verify that the device ABI is same as the library's ABI
- Only put the library file `libVkLayer_gfxreconstruct.so` into `install/` and not a whole `gfxr_layer` folder
- No more special `ResolveAndroidLibPath()` search logic for `libVkLayer_gfxreconstruct.so`

### DIFF
--- a/src/dive/utils/version_info_test.cpp
+++ b/src/dive/utils/version_info_test.cpp
@@ -28,8 +28,8 @@ namespace
 
 using ::absl_testing::IsOkAndHolds;
 using ::absl_testing::StatusIs;
-using testing::ElementsAre;
 using testing::Pair;
+using testing::UnorderedElementsAre;
 
 TEST(GetDeviceLibraryVersionMapTest, BasicPass)
 {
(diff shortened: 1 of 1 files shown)


src/dive/utils/version_info_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+namespace
+{
+
+using ::absl_testing::IsOkAndHolds;
+using ::absl_testing::StatusIs;
+
+MATCHER_P(DeviceLibrariesMapEq, expected, "")
+{
+    EXPECT_EQ(arg.size(), expected.size());
```
wangra-google: if the arg.size() != expected.size(), at least one of the following tests could crash?

angela28chen: no crash, ExtraKeyFail testcase is for this

angela28chen: oh nvm I see what you mean, I'll make it a fatal eq

elviscapiaq (approved):
Thanks for doing this code change. Could you state what tests have you performed especially for dive_client_cli. Other than that, looks good to me.

src/dive/utils/version_info_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+TEST(GetDeviceLibraryVersionMapTest, BasicPass)
+{
+    std::string csv_content = "sha,0343dcaa2d0335a46dd673d19b56963d9b045879\n"
+                              "version,1.1.2\nbuild_type,Debug\nrelease_type,dev\nabi,arm64-v8a";
+
+    absl::StatusOr<std::map<std::string_view, std::string_view>> got = GetDeviceLibraryVersionMap(
+    csv_content);
+
+    EXPECT_THAT(got,
+                IsOkAndHolds(ElementsAre(Pair(VersionInfoConstants::kNameAbi, "arm64-v8a"),
```
wangra-google: nit: do we want to check order? if not maybe UnorderedElementsAre is better

wangra-google (approved):

src/dive/utils/version_info_test.cpp
angela28chen: ok fixed!

PR #602, opened 2025-11-26
Fix memory leak, and many lint warning.

hysw:
Fix two memory leak, and many lint warning.
    
- Created a tiny Guidelines Support Library
- Add some helper function to make sure we do qt ownership correctly.

### DIFF
--- a/src/dive/lint/gsl.h
+++ b/src/dive/lint/gsl.h
@@ -1,0 +1,30 @@
+/*
+ Copyright 2025 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+#pragma once
+
+// Guidelines Support Library
+// For .clang-tidy
+
+#include <type_traits>
+
+namespace gsl
+{
+
+template <class T, std::enable_if_t<std::is_pointer_v<T>, bool> = true>
+using owner = T;
+
+}
--- a/src/dive/ui/forward.h
+++ b/src/dive/ui/forward.h
@@ -23,11 +23,18 @@ class QAbstractProxyModel;
 class QCheckBox;
 class QComboBox;
 class QGroupBox;
+class QHBoxLayout;
 class QLabel;
+class QLineEdit;
+class QListWidget;
 class QPushButton;
 class QScrollArea;
 class QSortFilterProxyModel;
+class QSpinBox;
+class QStandardItemModel;
 class QTabWidget;
+class QTextEdit;
+class QVBoxLayout;
 
 namespace Dive
 {
@@ -62,6 +69,7 @@ class GfxrVulkanCommandModel;
 class GpuTimingModel;
 class GpuTimingTabView;
 class HoverHelp;
+class MainWindow;
 class OverlayHelper;
 class OverviewTabView;
 class PerfCounterModel;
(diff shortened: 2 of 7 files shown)


src/dive/lint/gsl.h
```diff
@@ -0,0 +1,31 @@
...
+
+// Guidelines Support Library
+// For .clang-tidy
+
+#include <type_traits>
+
+namespace gsl
+{
+
+template<class T, std::enable_if_t<std::is_pointer_v<T>, bool> = true>  //
```
footballhead: why the trailing //?

src/dive/ui/forward.h
```diff
@@ -18,16 +18,24 @@
 
 // Forward declarations for UI.
 
+class OverlayHelper;
```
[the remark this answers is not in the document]
hysw: Eh, I probably selected the wrong range when running sort+deduplicate in the editor.

hysw:
> Thanks for making this effort to make Dive ui code use "new" in a more intentioned way

Had a bit offline chat earlier with @footballhead , instead of disable lint check for pointer ownership, it seems sensible to explicitly annotate allocations.

ui/analyze_window.cpp
```diff
@@ -170,51 +173,47 @@ AnalyzeDialog::AnalyzeDialog(ApplicationController        &controller,
...
-    m_delete_replay_artifacts_layout = new QHBoxLayout();
-    m_delete_replay_artifacts_button = new QPushButton("&Delete Previous Replay Artifacts", this);
-    m_delete_replay_artifacts_layout->addWidget(m_delete_replay_artifacts_button);
+    m_delete_replay_artifacts_layout = QtNewUnowned<QHBoxLayout>();
+    m_delete_replay_artifacts_button = LayoutHelper{ m_delete_replay_artifacts_layout }
+                                       .New<QPushButton>("&Delete Previous Replay Artifacts", this);
 
     // Left Panel Layout
-    m_left_panel_layout = new QVBoxLayout();
+    m_left_panel_layout = QtNewUnowned<QVBoxLayout>();
```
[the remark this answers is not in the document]
hysw: Yes. Although that's a lot of reordering probably fitted for the next PR.

I intentionally did not add `AddWidget` and `AddLayout` since they encourage having unparented elements.

src/dive/ui/layout_helper.h
```diff
@@ -0,0 +1,92 @@
...
+#include "dive/ui/lint.h"
+
+// Examples
+//
+// Setup layout for a widget:
+//   auto layout = NewWidgetLayout<QHBoxLayout>(widget);
+//   auto* label = layout.New<QLabel>(tr("text"));
+//
+// Add a new item to an existing layout:
+//   auto* label = NewWidgetLayout{layout}.New<QLabel>(tr("text"));
```
footballhead: nit: since NewWidgetLayout is a funtion call, should the `{}` be `()` instead?
```suggestion
//   auto* label = NewWidgetLayout(layout).New<QLabel>(tr("text"));
```

src/dive/ui/layout_helper.h
```diff
@@ -0,0 +1,92 @@
...
+        {
+            return NewWidget<T, Args...>(std::forward<Args>(args)...);
+        }
+    }
+
+private:
+    LayoutT* absl_nonnull m_layout;
+};
+
+template<typename T, typename... Args>  //
```
footballhead: If we prefer AlwaysBreakTemplateDeclarations, don't need the trailing `//`
```suggestion
template<typename T, typename... Args>
```

src/dive/ui/layout_helper.h
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    template<typename T, typename... Args>  //
+    LayoutHelper<LayoutHelper> NewLayout(Args&&... args)
+    {
+        static_assert((std::is_convertible_v<T*, QWidget*>));
+        auto layout = DiveLint::QtOwned(new T(std::forward<Args>(args)...));
+        m_layout->addLayout(layout);
+        return LayoutHelper<LayoutHelper>{ layout };
+    }
```
footballhead: Shouldn't the template param be `T` instead of `LayoutHelper`? Otherwise `.NewLayout().NewLayout()` shouldn't work properly.
```suggestion
    template<typename T, typename... Args>  //
    LayoutHelper<T> NewLayout(Args&&... args)
    {
        static_assert((std::is_convertible_v<T*, QWidget*>));
        auto layout = DiveLint::QtOwned(new T(std::forward<Args>(args)...));
        m_layout->addLayout(layout);
        return LayoutHelper<T>{ layout };
    }
```

PR #605, opened 2025-11-27
Fix dive_client_cli lint error.

hysw:
- Remove unused includes.
- Fix almost all clang-tidy issues.
- Remaining cppcoreguidelines-pro-bounds-array-to-pointer-decay need to be fixed in constants.h
- Rewrite array initialization.

### DIFF
--- a/capture_service/dive_client_cli.cc
+++ b/capture_service/dive_client_cli.cc
@@ -15,23 +15,18 @@ limitations under the License.
 */
 
 #include <filesystem>
-#include <future>
 #include <iostream>
 #include <ostream>
 #include <string>
-#include <system_error>
 #include <thread>
 
-#include "absl/base/no_destructor.h"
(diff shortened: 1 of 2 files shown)


capture_service/dive_client_cli.cc
```diff
@@ -387,10 +385,14 @@ absl::Status TriggerPm4Capture(Dive::DeviceManager& mgr, const std::string& down
         return absl::UnavailableError("Connection failed: " + std::string(status.message()));
     }
 
-    absl::StatusOr<std::string> capture_file_path = client.StartPm4Capture();
-    if (!capture_file_path.ok())
+    std::string capture_file_path;
```
elviscapiaq: Absel recommends to treat an StatusOr as a pointer. That means, we should dereference this to get its value (*capture_file_path). Ref: https://abseil.io/tips/181

PR #609, opened 2025-11-27
Debian packaging

hysw:


scripts/package_linux.sh
```diff
@@ -0,0 +1,56 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+readonly DIVE_ROOT="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
+
+mkdir -p build/package_linux/device
+mkdir -p build/package_linux/host
+# /home/runner/work/dive/Qt
```
hysw:
```suggestion

```

wangra-google (dismissed):

CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ if(MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
 endif()
 set(THIRDPARTY_DIRECTORY "${CMAKE_SOURCE_DIR}/third_party")
+# TODO: remove set(CMAKE_INSTALL_PREFIX ...) once we fix search path and mac build.
```
footballhead: nit prefer TODOs with a bug. It lets us track the problem and provides more context for future readers who can determine if the TODO is stale or not

packaging/linux/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+set(CPACK_DEBIAN_PACKAGE_NAME "dive-profiler")
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER
+    "Dive Profiler Developers <dive-dev-team@google.com>"
+)
+set(CPACK_DEBIAN_PACKAGE_DESCRIPTION "Dive is a GPU Profiler.")
+set(CPACK_DEBIAN_PACKAGE_SECTION "devel")
+set(CPACK_DEBIAN_PACKAGE_PRIORITY "optional")
+set(CPACK_DEBIAN_PACKAGE_DEPENDS "")
+set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON)
+set(CPACK_PACKAGING_INSTALL_PREFIX "/opt/dive-profiler")
```
footballhead: Out of curiosity, any reason to prefer /opt over the normal /usr?

scripts/package_linux.sh
```diff
@@ -0,0 +1,56 @@
...
+pushd build/package_linux/device
+    cmake  -DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK_HOME}/build/cmake/android.toolchain.cmake \
```
footballhead: DRY. consider invoking build_android.sh instead to reduce the number of copies and improve maintainability

scripts/package_linux.sh
```diff
@@ -0,0 +1,56 @@
...
+        ${DIVE_ROOT} || exit 1
+    cmake --build . --config=Debug -j || exit 1
+    cmake --install . --prefix install || exit 1
+popd
+
+pushd build/package_linux/host
+    cmake -G "Ninja" \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DDIVE_ANDROID_PREBUILT="`pwd`/../device/install" \
+        -DCMAKE_C_COMPILER:FILEPATH=/usr/bin/clang-19 \
```
footballhead: any reason to prefer clang-19?

scripts/package_linux.sh
```diff
@@ -0,0 +1,56 @@
...
+        -DCMAKE_CXX_COMPILER:FILEPATH=/usr/bin/clang++-19 \
+        ${DIVE_ROOT} || exit 1
+    cmake --build . --config Release --target all || exit 1
```
footballhead: nit: `--target all` is the default so specifying it is redundant

```suggestion
    cmake --build . --config Release || exit 1
```

.github/workflows/presubmit.yml
```diff
@@ -452,3 +452,57 @@ jobs:
...
+      - name: Install dependency
+        run: |
+          sudo apt-get install --yes \
+          cmake clang-19 \
+          libsystemd-dev \
+          libbsd-dev \
+          ninja-build \
+          python3-mako \
+          libxcb-glx0-dev
+          which clang-19
```
footballhead: nit: All `apt-get install` invocations  could be collapsed into one line.

CMakeLists.txt
hysw: That's from #652

scripts/package_linux.sh
hysw: `build_android.sh` script is quite a bit different than the scripts here. They overlap in CMake options, and that's about it.

We probably should have a cmake preset.

.github/workflows/presubmit.yml
```diff
@@ -452,3 +452,43 @@ jobs:
...
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Set up Java 17
+        uses: actions/setup-java@v3
+        id: setup-java
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+      - name: Check out code
+        uses: actions/checkout@v3
```
hysw: Update github action pinned version.

.github/workflows/presubmit.yml
```diff
@@ -452,3 +452,62 @@ jobs:
       - name: Ninja log
         run: |
           cat build/.ninja_log
+      - name: Produce install artifacts
+        run: |
+          cmake --install build --prefix build/install_artifact --config Release
```
footballhead: nit: since we're not using the multi-config genreator, do we need --config? Also, this is a debug build
```suggestion
          cmake --install build --prefix build/install_artifact
```

.github/workflows/presubmit.yml
```diff
@@ -452,3 +452,62 @@ jobs:
...
+      - name: Download Android Artifacts
+        uses: actions/download-artifact@v5
+        with:
+          name: install_artifact_android
+          path: build/package_linux/dive_android
+      - name: Android Artifacts
+        run: ls -R build/package_linux/dive_android
+      - name: Build
+        run: |
+          mkdir -p build/package_linux/
```
footballhead: nit: since you downloaded `build/package_linux`, do you need to mkdir?
```suggestion
```

hysw:
> a more general question, we are currently saving capture files, or even crash dump file at the same folder as executables, would that be a problem with your change since the user may not have write permission with the executable install folder?

That's a related but separate question. We should follow platform idiom for those, e.g. probably `${HOME}/.dive-profiler` on Linux.

.github/workflows/presubmit.yml
hysw: Need to be updated according to #679

PR #613, opened 2025-11-28
Pm4 download fix

RenfengLiu:
as the dive is in the app bundle, the default "." is not writable. this added a input box to let user specify the download path, and default to user's home dir

### DIFF
--- a/third_party/gfxreconstruct/android/tools/replay/CMakeLists.txt
+++ b/third_party/gfxreconstruct/android/tools/replay/CMakeLists.txt
@@ -31,20 +31,19 @@ add_library(gfxrecon-replay
                 ${GFXRECON_SOURCE_DIR}/tools/replay/replay_pre_processing.h
                 ${GFXRECON_SOURCE_DIR}/tools/replay/android_main.cpp)
 
-# GOOGLE: Add dive root folder
 target_include_directories(gfxrecon-replay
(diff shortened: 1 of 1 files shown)


ui/trace_window.cpp
```diff
@@ -219,6 +219,15 @@ TraceDialog::TraceDialog(ApplicationController &controller, QWidget *parent) :
     m_gfxr_capture_file_local_directory_label->hide();
     m_gfxr_capture_file_local_directory_input_box->hide();
 
+    m_pm4_capture_file_local_directory_layout = new QHBoxLayout();
```
wangra-google: what is the current workflow for pm4 captures? are they saved to where executable is?

RenfengLiu: yes, it was hardcoded as ` std::string           download_path = ".";`

wangra-google: i am wondering if it is better to have the same custom path for gfxr and pm4 so that we don't have too much options in the UI. do you see any potential issue if the pm4 capture and gfxr capture uses the same folder? the pm4 capture directly from the app would have a different name anyways.

RenfengLiu: yes, we can, but I'm not sure how to do that, currently the capture save path is only available for gfxr capture. @GrantComm maybe knows more

PR #615, opened 2025-12-01
Fix replay without --device

footballhead:
Running replay without --device was failing:

```
$ dive_client_cli --command gfxr_replay --gfxr_replay_file_path /sdcard/Download/com.google.bigwheels.project_cube_xr.debug_trim_trigger_20251027T095805.gfxr

Error executing command 'gfxr_replay': Failed to deploy replay apk: Command `/usr/local/google/home/hitchens/.local/share/virtualenvs/dive-v1dp7pER/bin/python3 /usr/local/google/home/hitchens/git/dive/install/gfxrecon.py install-apk /usr/local/google/home/hitchens/git/dive/install/gfxr-replay.apk -s ` failed with return code 512, error: usage: gfxrecon.py install-apk [-h] [-s DEVICE_ID] file
gfxrecon.py install-apk: error: argument -s/--select: expected one argument
```

`--device` is an optional parameter so it may be empty. If it is, I expect an adb device to be chosen.

`--device` is stored in `opts.serial`. While `GetTargetDevice()` was returning a valid device, `opts.serial` wasn't updated. However, `opts.serial` is required for `DeviceManager::DeployReplayApk()`.

This change ensures that `opts.serial` is set to a valid serial.

### DIFF
--- a/capture_service/dive_client_cli.cc
+++ b/capture_service/dive_client_cli.cc
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+#include <algorithm>
 #include <filesystem>
 #include <future>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


capture_service/dive_client_cli.cc
```diff
@@ -40,6 +41,12 @@ limitations under the License.
 #include "network/tcp_client.h"
 #include "utils/version_info.h"
 
+namespace
```
footballhead: FYI I included more things in the anonymous namespace because GlobalOptions, etc doesn't require external linkage

hysw (dismissed):

PR #616, opened 2025-12-01
Force screen on when device connected

RenfengLiu:


RenfengLiu:
Looks like this is still not working,

PR #617, opened 2025-12-02
Default the capture type to gfxr capture instead of PM4

RenfengLiu:


wangra-google (approved):

wangra-google:
Thanks for making this change, i was about to do the same haha

hysw (approved):

PR #618, opened 2025-12-02
Enable showing close button for the Trace dialog

RenfengLiu:


RenfengLiu:
From
image

TO
image

ui/main_window.cpp
```diff
@@ -565,7 +565,7 @@ MainWindow::MainWindow(ApplicationController &controller) :
     // Retrieve the available metrics
     LoadAvailableMetrics();
 
-    m_trace_dig = new TraceDialog(m_controller, this);
+    m_trace_dig = new TraceDialog(m_controller, nullptr);
```
hysw: Any reason it is unparented?

RenfengLiu: looks like mac only allow parent window to have the close button

hysw (dismissed):

wangra-google (approved):

PR #619, opened 2025-12-02
Moves capture workers out of trace window

GrantComm:
Moves capture workers outside of the trace window class. Follow up pr utilizes this when allowing unique local save directory with both gfxr and pm4 captures

**Manual Testing:**
Linux:
Open capture window, perform gfxr capture, perform pm4 capture

### DIFF
--- a/ui/capture_worker.h
+++ b/ui/capture_worker.h
@@ -37,10 +37,9 @@ public:
     void SetTargetCaptureDir(const std::string &target_capture_dir);
 signals:
     void CaptureAvailable(const QString &);
-    void DownloadedSize(int64_t size, int64_t total_size);
+    void DownloadedSize(qlonglong size, qlonglong total_size);
     void ErrorMessage(const QString &err_msg);
 
 protected:
     std::filesystem::path m_target_capture_dir;
-    int64_t               m_capture_size;
 };
--- a/ui/gfxr_capture_worker.cpp
+++ b/ui/gfxr_capture_worker.cpp
@@ -59,7 +59,8 @@ const std::map<std::string, std::string> &previous_timestamps)
 }
 
 //--------------------------------------------------------------------------------------------------
-absl::StatusOr<int64_t> GfxrCaptureWorker::getGfxrCaptureDirectorySize(Dive::AndroidDevice *device)
+absl::StatusOr<qlonglong> GfxrCaptureWorker::getGfxrCaptureDirectorySize(
+Dive::AndroidDevice *device)
 {
     // Retrieve the names of the files in the capture directory on the device.
     std::string                 ls_command = "shell ls " + m_source_capture_dir;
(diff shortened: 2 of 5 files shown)


ui/capture_worker.h
```diff
@@ -0,0 +1,79 @@
...
+    bool             m_gfxr_capture;
+    int64_t          m_downloaded_size;
+};
+
+class CaptureWorker : public QThread
+{
+    Q_OBJECT
+
+public:
+    CaptureWorker(QProgressDialog *pd) :
```
hysw: Usually QObject has it's parent as parameter, having a QProgressDialog here could be confusing.

GrantComm: Done. Removed the progress dialog and set the parent as the parameter

ui/capture_worker.h
```diff
@@ -0,0 +1,79 @@
...
+ limitations under the License.
+*/
+
+#pragma once
+
+#include <QThread>
+#include <QProgressDialog>
+#include <filesystem>
+
+class ProgressBarWorker : public QThread
```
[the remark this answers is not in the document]
GrantComm: Done, removed.

ui/gfxr_capture_worker.cpp
```diff
@@ -0,0 +1,247 @@
...
+            std::string get_file_size_command = "shell stat -c %s " + path;
+
+            // Get the timestamp for last time the file was updated.
+            std::string get_file_update_timestamp_command = "shell stat -c %Y " + path;
+
+            absl::StatusOr<std::string> str_num = device->Adb().RunAndGetResult(
+            get_file_size_command);
+            absl::StatusOr<std::string> file_update_timestamp = device->Adb().RunAndGetResult(
+            get_file_update_timestamp_command);
+            int64_t num = std::stoll(str_num->c_str());
```
wangra-google: maybe better use absl::SimpleAtoi to avoid handling exception.

ui/gfxr_capture_worker.cpp
```diff
@@ -0,0 +1,247 @@
...
+            // If a file size is 0, then the file has finished being written to yet. Sleep and
+            // restart the size calculation.
+            if (num == 0)
+            {
+                QThread::msleep(10);
```
wangra-google: is it possible for some reason the file is not able to be written? wondering if we should have a timeout counter.

ui/gfxr_capture_worker.cpp
GrantComm: Added

ui/gfxr_capture_worker.cpp
```diff
@@ -81,56 +82,111 @@ absl::StatusOr<int64_t> GfxrCaptureWorker::getGfxrCaptureDirectorySize(Dive::And
...
-            // Add the timestamp for the last time the file was udpated.
-            current_timestamps[file] = file_update_timestamp->data();
+            // Add the timestamp and update the total size.
+            current_timestamps[file_name] = file_timestamp;
+            total_size += file_size;
+        }
 
-            // Update the total size of the gfxr capture directory.
-            size += std::stoll(str_num->c_str());
+        // If a file size is 0, then the file has finished being written to yet. Sleep and
```
hysw: This line of comment is confusing.

ui/gfxr_capture_worker.cpp
wangra-google: resolving this since we are going with the adaptive timeout solution that Peter suggested.

PR #635, opened 2025-12-05
CI refactor by gemini

RenfengLiu:
This enable build cache for Macos/Linux/Android/Windows build
Now the cache will only be generated from main and be used in each PR


hysw:
We may want to reduce the combination a bit on the CI and have a continuous build with all the combinations.

PR #638, opened 2025-12-05
 Adds Crashpad as a submodule from Backtrace Labs fork

elviscapiaq:
- The original Google Crashpad repository does not support CMake, relying instead on Gn for building. To integrate easily, this   commit uses the well-maintained Backtrace Labs fork (Apache 2.0 license).
- A dive_crashpad library is defined to use Crashpad.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -14,6 +14,7 @@ Checks:
   - "-cppcoreguidelines-avoid-non-const-global-variables" # ABSL_FLAG
   - "-cppcoreguidelines-use-enum-class"
   - "-readability-redundant-access-specifiers" # qt
+  - "-abseil-unchecked-statusor-access" # crashes clangd
 
 CheckOptions:
   - key: readability-implicit-bool-conversion.AllowPointerConditions
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Examples:
 - Install the dependencies on device and start the package and do a capture after the applications runs 5 seconds.
     ```sh
     # On Linux
-    ./dive_client_cli --command capture --package YOUR.PACKAGE.NAME.HERE --type vulkan_openxr --trigger_capture_after 5 --download_dir LOCAL/DIR/TO/SAVE/CAPTURES
+    ./dive_client_cli --command capture --package YOUR.PACKAGE.NAME.HERE --type vulkan_openxr --trigger_capture_after 5s --download_dir LOCAL/DIR/TO/SAVE/CAPTURES
     ```
 
 - Install the dependencies on device and start the package
--- a/capture_service/dive_client_cli.cc
+++ b/capture_service/dive_client_cli.cc
@@ -63,7 +63,7 @@ struct GlobalOptions
     Dive::AppType app_type;
     std::string download_dir;
     std::string gfxr_capture_file_dir;
-    int trigger_capture_after;
+    absl::Duration trigger_capture_after;
 
     Dive::GfxrReplaySettings replay_settings;
 };
@@ -143,6 +143,27 @@ absl::Status ValidateRunOptions(const GlobalOptions& options)
     return absl::OkStatus();
 }
 
+// Validates options for --command pm4_capture
+absl::Status ValidatePm4CaptureOptions(const GlobalOptions& options)
+{
+    RETURN_IF_ERROR(ValidateRunOptions(options));
+
+    if (options.trigger_capture_after < absl::ZeroDuration())
+    {
+        return Dive::InvalidArgumentError("--trigger_capture_after must be non-negative");
+    }
+
+    if (options.trigger_capture_after == absl::InfiniteDuration() ||
+        options.trigger_capture_after == -absl::InfiniteDuration())
+
+    {
+        return Dive::InvalidArgumentError(absl::StrCat(
+            "Unsupported --trigger_capture_after value: ", options.trigger_capture_after));
+    }
+
+    return absl::OkStatus();
+}
+
 // Helper to validate options for GFXR replay.
 absl::Status ValidateGfxrReplayOptions(const GlobalOptions& options)
 {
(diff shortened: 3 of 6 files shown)


ui/CMakeLists.txt
```diff
@@ -380,3 +381,22 @@ add_dependencies(${PROJECT_NAME} notice_file)
...
+        ${CMAKE_COMMAND} -E copy_if_different $<TARGET_FILE:handler>
+        $<TARGET_FILE_DIR:${PROJECT_NAME}>
+    COMMENT "Copying Crashpad handler to output directory"
+)
+
+set(APP_BUNDLE_NAME "Dive.app")
+if(APPLE)
+    install(
+        PROGRAMS $<TARGET_FILE:handler>
+        DESTINATION "${APP_BUNDLE_NAME}/Contents/Helpers"
```
elviscapiaq: I am not sure if this is the correct path for macOS. @angela28chen , Could you help to determine if this is the correct path?

ui/main.cpp
```diff
@@ -69,6 +90,71 @@ ABSL_RETIRED_FLAG(bool, reverse, false, "Qt flag reverse");
...
+                                        /*asynchronous_start=*/false);
+
+    if (!success)
+    {
+        return absl::InternalError("Failed to start Crashpad handler.");
+    }
+    return absl::OkStatus();
+}
+
 class CrashHandler
```
elviscapiaq: I am preparing a second commit that is going to replace this absel CrashHandler implementation with Crashpad. Also, in this second commit, I am going to set up cmake to automatically upload debug symbols to our Crashpad server, so we can see the crashes sent by users in a human-readable format.

ui/main.cpp
```diff
@@ -69,6 +90,71 @@ ABSL_RETIRED_FLAG(bool, reverse, false, "Qt flag reverse");
...
 //--------------------------------------------------------------------------------------------------
+absl::Status InitializeCrashpad()
+{
+    absl::StatusOr<std::filesystem::path> exe_dir = Dive::GetExecutableDirectory();
+    if (!exe_dir.ok())
+    {
+        return exe_dir.status();
+    }
+
+    std::filesystem::path handler_path = *exe_dir / kHandlerBinary;
```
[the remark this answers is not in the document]
elviscapiaq: I fixed the path where the `crashpad_handler` goes, and it is under `dive.app/Contents/MacOS` which is the same as the dive binary, so we should not have any issue.

ui/CMakeLists.txt
elviscapiaq: I updated the path to `dive.app/Contents/MacOS` where is the path the DIve binary goes on MacOS based on my investigation. Would like to confirm if everything is working for MacOS. @angela28chen @GrantComm once you have time, could you help me to test this PR for macOS, mainly I am just guessing based on the directory structure for macOS we have.

ui/CMakeLists.txt
```diff
@@ -338,6 +338,8 @@ if(APPLE)
             "
         message(STATUS \"Running macdeployqt...\")
         execute_process(COMMAND \"macdeployqt\" \"${CMAKE_INSTALL_PREFIX}/dive.app\")
+        message(STATUS \"Copying Crashpad handler to app bundle\")
+        execute_process(COMMAND \"${CMAKE_COMMAND}\" -E copy \"${CMAKE_INSTALL_PREFIX}/$<TARGET_FILE_NAME:handler>\" \"${CMAKE_INSTALL_PREFIX}/dive.app/Contents/MacOS/\")
```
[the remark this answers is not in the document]
wangra-google: i am confused since you also did another copy in line 391, are those 2 copies doing the same thing?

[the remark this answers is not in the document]
elviscapiaq: I am copying in the line 391 for development purposes, so we can execute the dive binary without installing anything, whereas in this part I am copying for installing purposes, that is, we use `cmake --install build` to put everything in the./install folder (including crashpad_handler) and then we start the packaging process from that.

angela28chen:
Do we want crashpad for only dive UI or also for CLIs?

ui/CMakeLists.txt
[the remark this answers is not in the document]
GrantComm: Confirmed the crashpad handler is initialized and working on MacOS with and without the install.

wangra-google (approved):

ui/main.cpp
```diff
@@ -69,6 +90,125 @@ ABSL_RETIRED_FLAG(bool, reverse, false, "Qt flag reverse");
...
+    {
+        database->GetSettings()->SetUploadsEnabled(true);
+    }
+    else
+    {
+        return absl::InternalError("Failed to initialize Crashpad database.");
+    }
+
+    std::string version = Dive::GetHostShortVersionString();
+    if (version.size() > kMaxCrashpadVersionLength)
```
MichaelAnttila: Given that we are already close to Crashpad's max version length, I feel like we should be more robust here.  Maybe we should add a "GetHostShortVersionString(int maxLength)" or "GetHostExtremelyShortVersionString()" that guarantees that the version will be less than the max length.  I would hate to have crash logging stop working just because someone accidentally made the short version string too long.

This can be done in a subsequent PR.  I just wanted to note it here.

src/dive/crashpad/crashpad_utils.h
```diff
@@ -0,0 +1,41 @@
...
+constexpr char kNoRateLimitFlag[] = "--no-rate-limit";
+constexpr int kMaxCrashpadVersionLength = 30;
+
+// Returns the platform-specific writable directory for application data.
+// Windows: %LOCALAPPDATA%/Dive
+// macOS: ~/Library/Application Support/Dive
+// Linux: ~/.local/share/Dive
+absl::StatusOr<std::filesystem::path> GetWritableRoot();
+
+}  // namespace Dive
```
hysw: nit: newline at the end of files (all files in dive/crashpad seems to miss new lines)

src/dive/crashpad/crashpad_utils.h
```diff
@@ -0,0 +1,41 @@
...
+#pragma once
+
+#include <filesystem>
+
+#include "absl/status/statusor.h"
+
+namespace Dive
+{
+
+constexpr char kHandlerBinary[] = "crashpad_handler";
```
angela28chen: Perhaps these should be bundled together in a struct so that we can avoid collisions for other constants in Dive namespace? Something like `kMetricsDirectory` seems too generic otherwise.

angela28chen (commented):
Thanks for making a separate library for crashpad stuff to start. Perhaps this PR shouldn't be touching files under `ui/` anymore?

wangra-google (dismissed):

src/dive/crashpad/crashpad_utils.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <filesystem>
+
+#include "absl/status/statusor.h"
+
+namespace Dive
+{
+
+inline constexpr char kProductName[] = "Dive";
+inline constexpr char kHandlerBinary[] = "crashpad_handler";
```
angela28chen: nit: Suggesting `kCrashHandlerBinary` since there's a lot of variables under Dive namespace I think it's worth being more specific

angela28chen (approved):

PR #640, opened 2025-12-08
[PoC] Restructure build and install folders to make it more similar across different platforms, build types, and release types

angela28chen:
Changes:
- Switch from Ninja to Ninja Multi-Config so that the build folder structure on Mac/Linux parallels Windows better (since VS is a multi-config generator)
- Rather than `DIVE_ROOT/install`, use `DIVE_ROOT/build/pkg/<host|device>` for dev builds and similar `TOP_LEVEL/build/pkg/<host|device>` for release binaries. 
- The provided utility for locating the files is `Dive::ResolveDevicePath()`. Refactor other areas where the same logic is being performed to use this instead


angela28chen:
Going to attempt to break this into smaller PRs for easier reviews

PR #644, opened 2025-12-08
Resolves install path +  moving command_utils

elviscapiaq:
- This change resolves the Dive install path issue. In addition to the existing dive_install_dir, it adds support for relative paths `(./install` and `.`) relative to the Dive UI binary location. This should cover all necessary cases for resolving paths for Android libraries and artifacts.

- The change was applied for both: Dive UI and Dive CLI.

- Also, it moves the commands utils from ./capture_service to ./dive/utils.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,9 +22,6 @@ third_party/freedreno/obj
 g_*.h
 g_*.cpp
 
-.ipynb_checkpoints/
-local_notebooks/*.ipynb
-
 .env
 
 install/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,6 +248,10 @@ if(ANDROID)
         DESTINATION "${CMAKE_BINARY_DIR}/gfxr_layer"
     )
     install(
+        CODE
+            "file(REMOVE \"\${CMAKE_INSTALL_PREFIX}/libVkLayer_gfxreconstruct.so\")"
+    )
+    install(
         FILES
             ${CMAKE_BINARY_DIR}/gfxr_layer/jni/${ANDROID_ABI}/libVkLayer_gfxreconstruct.so
         DESTINATION ${CMAKE_INSTALL_PREFIX}
--- a/dive_core/capture_event_info.cpp
+++ b/dive_core/capture_event_info.cpp
@@ -491,4 +491,51 @@ uint32_t Util::GetIndexCount(const IMemoryManager &mem_manager,
     return index_count;
 }
 
+//--------------------------------------------------------------------------------------------------
+std::optional<VkPrimitiveTopology> Util::GetTopology(const IMemoryManager &mem_manager,
+                                                     uint32_t              submit_index,
+                                                     uint64_t              va_addr,
+                                                     Pm4Type7Header        header)
+{
+    // Get primitive-type. The primitive-type is always the least-significant 6-bits of the
+    // given word
+
+    // For all draw call PM4 packets other than the INDX variety, the relevant DWORD is
+    // the first dword after the header
+    uint64_t prim_dword_addr = va_addr + sizeof(header);
+    if (header.opcode == CP_DRAW_INDX)
+        prim_dword_addr += sizeof(uint32_t);  // 2nd dword
+
+    uint32_t prim_dword;
+    DIVE_VERIFY(
+    mem_manager.RetrieveMemoryData(&prim_dword, submit_index, prim_dword_addr, sizeof(prim_dword)));
+    pc_di_primtype gpu_type = static_cast<pc_di_primtype>(prim_dword & 0x3f);
+    switch (gpu_type)
+    {
+    case DI_PT_LINELIST:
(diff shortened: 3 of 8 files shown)


hysw:
> If it is for Release, then the path considered are ./install and . relative to the Dive UI binary location. For Debug builds, we keep the original mechanism to find the Dive install path.

Having divergent code path is probably not a good idea.

hysw:
> Also, it moves the commands utils from ./capture_service to ./dive/utils.

If it's OS dependent, it's probably more suitable for `//src/dive/os/...`

elviscapiaq:
> > If it is for Release, then the path considered are ./install and . relative to the Dive UI binary location. For Debug builds, we keep the original mechanism to find the Dive install path.
> 
> Having divergent code path is probably not a good idea.

That is a good point. I think I was overkilling the solution. I simplified the current PR and no more divergent code path.

elviscapiaq:
> > I tested on Linux for both: Dive CLI and Dive UI, however, I am missing the tests for WIndows. I woudl appreciate if someone can confirm, otherwise I can do tomorrow.
> 
> i can test windows for you. anything specific you want me to verify?

Thanks. Could you test for release binaries? You should not see any error for pm4/gfxr capture. Also, you should not need to use `windeploy_qt` for it, but make sure that the `DIVE_DEVICE_LIBRARIES_VERSION_FILENAME` and `kProfilingPluginShaPath` are located at `.` or `./install` relative paths.

src/dive/os/command_utils_win32.cc
```diff
@@ -14,13 +14,14 @@ See the License for the specific language governing permissions and
...
 
-#include "command_utils.h"
-
 #include <vector>
+
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
 #include "absl/strings/ascii.h"
-#include "common/log.h"
+#include "command_utils.h"
```
hysw:
```suggestion
#include "dive/os/command_utils.h"
```

src/dive/utils/version_info.cpp
```diff
@@ -194,12 +195,39 @@ std::string GetDeviceLibrariesVersionInfo(const std::string& csv_content)
...
+            auto canonical_path = std::filesystem::canonical(potential_path);
+            std::cout << "Found " << file_name << " at " << canonical_path << "\n";
+            return canonical_path;
+        }
+    }
+    std::cerr << "Could not find '" << file_name << "' in any of the search paths.\n";
+    return {};
+}
+
 std::string GetLongVersionString()
```
hysw: Pass the device libraries and profile plugin path in the parameter, so we don't have repeated logic searching for them at multiple places.
```cpp
struct InstallInfo{
  std::filesystem::path device_libraries_path;
  std::filesystem::path profiling_plugin_path;
};
std::string GetLongVersionString(const InstallInfo&);
absl::StatusOr<std::string> GetDeviceLibraryInfo(const std::filesystem::path& device_libraries_path, std::string_view key);
```

angela28chen (approved):

elviscapiaq:
> > > > I tested on Linux for both: Dive CLI and Dive UI, however, I am missing the tests for WIndows. I woudl appreciate if someone can confirm, otherwise I can do tomorrow.
> > > 
> > > 
> > > i can test windows for you. anything specific you want me to verify?
> > 
> > 
> > Thanks. Could you test for release binaries? You should not see any error for pm4/gfxr capture. Also, you should not need to use `windeploy_qt` for it, but make sure that the `DIVE_DEVICE_LIBRARIES_VERSION_FILENAME` and `kProfilingPluginShaPath` are located at `.` or `./install` relative paths.
> 
> tested on windows (UI + CLI), it is working fine. windeploy_qt is still required for Dive to run on a different windows pc though. is that expected (you mentioned no need to run windeploy_qt )?

We do not require windeployqt for testing this current change, as the objective is only to confirm that path resolution is working correctly by checking for relative paths instead of full paths. However, windeployqt is necessary for packaging our release binaries, as its main goal is to ensure end-users do not have to install the Qt libraries separately.

src/dive/os/command_utils_win32.cc
elviscapiaq: Done.

elviscapiaq:
> > * This change resolves the Dive install path depending from build type. If it is for Release, then the path considered are ./install and . relative to the Dive UI binary location. For Debug builds, we keep the original mechanism to find the Dive install path.
> > * The change was applied for both: Dive UI and Dive CLI.
> > * Also, it moves the commands utils from ./capture_service to ./dive/utils.
> 
> just to confirm, the "Release" above means both RelWithDebInfo and Release, right? if that is the case, can you update the description to add RelWithDebInfo as well? thanks

Yes, that is right. I updated the description to encapsule the current change as it is a simpler soluition than the initial one.
Rather than separating search paths for Debug or Release (including RelWithDebInfo) modes, now we just look for all possible paths covering for both Debug and Release mode.

PR #651, opened 2025-12-10
Add build with sanitizer option.

hysw:


wangra-google (approved):

ui/main_window.cpp
```diff
@@ -445,7 +445,6 @@ MainWindow::MainWindow(ApplicationController &controller) :
 
         m_pm4_event_search_bar = new SearchBar(this);
         m_pm4_event_search_bar->setObjectName("Event Search Bar");
-        m_pm4_event_search_bar->setView(m_pm4_command_hierarchy_view);
```
wangra-google: do we need to move this after `m_pm4_command_hierarchy_view = new DiveTreeView(m_data_core->GetCommandHierarchy());`?

hysw: That line already exists.

PR #652, opened 2025-12-10
Improve CMake install semantics.

hysw:
- Exclude third_party from all targets.
  - Not having additional install target.
  - Faster build.
- Remove CMAKE_INSTALL_PREFIX from install(...).
  - Now `cmake --install build --prefix install_prefix` works

### DIFF
--- a/dive_core/data_core.cpp
+++ b/dive_core/data_core.cpp
@@ -942,12 +942,30 @@ void CaptureMetadataCreator::FillColorBlendState(EventStateInfo::Iterator event_
 
         event_state_it->SetAttachment(rt_id, attach);
 
-        const bool logic_op_enabled = (rb_mrt_control.bitfields.ROP_ENABLE == 1);
-        // Be careful!!! Here we assume the enum `VkLogicOp` matches exactly `a3xx_rop_code`
-        const VkLogicOp op = static_cast<VkLogicOp>(rb_mrt_control.bitfields.ROP_CODE);
-        event_state_it->SetLogicOpEnabled(rt_id, logic_op_enabled);
-        event_state_it->SetLogicOp(rt_id, op);
-
+        // 'VkLogicOp' does not map directly to 'a3xx_rop_code'.  Use lookup table.
+        const bool      logic_op_enabled = (rb_mrt_control.bitfields.ROP_ENABLE == 1);
+        const VkLogicOp lookup[] = { VK_LOGIC_OP_CLEAR,          // ROP_CLEAR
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
hysw:
```suggestion
        DESTINATION ${DIVE_INSTALL_DESTINATION}
```

CMakeLists.txt
```diff
@@ -38,6 +38,15 @@ option(
     ON
 )
 
+set(DIVE_INSTALL_DESTINATION
+    "." # Replace this with "dive-profiler"
+    CACHE STRING
+    "Dive install location"
+    FORCE
```
hysw:
```suggestion
    
```

hysw:
Reverted MacOS changes, that part need a bit more work.
A bit more cleanup so the instructions are clear.

---

> If I'm understanding correctly now, this PR is correcting nonstandard usage that we previously had of CMAKE_INSTALL_PREFIX used as DESTINATION inside the install() command in Dive code. Maybe this can be said explicitly in the PR title or description near the "Now ... works" part?

The intentional of this PR is removing the stuff that third_party/* put into the install folder. I had to fix DESTINATION part to get is somewhat working.

footballhead (approved):

ui/CMakeLists.txt
```diff
@@ -210,6 +210,7 @@ endif()
 
 if(APPLE)
     install(TARGETS ${PROJECT_NAME} BUNDLE DESTINATION .)
+    # TODO: fix mac installation directory.
```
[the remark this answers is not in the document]
angela28chen: b/462767957 (Edit: Actually nevermind this feels like it can be a separate bug)

CMakeLists.txt
```diff
@@ -248,26 +258,29 @@ if(ANDROID)
         INPUT
             "${CMAKE_CURRENT_SOURCE_DIR}/third_party/gfxreconstruct/android/layer/build/outputs/aar/layer-${CMAKE_BUILD_TYPE_LOWERCASE}.aar"
         DESTINATION "${CMAKE_BINARY_DIR}/gfxr_layer"
+        # TOUCH # Note: CMake 3.24 feature
```
footballhead: Would be great to be able to use this. I made a bug to investigate upgrading cmake past 3.22

hysw:
Tested on Linux, ui still seems to work. This is supposed to be a no-op, although there might be hard coded dependency somewhere that I'm not aware.

PR #653, opened 2025-12-10
Fix the use description for gfxr_capture_file_dir flag

elviscapiaq:


README.md
```diff
@@ -97,11 +97,13 @@ Examples:
 - Install the dependencies on device, start the package, and initiate a GFXR capture.
     ```sh
     # On Linux
-    ./dive_client_cli --command gfxr_capture --package YOUR.PACKAGE.NAME.HERE --type vulkan_openxr --gfxr_capture_file_dir REMOTE/DIR --download_dir LOCAL/DIR/TO/SAVE/CAPTURES
+    ./dive_client_cli --command gfxr_capture --package YOUR.PACKAGE.NAME.HERE --type vulkan_openxr --gfxr_capture_file_dir REMOTE/SUBDIR --download_dir LOCAL/DIR/TO/SAVE/CAPTURES
     ```
 
 Then you can follow the hint output to trigger a capture by pressing key `g` and `enter`, stopping it with the same key combination, or exiting by pressing key `enter`.
 
+The `--gfxr_capture_file_dir` flag specifies a temporary subdirectory on the Android device (under `/sdcard/Download`) where the GFXR capture is stored. This subdirectory name is mirrored on the host within `--download_dir`. Its default value is `gfxr_capture`.
```
angela28chen: nit: I feel like this info is easily accessible and basically the same as what you'd get with `dive_client_cli --help` so omitting it from README.md could be fine? I think the tweak from REMOTE/DIR->REMOTE/SUBDIR is useful though.

PR #654, opened 2025-12-10
[DO NOT SQUASH] Fast-forward third_party/gfxreconstruct to 321e90045

footballhead:
`subtree pull` in order to get [Improve developer build times by hiding SHA](https://github.com/LunarG/gfxreconstruct/commit/d582d32f0c08d534c002063b85d37f19955a9dfe).

This fast forwards the subtree to [321e90045](https://github.com/LunarG/gfxreconstruct/commit/321e90045) which was the latest commit on `dev` when `git subtree pull` was run

# Reviewer Guide

**Reviewers should only need to review  this commit: https://redirect.github.com/google/dive/pull/654/changes/84d0aaeb4417539391b5b901a8b5d00a1278bc57**

To reduce reviewer burden, I committed the merge conflict markers and fixed them in the following commit. This means that you don't need to wade through 300+ files and can focus on the 20 or so that I actually touched. 

While the commits look confusing, they're fairly straight-forward:

- a96ad06c9b807e1f65bddc499863e81ad4ba68fb: Squash commit on subtree branch generated by `subtree pull`; can largely be ignored since it should match upstream GFXR identically
- b1887f7b2c119633f7a29632e32c5924413d634a: This is #732 which I wanted in this PR for testing. Ideally, #732 is merged first.
- 2e57721d35f649e3c10dce07d2f06948b7bd3473: The merge commit generated by `subtree pull`. This contains merge conflict markers; can largely be ignored
-  https://redirect.github.com/google/dive/pull/654/changes/84d0aaeb4417539391b5b901a8b5d00a1278bc57: My resolution to merge conflicts, as well as other required fixes. **You probably only need to review this**

# Conflicts

## FileProcessor

`FileProcessor` was refactored. While this doesn't alter functionality, it breaks some of our modifications:

- `ProcessStateMarker()` was removed. Its logic now lives in `ProcessVisitor`, `DispatchVisitor`, `ProcessStateEndMarker()`, and `ProcessStateBeginMarker()`.
- `ProcessFrameMarker()` was removed. Its logic now lives in `ProcessVisitor`, `DispatchVisitor`, and `ProcessFrameEndMarker()`
- `ReadBlockHeader()` was effectively replaced. Its logic now lives largely in `BlockProcessor`. There is still a function with the same name but it's functionally different.
- `ActiveFileContext` now stores a `FileInputStreamPtr` instead of a filename. All file operations use `FileInputStreamPtr`
- `SeekActiveFile()` signature was modified to take a `FileInputStreamPtr` instead of a `std::string`

This broke the following features:

- Single frame looping
- PM4 dump
- RenderDoc capture
- Block modification

### Single frame looping

We need to react to State End and Frame End markers. Now, we mark `FileProcessor::ProcessStateEndMarker()` and `FileProcessor::ProcessFrameEndMarker()` as virtual and override those in `DiveFileProcessor`. (Although, in the future, this functionality should be moved into `DiveVulkanReplayConsumer`. This derives from `MarkerConsumerBase` and provides the same functions for consumers to override but in a more stable way).

Given the changes to `FileProcessor`, we can drastically simplify the `DiveFileProcessor::ProcessFrameEndMarker()` implementation. The previous implementation started with a copy of `FileProcessor::ProcessFrameMarker()` then modified it to support looping. Now, the only thing that we need to work around is the assert in `FileProcessor::ProcessFrameDelimiter()` (which ensures frames are processed in order) by manipulating the capture frame number.

Now that all file operations (seek, tell) require `FileInputStreamPtr`, we store a weak reference to the GFXR file instead of a `std::string` in `DiveFileProcessor`.

`FileProcessor::TellFile()` was moved to `FileInputStream::FileTell()`. This class abstracts the file descriptor so it's a logical place to expose a tell API.

Since `DiveFileProcessor` relies on `SeekActiveFile()`, ensure the required function is still `protected`.

### PM4 Dump, RenderDoc Capture

Like single frame looping, this is fixed by making `FileProcessor::ProcessStateEndMarker()` and `FileProcessor:ProcessFrameEndMarker()` virtual. (This should also be moved into `DiveVulkanReplayConsumer`.)

### Block Modification

I ensured that `FileProcessor::StoreBlockInfo()` is still called right before block reads.

`FileProcessor::StoreBlockInfo` is no longer called when EOF is reached. As a result, `DiveBlockData::FinalizeOriginalBlocksMapSizes` has been updated to take the file size in order to calculate the size of the final block.

## VulkanReplayConsumerBase

Some queue flag variables were renamed. When applying our patch, the names were updated.

## replay_settings.h

Ensured that our flags for GPU time and single frame looping are still present.

## android_main

Despite the addition of `--capture` (a.k.a. "recapture"), ensured that Dive additions persist.

### DIFF
--- a/gfxr_ext/decode/dive_file_processor.cpp
+++ b/gfxr_ext/decode/dive_file_processor.cpp
@@ -44,9 +44,6 @@ bool ShouldCreateRenderDocCapture()
 
 }  // namespace
 
-// TODO GH #1195: frame numbering should be 1-based.
-const uint32_t kFirstFrame = 0;
-
 void DiveFileProcessor::SetLoopSingleFrameCount(uint64_t loop_single_frame_count)
 {
     loop_single_frame_count_ = loop_single_frame_count;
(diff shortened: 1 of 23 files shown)


footballhead:
At 466d6fd:
```
The following tests FAILED:
	  3 - Regression_StackOverflow_MacOS (Subprocess aborted)
	  4 - DiveUI_Screenshot (Subprocess aborted)
	  5 - DiveUI_Screenshot_NativeStyle (Subprocess aborted)
	 90 - ModifyGfxrSeparateAssetsNoOp (Failed)
```

footballhead:
> The following tests FAILED:
> 	  3 - Regression_StackOverflow_MacOS (Subprocess aborted)
> 	  4 - DiveUI_Screenshot (Subprocess aborted)
> 	  5 - DiveUI_Screenshot_NativeStyle (Subprocess aborted)
> 	 90 - ModifyGfxrSeparateAssetsNoOp (Failed)

StoreBlockInfo wasn't being called. It was originally in ReadBlockHeader, then was shuffled into PeekBlockHeader by https://github.com/LunarG/gfxreconstruct/commit/3131d856bbc113436cd2276971b128fcdf6eab6d; the two functions are very similar so it probably tripped up the auto conflict resolition. But in that same commit, PreloadFileProcessor stopped using PeekBlockHeader. So there are no more callers

src/dive/common/CMakeLists.txt
```diff
@@ -15,4 +15,7 @@
 #
 
 add_library(dive_status status.h status.cpp)
-target_link_libraries(dive_status PUBLIC dive_src_includes absl::stacktrace)
+target_link_libraries(
+    dive_status
+    PUBLIC dive_src_includes absl::stacktrace absl::symbolize
```
footballhead: This can be split into a separate PR. This has always been an issue, my change just triggered it.

third_party/gfxreconstruct/framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -205,8 +205,7 @@ static uint32_t GetHardwareBufferFormatBpp(uint32_t format)
 
 VulkanReplayConsumerBase::VulkanReplayConsumerBase(std::shared_ptr<application::Application> application,
                                                    const VulkanReplayOptions&                options) :
-    options_(options),
-    loader_handle_(nullptr), get_instance_proc_addr_(nullptr), create_instance_proc_(nullptr),
+    options_(options), loader_handle_(nullptr), get_instance_proc_addr_(nullptr), create_instance_proc_(nullptr),
```
footballhead: is this a clang-format fix? Might want to avoid making those to avoid merge conflicts

third_party/gfxreconstruct/framework/util/span.h
```diff
@@ -160,7 +160,7 @@ class DataSpan
 
     // Create an unowned DataSpan from a pointer or a span... OutputSpan is default initialized no reason to double
     // store the span but *do* set the holds to something value.
-    DataSpan(const DataType* data, size_type size) : data_(data), size_(size), store_(OutputSpan())
+    DataSpan(const DataType* data, size_type size) : size_(size), data_(data), store_(OutputSpan())
```
footballhead: This is a bug in upstream that I needed to fix in order to compile

third_party/gfxreconstruct/tools/replay/android_main.cpp
```diff
@@ -75,8 +75,8 @@ const char kLayerProperty[]      = "debug.vulkan.layers";
 
 const int32_t kSwipeDistance = 200;
 
-void        ProcessAppCmd(struct android_app* app, int32_t cmd);
-int32_t     ProcessInputEvent(struct android_app* app, AInputEvent* event);
+void    ProcessAppCmd(struct android_app* app, int32_t cmd);
```
footballhead: is this a clang-format fix? Might want to avoid making those to avoid merge conflicts

gfxr_ext/decode/dive_file_processor.cpp
```diff
@@ -92,164 +89,142 @@ bool DiveFileProcessor::WriteFile(const std::string& name, const std::string& co
...
-    uint64_t frame_number = 0;
-    bool success = ReadBytes(&frame_number, sizeof(frame_number));
-
-    if (success)
-    {
-        // Validate frame end marker's frame number matches first_frame_ when
-        // capture_uses_frame_markers_ is true.
-        GFXRECON_ASSERT((marker_type != format::kEndMarker) || (!UsesFrameMarkers()) ||
-                        (frame_number == GetFirstFrame()));
+    // Single frame looping, despite the frame number never technically changing, increments
```
footballhead: Comment from @angela28chen on 1c268070fd9ecccad42f351b077c0a9974b93492:

> How about filing an internal cleanup bug to keep current_frame_number_ at 0 and moving these details there?

angela28chen (dismissed):

PR #660, opened 2025-12-16
Componentize DiveUI (part 1).

hysw:


ui/sqtt/ruler_graphics_item.cpp
```diff
@@ -14,12 +14,15 @@
  limitations under the License.
 */
 
-#include "ruler_graphics_item.h"
+#include "ui/sqtt/ruler_graphics_item.h"
```
wangra-google: since you are moving the file anyways, we should probably remove the folder sqtt since we don't have any sqtt support in current Dive

src/dive/ui/components/settings/settings.cpp
```diff
@@ -89,7 +91,7 @@ Settings::DisplayUnit Settings::ReadRulerDisplayUnit()
 {
     QSettings   settings;
     DisplayUnit display_unit = (DisplayUnit)settings.value("rulerDisplayUnit", 0).toInt();
-    DIVE_ASSERT(display_unit == kCycle || display_unit == kMs || display_unit == kUs ||
+    ABSL_ASSERT(display_unit == kCycle || display_unit == kMs || display_unit == kUs ||
```
wangra-google: would it be better to keep DIVE_ASSERT and change it to call ABSL_ASSERT internally? otherwise you need to replace all DIVE_ASSERTs in Dive proj which could be quite a lot

ui/sqtt/ruler_graphics_item.cpp
hysw: #682

PR #665, opened 2025-12-17
Remove PEFFETTO_TRACE_READER_LIB

footballhead:
Perfetto was removed in PR #287. Nothing sets this CMake var anymore.

### DIFF
--- a/cli/CMakeLists.txt
+++ b/cli/CMakeLists.txt
@@ -61,10 +61,7 @@ target_link_libraries(
 )
 
 add_executable(${PROJECT_NAME} ${HDR_FILES} "main.cpp")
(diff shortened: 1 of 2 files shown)


elviscapiaq (approved):
LGTM. I wonder why we were using perfetto trace reader before.

footballhead:
> LGTM. I wonder why we were using perfetto trace reader before.

This was added in #88 and expanded upon in #92. It was integrated into CaptureData but the implementation never added any UI elements; it only printed info to stdout. I believe that Perfetto has some GPU timing information but I think we wrote GPUTime to replace that.

hysw (approved):

PR #666, opened 2025-12-17
Refactor how device resources are located and deployed

angela28chen:
Changes:
- Renaming of "device libraries" -> "device resources"
- Gather some file name and file path constants into `src/dive/utils/device_resources_constants.h`, with some renaming for clarity
- Gather logic for locating the deployment directory (currently `DIVE_ROOT/install`) and provide util `ResolveResourcesLocalPath()`
- Switch on-device deployment folder `/data/local/tmp -> /data/local/tmp/dive` so that full cleanup is easier
- Implement some helper methods for class `AndroidDevice` around deploying files

### DIFF
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -1436,7 +1436,7 @@ absl::Status AndroidDevice::DeployDeviceResource(const std::string_view      &fi
         host_full_lib_path = *ret;
     }
 
-    RETURN_IF_ERROR(m_adb.Run(absl::StrFormat("shell mkdir -p %s", target_dir)));
+    RETURN_IF_ERROR(m_adb.Run(absl::StrFormat(R"(shell mkdir -p '%s')", target_dir)));
 
     return m_adb.Run(
     absl::StrFormat(R"(push "%s" "%s")", host_full_lib_path.generic_string(), target_dir));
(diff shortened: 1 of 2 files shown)


src/dive/utils/CMakeLists.txt
```diff
@@ -27,6 +27,17 @@ if(NOT ANDROID)
     )
 endif()
 
+# === device_libraries =========================================================
+
+if(NOT ANDROID)
+    add_library(device_libraries device_libraries.h device_libraries.cpp)
```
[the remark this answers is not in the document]
angela28chen: Done

capture_service/device_mgr.cc
```diff
@@ -1461,4 +1423,38 @@ const std::filesystem::path &on_device_screenshot_dir)
...
+    RETURN_IF_ERROR(m_adb.Run(absl::StrFormat("shell mkdir -p %s", target_dir)));
+
+    return m_adb.Run(
+    absl::StrFormat(R"(push "%s" "%s")", host_full_lib_path.generic_string(), target_dir));
+}
+
+absl::Status AndroidDevice::CopyWithPermissions(std::string_view package,
+                                                std::string_view file_name)
+{
+    return m_adb.Run(absl::StrFormat(R"(shell run-as %s cp "%s/%s" .)",
```
wangra-google: i could be wrong, but i feel that on Windows, the quotes in "%s/%s" might be removed. so if you have space in the string, this func might fail on windows. Just to be safe, can you verify if that is the case? thanks

angela28chen: I just tested with spaces in file paths, what I have found is that you need to use double quotes for adb push and single quotes for adb shell, I've modified my PR accordingly.

wangra-google (approved):

elviscapiaq (approved):

PR #674, opened 2025-12-19
Refactor TriggerGfxrCapture

footballhead:
It was confusing to read: the focus was more on the user input rather than the steps that it took to capture. This also made it hard to modify. I've refactored it so that it reads more like step by step instructions on how capture is performed.

Additionally,

- Replace `std::chrono` with Abseil time. Abseil is more straight-forward to use. This is also consistent with our style guide.
- Mark `AndroidDevice::TriggerScreenCapture` as const so that I can use `const AndroidDevice&`
- Use `absl::Cleanup` to ensure that folders are removed up in all cases. This makes functions easier to write and reason about since I can use early returns and be guaranteed that cleanup is run.
- Simplify `IsCaptureFinished` to `bool` since the additional information provided by `absl::Status` was not being used
- Move the capture loop outside of `TriggerGfxrCapture` and into the parent command handler. This reduces nesting and complexity of `TriggerGfxrCapture`
- Return one of 3 values from `TriggerGfxrCapture` to provide the caller with more context:
    - `kContinue`: Keep capturing
    - `kStop`: Stop capturing
    - `!ok()`: Error
- Use aliases for commonly used members of parameters to improve readability.
- Remove all abbreviations for readability. E.g. `ctx` becomes `context`, `mgr` becomes `device_manager`, `ret` becomes `status`, etc.
- Prefer `absl::StrCat` for more efficient string concatenation and to remove explicit `std::string` creation
- Simplify messages printed to the user. Particularly, I removed the note about corrupted capture files because there currently isn't a way that the user can trigger than condition.
- Prefer "if with initializers" to scope the status variables and make code more succinct
- Remove unnecessary local variables where readability wasn't hampered. E.g. didn't need to explicitly create separate `std::filesystem::path` from `gfxr_capture_directory`.
- Remove linking `absl::algorithm`. As far as I can tell, this was never used.
- Use `RETURN_IF_ERROR` where appropriate to make code more concise
- Use function argument comments to clarify the meaning of booleans, etc.
- Use more specific function parameters. E.g. a function that only needs `AdbSession` doesn't need to take an entire `Device` or `DeviceManager`.
- Mark parameters that are not mutated as `const`.
- Use `absl::Status::IgnoreError` to notate that the error is intentionally ignored (as per the docs)


capture_service/dive_client_cli.cc
```diff
@@ -560,103 +565,91 @@ absl::Status RetrieveGfxrCapture(Dive::DeviceManager& mgr, const GlobalOptions&
     return absl::OkStatus();
 }
 
-// Triggers a GFXR capture on the device, allowing for multiple captures and screenshot.
-absl::Status TriggerGfxrCapture(Dive::DeviceManager& mgr, const GlobalOptions& options)
+// Triggers a GFXR capture and screenshot on the device.
+absl::StatusOr<bool> TriggerGfxrCapture(Dive::DeviceManager& device_manager,
```
footballhead: Can this be made const?
```suggestion
absl::StatusOr<bool> TriggerGfxrCapture(const Dive::DeviceManager& device_manager,
```

capture_service/dive_client_cli.cc
```diff
@@ -710,7 +703,7 @@ absl::Status CmdPm4Capture(const CommandContext& ctx)
     }
 
     std::cout << "Waiting " << ctx.options.trigger_capture_after << " seconds...\n";
-    std::this_thread::sleep_for(std::chrono::seconds(ctx.options.trigger_capture_after));
+    absl::SleepFor(absl::Seconds(ctx.options.trigger_capture_after));
```
footballhead: Don't touch PM4 case for code that cleans up GFXR capture. Do this in a separate PR
```suggestion
    std::this_thread::sleep_for(std::chrono::seconds(ctx.options.trigger_capture_after));
```

hysw (approved):

capture_service/dive_client_cli.cc
```diff
@@ -560,103 +566,100 @@ absl::Status RetrieveGfxrCapture(Dive::DeviceManager& mgr, const GlobalOptions&
...
+        absl::StrCat("Error starting gfxr runtime capture: ", status.message()));
+    }
+
+    if (absl::Status status = device.TriggerScreenCapture(gfxr_capture_file_dir); !status.ok())
+    {
+        return Dive::InternalError(
+        absl::StrCat("Error creating capture screenshot: ", status.message()));
+    }
+
+    // Ask the user what to do.
```
footballhead: Let the code speak for itself. Obvious comments add visual clutter. Determine which of these comments are necessary; remove the rest.

capture_service/dive_client_cli.cc
```diff
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 #include <algorithm>
+#include <chrono>
```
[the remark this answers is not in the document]
footballhead: https://github.com/google/dive/pull/684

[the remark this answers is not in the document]
footballhead: I've removed the IWYU for chrono it from this PR with the assumption that #684 will clean up the uses that necessitated the IWYU in the first place.

hysw (approved):

elviscapiaq (approved):

hysw (approved):

PR #676, opened 2025-12-19
Restructure nesting of labels

GrantComm:
Restructures nesting of labeling to match renderdoc
**Manual Testing:**
Linux, MacOS
- Load gfxr capture, replay with renderdoc enabled, compare structure of dive tree view to that of renderdoc

### DIFF
--- a/dive_core/gfxr_vulkan_command_hierarchy.cpp
+++ b/dive_core/gfxr_vulkan_command_hierarchy.cpp
@@ -201,11 +201,14 @@ GfxrVulkanCommandHierarchyCreator::GfxrVulkanCommandHierarchyCreator(
 GfxrVulkanCommandHierarchyCreator::~GfxrVulkanCommandHierarchyCreator() {}
 
 //--------------------------------------------------------------------------------------------------
-uint64_t GfxrVulkanCommandHierarchyCreator::GetActiveParent(uint64_t default_parent) const
+void GfxrVulkanCommandHierarchyCreator::AddChildToActiveParent(uint64_t child_node_index,
+                                                               uint64_t default_parent)
 {
-    return m_parsing_state.parent_node_index_stack.empty()
-               ? default_parent
-               : m_parsing_state.parent_node_index_stack.top();
+    uint64_t parent = m_parsing_state.parent_node_index_stack.empty()
+                          ? default_parent
+                          : m_parsing_state.parent_node_index_stack.top();
+
+    AddChild(CommandHierarchy::TopologyType::kAllEventTopology, parent, child_node_index);
 }
 
 //--------------------------------------------------------------------------------------------------
@@ -226,8 +229,7 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
             AddNode(NodeType::kGfxrVulkanBeginCommandBufferNode, vk_cmd_string_stream.str());
         m_parsing_state.command_buffer_node_index = cmd_buffer_index;
         GetArgs(vulkan_cmd_args, m_parsing_state.command_buffer_node_index);
-        AddChild(CommandHierarchy::TopologyType::kAllEventTopology,
-                 GetActiveParent(m_parsing_state.submit_node_index), cmd_buffer_index);
+        AddChildToActiveParent(cmd_buffer_index, m_parsing_state.submit_node_index);
         m_parsing_state.parent_node_index_stack.push(cmd_buffer_index);
     }
     else if (vulkan_cmd_name == "vkEndCommandBuffer")
@@ -244,8 +246,7 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
 
         GetArgs(vulkan_cmd_args, cmd_buffer_index);
 
-        AddChild(CommandHierarchy::TopologyType::kAllEventTopology,
-                 GetActiveParent(m_parsing_state.submit_node_index), cmd_buffer_index);
+        AddChildToActiveParent(cmd_buffer_index, m_parsing_state.submit_node_index);
     }
     else if (vulkan_cmd_name.find("BeginDebugUtilsLabelEXT") != std::string::npos)
     {
@@ -253,9 +254,8 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
         uint64_t begin_debug_utils_label_cmd_index =
             AddNode(NodeType::kGfxrBeginDebugUtilsLabelCommandNode, label_name.c_str());
         GetArgs(vulkan_cmd_args, begin_debug_utils_label_cmd_index);
-        AddChild(CommandHierarchy::TopologyType::kAllEventTopology,
-                 GetActiveParent(m_parsing_state.submit_node_index),
-                 begin_debug_utils_label_cmd_index);
+        AddChildToActiveParent(begin_debug_utils_label_cmd_index,
+                               m_parsing_state.submit_node_index);
         m_parsing_state.parent_node_index_stack.push(begin_debug_utils_label_cmd_index);
     }
     else if (vulkan_cmd_name.find("EndDebugUtilsLabelEXT") != std::string::npos)
@@ -274,8 +274,7 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
         uint64_t vk_cmd_index =
             AddNode(NodeType::kGfxrVulkanDrawCommandNode, vk_cmd_string_stream.str());
         GetArgs(vulkan_cmd_args, vk_cmd_index);
-        AddChild(CommandHierarchy::TopologyType::kAllEventTopology,
-                 GetActiveParent(m_parsing_state.command_buffer_node_index), vk_cmd_index);
+        AddChildToActiveParent(vk_cmd_index, m_parsing_state.command_buffer_node_index);
     }
     else if (vulkan_cmd_name.find("vkCmdBeginRenderPass") != std::string::npos)
     {
@@ -288,8 +287,7 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
         uint64_t vk_cmd_index =
             AddNode(NodeType::kGfxrVulkanBeginRenderPassCommandNode, vk_cmd_string_stream.str());
         GetArgs(vulkan_cmd_args, vk_cmd_index);
-        AddChild(CommandHierarchy::TopologyType::kAllEventTopology,
-                 GetActiveParent(m_parsing_state.command_buffer_node_index), vk_cmd_index);
+        AddChildToActiveParent(vk_cmd_index, m_parsing_state.command_buffer_node_index);
         m_parsing_state.parent_node_index_stack.push(vk_cmd_index);
     }
     else if (vulkan_cmd_name.find("vkCmdEndRenderPass") != std::string::npos)
@@ -304,8 +302,7 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
         uint64_t vk_cmd_index =
             AddNode(NodeType::kGfxrVulkanEndRenderPassCommandNode, vk_cmd_string_stream.str());
         GetArgs(vulkan_cmd_args, vk_cmd_index);
-        AddChild(CommandHierarchy::TopologyType::kAllEventTopology,
-                 GetActiveParent(m_parsing_state.command_buffer_node_index), vk_cmd_index);
+        AddChildToActiveParent(vk_cmd_index, m_parsing_state.command_buffer_node_index);
     }
     else
     {
@@ -333,8 +330,7 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
 
         uint64_t vk_cmd_index = AddNode(node_type, vk_cmd_string_stream.str());
         GetArgs(vulkan_cmd_args, vk_cmd_index);
-        AddChild(CommandHierarchy::TopologyType::kAllEventTopology,
-                 GetActiveParent(m_parsing_state.command_buffer_node_index), vk_cmd_index);
+        AddChildToActiveParent(vk_cmd_index, m_parsing_state.command_buffer_node_index);
     }
 }
 
--- a/dive_core/gfxr_vulkan_command_hierarchy.cpp
+++ b/dive_core/gfxr_vulkan_command_hierarchy.cpp
@@ -201,20 +201,10 @@ GfxrVulkanCommandHierarchyCreator::GfxrVulkanCommandHierarchyCreator(
 GfxrVulkanCommandHierarchyCreator::~GfxrVulkanCommandHierarchyCreator() {}
 
 //--------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -385,6 +383,7 @@ bool GfxrVulkanCommandHierarchyCreator::ProcessGfxrSubmits(const GfxrCaptureData
             }
         }
     }
+    while (!m_cur_parent_node_index_stack.empty()) m_cur_parent_node_index_stack.pop();
```
wangra-google: i guess you can just do this? `m_cur_parent_node_index_stack = {};`

wangra-google: also i am curious in what case this m_cur_parent_node_index_stack is not empty? should we assert here?

angela28chen: By moving the clear here, is there a chance that we can early return out of `ProcessGfxrSubmits()` without having cleared `m_cur_parent_node_index_stack` and will that potentially cause an issue? I'm wondering if a deferred cleanup is more appropriate

angela28chen: Or potentially adding a clear at the beginning of this function

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -235,26 +225,34 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
...
-        AddChild(CommandHierarchy::TopologyType::kAllEventTopology, m_cur_submit_node_index,
-                 cmd_buffer_index);
+        AddChild(CommandHierarchy::TopologyType::kAllEventTopology,
+                 GetActiveParent(m_cur_submit_node_index), cmd_buffer_index);
+        m_cur_parent_node_index_stack.push(cmd_buffer_index);
     }
     else if (vulkan_cmd_name == "vkEndCommandBuffer")
     {
+        if (!m_cur_parent_node_index_stack.empty() &&
+            m_cur_parent_node_index_stack.top() == m_cur_command_buffer_node_index)
```
angela28chen: nit: put round brackets around the second clause for readability:

```
(!m_cur_parent_node_index_stack.empty() &&
            (m_cur_parent_node_index_stack.top() == m_cur_command_buffer_node_index))
```

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -235,26 +225,34 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
...
         GetArgs(vulkan_cmd_args, m_cur_command_buffer_node_index);
-        AddChild(CommandHierarchy::TopologyType::kAllEventTopology, m_cur_submit_node_index,
-                 cmd_buffer_index);
+        AddChild(CommandHierarchy::TopologyType::kAllEventTopology,
+                 GetActiveParent(m_cur_submit_node_index), cmd_buffer_index);
+        m_cur_parent_node_index_stack.push(cmd_buffer_index);
     }
     else if (vulkan_cmd_name == "vkEndCommandBuffer")
     {
+        if (!m_cur_parent_node_index_stack.empty() &&
```
angela28chen: Are there valid cases where this if condition would not be true? (mismatched vkEndCommandBuffer)? Should we log an error/warning?

dive_core/gfxr_vulkan_command_hierarchy.h
angela28chen: It seems like these 3 members (m_cur_*) are basically counters used to track things when parsing through the capture data. What about bundling them together into a struct, and then explicitly resetting/clearing them right before use and after completion (beginning and end of `ProcessGfxrSubmits()`)? I feel like this could make the code more readable and help eliminate future errors with forgetting to have cleared something (especially during non-fatal errors).

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: There could be a case where the last command buffer may not contain its closing matching command (such as vkEndCommandBuffer, vkCmdEndRenderPass, or EndDebugUtilsLabelEXT) due to the start/end of the gfxr capture. This just makes sure the stack is completely cleared.

GrantComm: Yes, an early return due to a parsing error would bypass the closing clear(). Added a clear at the beginning of the function

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: Done

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: Yes, if a Render Pass (vkCmdBeginRenderPass) or a command-buffer label (vkCmdBeginDebugUtilsLabelEXT) is started but never closed inside the command buffer, those node indices remain on top of the stack.
When vkEndCommandBuffer is processed, the stack top will be that unclosed Render Pass/Debug Label node instead of the command buffer node (m_cur_command_buffer_node_index). Hence, the check returns false.
hierarchy

dive_core/gfxr_vulkan_command_hierarchy.h
GrantComm: Added ParsingState to hold those three members along with a Reset() function that resets them

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -201,20 +201,11 @@ GfxrVulkanCommandHierarchyCreator::GfxrVulkanCommandHierarchyCreator(
 GfxrVulkanCommandHierarchyCreator::~GfxrVulkanCommandHierarchyCreator() {}
 
 //--------------------------------------------------------------------------------------------------
-void GfxrVulkanCommandHierarchyCreator::ConditionallyAddChild(uint64_t node_index)
+uint64_t GfxrVulkanCommandHierarchyCreator::GetActiveParent(uint64_t default_parent) const
```
wangra-google: maybe it is cleaner if we put AddChild in the helper function? something like:
```
void GfxrVulkanCommandHierarchyCreator::AddChildToActiveParent(uint64_t child_node_index, uint64_t default_parent)
{
    uint64_t parent = m_parsing_state.parent_node_index_stack.empty()
                          ? default_parent
                          : m_parsing_state.parent_node_index_stack.top();
                          
    AddChild(CommandHierarchy::TopologyType::kAllEventTopology, parent, child_node_index);
}
```

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -233,38 +224,48 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
...
+                 GetActiveParent(m_parsing_state.submit_node_index), cmd_buffer_index);
+        m_parsing_state.parent_node_index_stack.push(cmd_buffer_index);
     }
     else if (vulkan_cmd_name == "vkEndCommandBuffer")
     {
+        if (!m_parsing_state.parent_node_index_stack.empty() &&
+            (m_parsing_state.parent_node_index_stack.top() ==
+             m_parsing_state.command_buffer_node_index))
+        {
+            m_parsing_state.parent_node_index_stack.pop();
```
wangra-google, line 241: just want to make sure i understand this correctly:
- BeginDebugUtilsLabelEXT pushes a label node onto the stack.
- The trace hits vkEndCommandBuffer.
- The code checks the top of the stack. It sees the debug_label node, not the command_buffer_node_index.
- The if statement evaluates to false since the top is not command_buffer_node_index. It does not pop.
- The vkEndCommandBuffer finishes.

Now the stack has the unclosed debug_label and the old cmdbuffer in it. Every single command in the next command buffer will be parented to the unclosed label from the previous command buffer. Would this be problematic?

dive_core/gfxr_vulkan_command_hierarchy.cpp
wangra-google: sorry a bit confused here, the screenshot looks wrong? is this before or after your fix?

dive_core/gfxr_vulkan_command_hierarchy.cpp
angela28chen: Thanks!

dive_core/gfxr_vulkan_command_hierarchy.cpp
angela28chen: Is it a valid GFXR capture if a renderpass or a label is started within a command buffer but not closed before the end of the command buffer?

dive_core/gfxr_vulkan_command_hierarchy.h
angela28chen: Good idea, thanks!

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: Added AddChild(...) call to AddChildToActiveParent(...)

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: > sorry a bit confused here, the screenshot looks wrong? is this before or after your fix?

This is after the change, which ensures the commands are nestled under the correct labels. In this case, the MainLightShadow label begins under one CommandBuffer, contains the end of that command buffer, and the beginning of another command buffer. Attached is the renderdoc equivalent
Screenshot 2026-05-21 at 2 06 24 PM

wangra-google: ah, ok, so basically the app is calling vkCmdEndDebugUtilsLabelEXT for MainLightShadow & Shadows.DrawSRPBatcher while recording another cmd buffer?

wangra-google (approved):

dive_core/gfxr_vulkan_command_hierarchy.cpp
wangra-google: had discussions with Grant offline, the vkCmdEndDebugUtilsLabelEXT for MainLightShadow & Shadows.DrawSRPBatcher are called while recording the next cmdbuffer.

wangra-google: > Is it a valid GFXR capture if a renderpass or a label is started within a command buffer but not closed before the end of the command buffer?

not for renderpass, but yes for debug label

dive_core/gfxr_vulkan_command_hierarchy.cpp
wangra-google: as discussed offline, this would be problematic for treeview anyways. current behavior is the same as renderdoc, so we would leave it as this for now.

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -233,38 +227,45 @@ void GfxrVulkanCommandHierarchyCreator::OnCommand(
...
         std::string label_name = vulkan_cmd_args["pLabelInfo"]["pLabelName"];
-
         uint64_t begin_debug_utils_label_cmd_index =
             AddNode(NodeType::kGfxrBeginDebugUtilsLabelCommandNode, label_name.c_str());
         GetArgs(vulkan_cmd_args, begin_debug_utils_label_cmd_index);
-        ConditionallyAddChild(begin_debug_utils_label_cmd_index);
-        m_cur_parent_node_index_stack.push(begin_debug_utils_label_cmd_index);
+        AddChildToActiveParent(begin_debug_utils_label_cmd_index,
+                               m_parsing_state.submit_node_index);
+        m_parsing_state.parent_node_index_stack.push(begin_debug_utils_label_cmd_index);
```
angela28chen, line 259: I wonder if it's better to omit this (not have the debug label be a possible parent but instead keep it flat, since it's not matched well to closing debug label cmds)

dive_core/gfxr_vulkan_command_hierarchy.h
```diff
@@ -75,11 +75,23 @@ class GfxrVulkanCommandHierarchyCreator
     // Updates m_node_children
     void AddChild(CommandHierarchy::TopologyType type, uint64_t node_index,
                   uint64_t child_node_index);
-    void ConditionallyAddChild(uint64_t node_index);
+    void AddChildToActiveParent(uint64_t child_node_index, uint64_t default_parent);
```
angela28chen, line 78: Now I'm wondering if we need both "AddChild()" and "AddChildToActiveParent()". It seems like we could always go with the latter if during the `ParsingState.Reset()` call we make sure to set up the stack with the root node. Might need to make other adjustments to when things are put on the stack to make the `GetArgs()` parsing work as well.

Doesn't have to be addressed in this PR, just bringing this up to consider for the future, since having the command hierarchy code internally consistent would make it easier to follow.

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: I am not sure about that since then it wouldn't be clear which commands belong to a specific label. Is the screenshot what you had in mind?
Screenshot 2026-05-22 at 10 41 11 AM

footballhead (commented):
I'm not sure that you can neatly mix nesting of debug label regions and submits/command buffers/render passes. This is what the docs say:

> An application may open a debug label region in one command buffer and close it in another, or otherwise split debug label regions across multiple command buffers or multiple queue submissions. When viewed from the linear series of submissions to a single queue, the calls to vkCmdBeginDebugUtilsLabelEXT and vkCmdEndDebugUtilsLabelEXT must be matched and balanced.
>
> \- https://docs.vulkan.org/refpages/latest/refpages/source/vkCmdEndDebugUtilsLabelEXT.html

Even a simple case like this is valid and would produce unexpected results:

- begin debug utils label
- begin render pass
- end debug utils label
- end render pass

My understanding is that it would be rendered as:

- begin debug label region (1)
  - begin render pass (2)
  - end debug label region (3)
-  end render pass (4)

And (1) would be left on the stack

If the intent is to match RenderDoc, I'd dig a bit more into Renderdoc's behavior. It looks like it might only use the debug label regions for nesting and ignore the submit/command buffer/renderpass nesting

dive_core/gfxr_vulkan_command_hierarchy.cpp
angela28chen: I think Michael talks about it in more detail in his comment, but basically if the debug labels don't have to begin and end in the same command buffer (as per Ran's comment in chat) then the nesting gets messed up when we use the same stack.

e.g. imagine the round brackets are command buffer and submits nesting properly, while the square brackets is a debug label that starts and ends in a separate command buffer:

```
( ( ) ( [ *) ( ] ) )
```

I think the stack would be messed up because it's no longer LIFO at the * sign, so you can't pop out anything from the stack properly after this point until the `]` is reached

PR #679, opened 2025-12-22
Move from Ninja to Ninja Multi-Config

angela28chen:
This is to make the directory structure of the build folder more similar across different platform, since on Windows we use Visual Studio which is Multi-Config. 

Announcements:
- `CMAKE_BUILD_TYPE` removed and disallowed, do not use this flag anymore
- Now we require installing host tools under `install/` as part of the build process, check out the updated documentation

Changes:
- Switched generators ("Ninja" -> "Ninja Multi-Config") when using cmake to generate build files for:
  - Host tools (host platforms: Linux/macOS, Windows remains VS generator)
  - Device resources (host platforms: Linux/macOS/Windows)
- Updated scripts, documentation
- Updated github workflows
  - Added new step to run cmake install
- Renamed regression test "Regression_StackOverflow_MacOS" -> "Regression_StackOverflow" because it runs on all host platforms and the name is misleading

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -298,7 +298,7 @@ if(NOT ANDROID)
     # This has been fixed in https://github.com/google/dive/pull/541
     if(APPLE)
         set(DIVE_EXECUTABLE_PATH
-            ${CMAKE_BINARY_DIR}/ui/dive.app/Contents/MacOS/dive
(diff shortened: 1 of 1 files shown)


GrantComm (approved):

angela28chen:
Note that the internal release process documentation we have will need to be updated, but I plan to do that at the end of this series of changes.

PR #684, opened 2026-01-05
Replace chrono with absl::time in dive_client_cli

footballhead:
Google prefers Abseil's time library over std::chrono since it's simpler, easier to use, and more performant.


elviscapiaq (approved):
Thanks for updating it!

PR #687, opened 2026-01-05
Replace Pm4InfoInit with call once init.

hysw:
Note: "pm4_info.h" is a generated file, and have convoluted include path requirements.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,10 +21,11 @@ project(dive VERSION 1.1.2)
 # ------------------------------------------------------------------------------
 # Global properties
 
-if(CMAKE_BUILD_TYPE)
+get_property(is_multi_config GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+if(NOT is_multi_config)
     message(
         FATAL_ERROR
-        "Due to using Multi-Config generators, the flag CMAKE_BUILD_TYPE is not supported. Try -DCMAKE_BUILD_TYPE=\"\" or a clean build."
+        "Currently we only support Multi-Config generators, e.g. try -G \"Ninja Multi-Config\" rather than -GNinja."
     )
 endif()
 


wangra-google (dismissed):

wangra-google (approved):

hysw:
> Still not understanding what replacing Pm4InfoInit with a custom init has to do with dive_core include path.
Doesn't replacing "pm4_info.h" with "dive_core/pm4_info.h" do the same thing?

`pm4_info.h` is a generated file that resides in `${DIVE}/build/dive_core/pm4_info.h` which requires `-I${DIVE}/build/dive_core` to be included in every dependent target, having a "dive_core/dive_core_init.h" means the the normal `-I${DIVE}` is sufficient.

--- 
Note: This is the include path for dive ui, and about half of them does not need to be there
```sh
-I/repo/dive/build/ui
-I/repo/dive/ui
-I/repo/dive/build/ui/dive_ui_autogen/include_Debug
-I/repo/dive/build/dive_core
-I/repo/dive/third_party/libarchive/libarchive
-I/repo/dive/LibArchive_INCLUDE_DIRS
-I/repo/dive/third_party/Vulkan-Headers/include
-I/repo/dive
-I/repo/dive/build
-I/repo/dive/third_party/abseil-cpp
-I/repo/dive/src
-I/repo/dive/dive_core/../third_party/gfxreconstruct/framework
-I/repo/dive/build/dive_core/freedreno/isa
-I/repo/dive/third_party/mesa/src/freedreno/common
-I/repo/dive/third_party/mesa/src
-I/repo/dive/third_party/mesa/include
-I/repo/dive/src/dive
-I/repo/dive/build/generated/Debug/include
```

hysw:
> And just to be clear, we still need I${DIVE}/build/dive_core for dive_core project itself still, since dive_core_init.cpp includes pm4_init.h. But the benefit is that this I${DIVE}/build/dive_core is only needed for dive_core and not everywhere else. Is that correct?

Yes. dive_core itself still include everything it needs, just that it does not need to pollute the include directory of other targets. Although there's still some reference to "adreno.h" that need to be cleaned up, which I'll do in a different PR.

A secondary benefit of not having include to a generated file is that all the static analysis tool works better (e.g. github can find symbol definition of `InitDiveCore` but not `Pm4InfoInit`)

PR #692, opened 2026-01-06
Allow opening archive files.

hysw:
- Add a "dive aware" archive extractor.
  - Only extract relevant files, and discard directory hierarchy of the archive.
- Extract content to temp file if we opened an archive.
- Update path lookup logic.

### DIFF
--- a/src/dive/utils/CMakeLists.txt
+++ b/src/dive/utils/CMakeLists.txt
@@ -24,7 +24,7 @@ if(NOT ANDROID)
     target_link_libraries(
         dive_archive
         PUBLIC dive_src_includes absl::statusor
-        PRIVATE dive_import_libarchive
+        PRIVATE LibArchive::LibArchive
     )
 endif()
 
--- a/cli/CMakeLists.txt
+++ b/cli/CMakeLists.txt
@@ -37,7 +37,6 @@ include_directories(
     ${THIRDPARTY_DIRECTORY}/Vulkan-Headers/include
     ${CMAKE_SOURCE_DIR}
     ${CMAKE_BINARY_DIR}
-    ${LibArchive_INCLUDE_DIRS}
 )
 
 add_definitions(-DLITTLEENDIAN_CPU)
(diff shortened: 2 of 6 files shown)


src/dive/utils/CMakeLists.txt
```diff
@@ -17,6 +17,17 @@
...
 list(APPEND CMAKE_MESSAGE_INDENT "  ")
 
+# === dive_archive =============================================================
+
+if(NOT ANDROID)
+    add_library(dive_archive dive_archive.h dive_archive.cpp)
+    target_link_libraries(
+        dive_archive
+        PUBLIC dive_src_includes absl::statusor
+        PRIVATE dive_import_libarchive
```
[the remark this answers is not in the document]
hysw: There's a bunch of weirdness in `libarchive.cmake` for the `find_package` failed case, I'll have to take a bit more time figuring out what is going on there.

src/dive/utils/dive_archive.cpp
```diff
@@ -0,0 +1,168 @@
...
+constexpr auto kKnownDiveArchiveExtension =
+    std::to_array<std::string_view>({".zip", ".tar", ".tar.gz", ".tgz"});
+
+// We don't produce captalized extension, so check for lower case only.
+constexpr auto kKnownDiveExtension =
+    std::to_array<std::string_view>({".rd", ".gfxr", ".gfxa", ".png", ".csv"});
+
+std::string_view GetArchiveEntryPath(struct archive_entry* entry)
+{
+    const char* pathname = archive_entry_pathname(entry);
```
footballhead: who owns pathname? what is the lifetime?

src/dive/utils/dive_archive.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+std::vector<std::filesystem::path> DiveArchive::ExtractTo(std::filesystem::path dst)
+{
+    constexpr size_t kBlockSize = 0x10000;  // 64KiB, we don't really care.
+    auto reader = ArchiveReadNew();
+    archive_read_support_format_all(reader.get());
+    archive_read_support_filter_all(reader.get());
+
+    if (auto res = archive_read_open_filename(reader.get(), m_path.string().c_str(), kBlockSize);
+        res != ARCHIVE_OK)
```
footballhead: nit: since res isn't used outside the if, can be omitted
```suggestion
    if (archive_read_open_filename(reader.get(), m_path.string().c_str(), kBlockSize) != ARCHIVE_OK)
```

PR #695, opened 2026-01-07
Save As .dive.zip

hysw:


src/dive/utils/dive_archive.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+        auto entry = ArchiveEntryNew();
+        archive_entry_set_pathname(entry.get(), component.filename().string().c_str());
+
+        struct stat st = {};
+        stat(component.string().c_str(), &st);
+        archive_entry_copy_stat(entry.get(), &st);
+
+        archive_write_header(writer.get(), entry.get());
+        WriteArchiveEntry(writer.get(), infile);
```
wangra-google: should we check the return value?

hysw: Probably, this is still a draft.

hysw:
> My concern is the consistency of the user experience. If we argue that "Save As" is necessary because manual file copying is a hassle, that logic should apply to the .csv too.

Yes, but you probably want "Export" csv instead of "Save As". 

> If a user clicks "Save As" expecting to save their current analysis (the lightweight .csv), but the tool forces them to wait for a 2GB bundle to be written, that creates friction. 

Dive can't open a csv file, and I don't think "Open" a csv file makes sense.

> I just want to ensure we don't introduce a new option that ends up confusing users more than it helps.

That's the argument to have it do exactly one thing - save everything into a bundle - and having csv export being a different menu item.

PR #699, opened 2026-01-12
Enable warnings as errors in capture_service

footballhead:
Force developers to deal with warnings, either by fixing or squelching, to improve code health.

Use the non-target-specific `add_compile_options` instead of `target_compile_options` since I want it to should apply by default to all targets

### DIFF
--- a/capture_service/dive_client_cli.cc
+++ b/capture_service/dive_client_cli.cc
@@ -18,6 +18,7 @@ limitations under the License.
 #include <filesystem>
 #include <future>
 #include <iostream>
+#include <optional>
 #include <ostream>
 #include <string>
 #include <system_error>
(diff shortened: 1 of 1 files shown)


capture_service/CMakeLists.txt
```diff
@@ -22,6 +22,13 @@ project(DiveService)
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
+add_library(warnings_as_errors INTERFACE)
```
footballhead: I tried to stick to using targets since it's the recommended approach in general but (if we want this to be a default) maybe using CMAKE_CXX_FLAGS is better. It would be fewer lines of code and we can't forget when we introduce new targets. Specific targets can use the default then opt out of checks that they don't want.

hysw: imo, add the compile option on directories that contains dive code, so no one forget to add the target to dependency. Also use `add_compile_options` instead of `CMAKE_CXX_FLAGS`?

PR #704, opened 2026-01-13
Exit if plugin loading failed during test.

hysw:


ui/application_controller.cpp
```diff
@@ -36,6 +37,9 @@ ApplicationController::ApplicationController() {}
 
 ApplicationController::~ApplicationController() {}
 
+bool ApplicationController::Interactive() { return m_impl->m_interactive; }
```
angela28chen: nit: probably personal preference but I think `IsInteractive()` is more clear?

angela28chen (dismissed):
Thanks!

ui/application_controller.cpp
```diff
@@ -77,6 +81,10 @@ bool ApplicationController::InitializePlugins()
     if (absl::Status load_status = m_impl->m_plugin_manager.LoadPlugins(plugins_dir_path);
         !load_status.ok())
     {
+        if (!Interactive())
```
angela28chen: Actually if it's possible to construct the error message used below for the warning box, and log it to qDebug() before the return, I think that might be more consistent?

footballhead (commented):
+1 to angela's suggestions

GrantComm (dismissed):

ui/application_controller.h
```diff
@@ -31,6 +32,9 @@ class ApplicationController : public QObject
     ApplicationController();
     ~ApplicationController();
 
+    bool IsInteractive();
```
wangra-google: bool IsInteractive() const;

ui/application_controller.cpp
```diff
@@ -28,6 +29,7 @@ struct ApplicationController::Impl
 {
     MainWindow* m_main_window = nullptr;
     QAction* m_advanced_option = nullptr;
+    bool m_interactive = true;
```
wangra-google, line 32: it doesn't really change the size here, but better to keep members sorted by size in general so that in case any struct grows, we don't accidentally accumulate memory padding.

hysw: I'd ignore sorting by size rules. Unless a object have millions of instance during runtime, it's just a few bytes, and RAM is not at hundreds dollar per kilobytes yet.

PR #708, opened 2026-01-14
Never store a partially-initialized AndroidDevice

footballhead:
Hide AndroidDevice two-part creation behind a factory function so that DeviceManager::m_device is never partially initialized.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -86,6 +86,12 @@ set DIVE_ROOT_PATH=C:\path\to\dive
     cmake --install build --prefix install --config Debug
     ```
 
+If you want to build a GFXR tool like `gfxrecon-convert`:
+
+```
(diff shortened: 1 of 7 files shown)


capture_service/device_mgr.h
```diff
@@ -103,7 +103,9 @@ absl::StatusOr<GfxrReplaySettings> ValidateGfxrReplaySettings(const GfxrReplaySe
 class AndroidDevice
 {
  public:
-    explicit AndroidDevice(const std::string& serial);
+    static absl::StatusOr<absl_nonnull std::unique_ptr<AndroidDevice>> Create(
```
footballhead: absl_nonnull causes compilation failures that seem unrelated to this change. I'm going to remove them for now and look at using them in a later PR

hysw (approved):

PR #710, opened 2026-01-14
Adds Crashpad as a static library

elviscapiaq:
- This change avoids adding any submodule to build Crashpad libraries and the crashpad_handler binary.
- The Crashpad libraries and the crashpad_handler binary are only built when they were not prebuilt.


elviscapiaq:
I fixed the issue on macOS. @angela28chen @GrantComm , could you help me to test on macOS?
To test you will need to add `#include "dive_crashpad/crashpad_client.h"` header in the ui/main.cpp
and in the main function , initialize crashpad with:

```
int main(int argc, char* argv[])
{
    Dive::AttachToTerminalOutputIfAvailable();
    std::vector<char*> positional_args = SetupFlags(argc, argv);

    if (auto ret = Dive::InitializeCrashpad(); ret.ok())
    {
        qDebug() << "Crashpad initialized successfully.";
    }
    else
    {
        qDebug() << "Crashpad initialization failed: " << ret.message().data();
    }
    ....
  }
```
Please let me know the output of that, once you launch Dive ui.

wangra-google:
a more general question: would there be a risk of having prebuild libs, but somehow compiler got updated to a different version, and then having abi mismatch? I guess the safest way is to let people to delete the prebuild folder, but that is kinda against using the prebuild as a cache? do we have some ways to warning people about this potential risk? i am more worried about potential runtime bugs caused by this which might be tricky to detect.

elviscapiaq:
> LGTM, just my earlier comment about the file structure in this PR

I do not see any comments about the file structure, maybe I lost them because I push with force this PR?
Do you know how to check previous comments?

angela28chen (approved):
OK, sounds reasonable to me to keep the crashpad stuff relatively together as opposed to separate top-level folders

PR #711, opened 2026-01-14
Resets command buffer model and view

GrantComm:
Resets command buffer model and view. Currently there is a crash when loading a .rd file, selecting a draw call and opening the packets tab view, then trying to load another .rd file. This is because the underlying topology data is cleared while loading a new file without the CommandBuffer view being notified, causing the UI to attempt to repaint a node index that no longer exists.


hysw:
For longer term solution, we probably want to check if we've acquired the trace
e.g.
```cpp
class CommandBufferModel {
  std::shared_ptr<bool> m_capture_acquired;
}

...
if(!*m_capture_acquired) { return; }
use m_command_hierarchy
...
```

PR #712, opened 2026-01-15
Properly escape commandline string.

hysw:


capture_service/dive_client_cli.cc
```diff
@@ -763,7 +763,7 @@ absl::Status CmdGfxrCapture(const CommandContext& context)
             absl::StrCat(Dive::kDeviceCapturePath, "/", context.options.gfxr_capture_file_dir);
         context.mgr.GetDevice()
             ->Adb()
-            .Run(absl::StrFormat("shell rm -rf %s", on_device_capture_directory))
+            .Shell(absl::StrFormat("rm -rf %s", on_device_capture_directory))
```
wangra-google: not part of your change, but i feel that this cmd would fail if there is any space in on_device_capture_directory without quote?

wangra-google: would this fail if there is a space in on_device_capture_directory?

i see you have 2 overload Shell(), and this will trigger the 
```
    absl::Status Shell(std::string_view command) const
    {
        return RunCommand("adb", {"-s", m_serial, "shell", std::string(command)}).status();
    }
``` 
which doesn't call "MakeUnixCommand?" on the command string?

src/dive/os/command_utils.cpp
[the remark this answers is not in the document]
hysw: e.g. Python expose both Windows and Linux path function, and you can run the following on Linux environment

```py
import pathlib
print(str(pathlib.PureWindowsPath("C:/Program Files")))
```

PR #719, opened 2026-01-19
Github workflow refactor

angela28chen:
- Cache linux and windows host tests as well
- Refactor host-specific building into reusable workflows that the presubmits and scheduled tests can draw from
- Use github workflow matrix for the different build_types and compilers
- Presubmits are now a subset of full tests (no Debug builds except Windows platform)
- Presubmits aren't cached anymore to preserve storage space
- Pin non-Github actions by hash for security
- Formatted .yml files (tweaking indentation)

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -12,7 +12,7 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref }}
   cancel-in-progress: true
 
-# Arbitrarily chosen subset of the full combination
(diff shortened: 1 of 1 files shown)


RenfengLiu (dismissed):

angela28chen:
> We probably need to disable status check part of branch protection before submit and reapply afterward.

Also I think we should hold off on merging in the evening today (otherwise mac presubmits may become way too slow until cache is repopulated?)

hysw (approved):

RenfengLiu (approved):

PR #720, opened 2026-01-20
Depend on libarchive target

footballhead:
Modern CMake is moving away from the old directory-based commands like `include_directories()` since they aren't very explicit (so apply to broadly) and are not very robust.

Case-in-point: this line (which is removed by this PR) doesn't actually add the libarchive headers to the include path:

```
include_directories(LibArchive_INCLUDE_DIRS)
```

It adds the literal "LibArchive_INCLUDE_DIRS" (which doesn't exist) to the include path.

find_package(LibArchive) will define the LibArchive::LibArchive target. To be consistent with that behavior, also define the same target if we build from source. Then, fix all consumers to link against the target.

### DIFF
--- a/.github/workflows/build_android.yml
+++ b/.github/workflows/build_android.yml
@@ -6,6 +6,9 @@ on:
       build_type:
         required: true
         type: string
+      restore_cache:
+        required: true
+        type: boolean
       upload_cache:
         required: true
         type: boolean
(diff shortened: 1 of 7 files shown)


libarchive.cmake
```diff
@@ -69,9 +69,8 @@ else()
     set(LibArchive_LIBRARIES archive_static)
     set(LibArchive_INCLUDE_DIRS ${THIRDPARTY_DIRECTORY}/libarchive/libarchive/)
```
angela28chen: are these still used or does this PR remove them? `LibArchive_LIBRARIES` and `LibArchive_INCLUDE_DIRS`

elviscapiaq (approved):

PR #723, opened 2026-01-21
Fix caching except for Windows

angela28chen:
The intention is to use weekdays scheduled runs (and the occasional manual kickoff) to create the cache, and presubmits can only restore those caches but not generate their own. Old caches will get deleted by Github's cache oldest-first deletion processes when the storage is full.

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -6,22 +6,15 @@ on:
       build_type:
         required: true
         type: string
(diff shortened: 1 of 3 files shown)


.github/workflows/build_windows.yml
hysw: Probably worth having both `restore_cache` and `upload_cache` as option

hysw (approved):

PR #728, opened 2026-01-26
Migrating build-related scripts to Python

angela28chen:
Choosing to implement these scripts in python instead for cross-platform support, human readability, and better support for complexity especially around flags. 
- Fixed a couple of minor operational logic bugs in the scripts, they should be more robust now
- `build_android.sh` and `build_android.bat` made into cross-platform `build_android.py`
  - Now accepts flag for `DIVE_RELEASE_TYPE`
- `deploy_mac_bundle_presubmit.sh` and `deploy_mac_bundle.sh` made into `deploy_mac_bundle.py`
  - Now called by the presubmit
- Added other CLIs to the host tools installion
  - These can be easily deleted from the official release build, if desired
- Update in-repo documentation

### DIFF
--- a/scripts/build_android.py
+++ b/scripts/build_android.py
@@ -104,7 +104,7 @@ def main(args):
     ]
     dive.echo_and_run(cmd)
 
-    # This step is necessary because cmake may not overwrite file of same name even if the contents have changed
+    # TODO: b/482108095 - This step is necessary because cmake may not overwrite file of same name even if the contents have changed
     if os.path.exists("build/pkg/device"):
         print("\nClearing install folder...")
         shutil.rmtree("build/pkg/device")


wangra-google (approved):

scripts/common_dive_utils.py
```diff
@@ -0,0 +1,38 @@
...
+        dive_root = my_env["DIVE_ROOT_PATH"]
+    else:
+        print("Predicting DIVE_ROOT_PATH")
+        dive_root = pathlib.Path(__file__).parent.parent.resolve()
+    
+    print("DIVE_ROOT_PATH={}".format(dive_root))
+    return dive_root
+
+def echo_and_run(cmd):
+    print("\n> {}".format(" ".join(cmd)))
```
footballhead: Consider f-strings. I personally think they tend to be more readable: it's more succinct and joining the template with the parameters reduces indirection

scripts/build_android.py
```diff
@@ -0,0 +1,109 @@
...
+        help="Description string for Device Libraries version info (default: dev)")
+    parser.add_argument("--build_type", default="Debug", choices=["Debug", "RelWithDebInfo", "Release"], 
+        help="Build type for Dive libraries " 
+        "(excluding GFXR gradle build which is always Debug) (default: Debug)")
+    parser.add_argument("--clean_build", action="store_true",
+        help="Clean device build folders before rebuilding")
+    return parser.parse_args()
+
+def main(args):
+    owd = os.getcwd()
```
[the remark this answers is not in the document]
angela28chen: "old working directory"? I'm fine to change it if you have a better alternative

scripts/deploy_mac_bundle.py
```diff
@@ -0,0 +1,89 @@
...
+        cmd = ["codesign", "--force", "--deep", "--sign", "-", "dive.app"]
+        dive.echo_and_run(cmd)
+
+    print("\nApplication now at {}/dive.app".format(os.getcwd()))
+
+    os.chdir(owd)
+    return
+
+if __name__ == "__main__":
+    start = timer()
```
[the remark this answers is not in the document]
angela28chen: I see your point about unintentional global variables, I'm going to try and address this with a class rather than context manager (mainly because I find the former easier to understand and I think it covers this case)

scripts/deploy_mac_bundle.py
```diff
@@ -0,0 +1,86 @@
...
+    dive.echo_and_run(cmd)
+
+    print("\nSetting up dive.app...")
+    # New dive.app was installed with host tools, move it and run macdeployqt
+    if os.path.exists("./host/dive.app"):
+        if os.path.exists("./dive.app"):
+            shutil.rmtree("./dive.app")
+        shutil.move("./host/dive.app", "./dive.app")
+
+        cmd = ["macdeployqt", "dive.app"]
```
footballhead: Pyhton recommends using fully qualified paths for executables, using `shutil.which` in order to find them

https://docs.python.org/3/library/subprocess.html#subprocess.Popen

scripts/build_android.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        prog="build_android",
+        description='This script automates the standard build process for Dive Device Libraries',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        "--dive_release_type", 
+        default="dev",
```
footballhead: Optional: should we validate that the value is one of the expected `<dev|canary|release|internal>`?

scripts/build_android.py
```diff
@@ -0,0 +1,128 @@
...
+        if os.path.exists("build/pkg/device"):
+            shutil.rmtree("build/pkg/device")
+
+
+    print("\nGenerating build files with cmake...")
+    cmd = [cmake_exec, ".", 
+        f"-DCMAKE_TOOLCHAIN_FILE={android_ndk_home}/build/cmake/android.toolchain.cmake",
+        "-GNinja Multi-Config",
+        "-Bbuild/device",
+        "-DCMAKE_MAKE_PROGRAM=ninja",
```
footballhead: nit: is this necessary if we ask for Ninja Multi-Config generator?

scripts/common_dive_utils.py
```diff
@@ -0,0 +1,51 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import pathlib
+import subprocess
+from timeit import default_timer
```
footballhead: nit: Google python style recommends using import for packages and modules, not classes https://google.github.io/styleguide/pyguide.html#22-imports

```suggestion
import timeit
```

then usage would look something like

```python
start = timeit.default_timer()
```

footballhead (approved):

wangra-google (approved):

PR #739, opened 2026-02-06
Refactor function to find Python 3 on path

angela28chen:


capture_service/device_mgr.cc
```diff
@@ -47,49 +47,27 @@ namespace Dive
 namespace
 {
 
-std::string GetPythonPath()
+absl::StatusOr<std::string> GetPython3Executable()
 {
     std::string python_path;
```
footballhead: This looks unused and can be removed
```suggestion
```

capture_service/device_mgr.cc
```diff
@@ -47,49 +47,27 @@ namespace Dive
...
-    absl::StatusOr<std::string> result = Dive::RunCommand("where python");
-    if (result.ok())
+    std::vector<std::string> python_names = {"python3", "python"};
+
+    for (const std::string& python_name : python_names)
     {
-        std::vector<std::string> lines = absl::StrSplit(*result, '\n');
-        for (const auto& line : lines)
+        std::string cmd = absl::StrFormat("%s --version", python_name);
+        absl::StatusOr<std::string> res = RunCommand(cmd);
```
footballhead: nit: prefer inlining single use variables https://abseil.io/tips/161

```suggestion
        absl::StatusOr<std::string> res = RunCommand(absl::StrFormat("%s --version", python_name));
```

capture_service/device_mgr.cc
```diff
@@ -47,49 +47,27 @@ namespace Dive
...
+        if (res.ok())
         {
-            std::string current_path = std::string(absl::StripAsciiWhitespace(line));
-            if (!current_path.empty() && !absl::StrContains(current_path, "WindowsApps"))
```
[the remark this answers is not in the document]
footballhead: +1 i had the same question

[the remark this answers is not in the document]
angela28chen: I think this might be a hardcoded check to avoid executable called python (installed at `C:\Users\<username>\AppData\Local\Microsoft\WindowsApps\python.exe` on my computer) which just launches the microsoft store popup to prompt you to install python. I find that if you run this with `--version` however, it doesn't cause a popup

capture_service/device_mgr.cc
angela28chen: Went with your suggestion about guard clause to reduce nesting so I don't think this applies anymore

wangra-google (approved):

footballhead (approved):

PR #752, opened 2026-02-13
Introduce absl logger for src/ and capture_service/

angela28chen:
- The intention is to eventually have all dive logging via the absl logger, but the first step is starting with these two directories
  - `capture_service/` is chosen because it logs some lines to adb logcat
- Rather than debug logging, using "INFO" severity, which will be seen in non-Debug builds
- The logging for denoting the begin/end of shorter functions has been removed, but I chose to keep it for the more complex ones.

Known Issues:
- The file/line reported for the helper functions in `log_utils.cpp` is that same file, when it could be the caller's info. Potentially addressed in the future with `std::source_location`
- When tested with direct PM4 capture flow, the name shown in logcat is "native" and not "dive_server" because the entry point is not the `server.cc` file which contains the setup. Potentially addressed in the future when `layer/` code is migrated to absl logger
- No way yet to control the verbosity of logs. Potentially addressed in the future with `absl::SetGlobalVLogLevel`

### DIFF
--- a/src/dive/log/android_log_sink.h
+++ b/src/dive/log/android_log_sink.h
@@ -29,7 +29,7 @@ class AndroidLogSink final : public absl::LogSink
 
     void Send(const absl::LogEntry& entry) override;
 
-    std::string_view GetAndroidTag() { return m_android_tag; }
+    std::string_view GetAndroidTag() const { return m_android_tag; }
     void SetAndroidTag(std::string_view android_tag) { m_android_tag = android_tag; }
 
  private:


angela28chen:
My plan is to file internal bugs for outstanding work after this PR is reviewed and merged

src/dive/log/android_log_sink.h
```diff
@@ -0,0 +1,35 @@
...
+{
+
+class AndroidLogSink final : public absl::LogSink
+{
+ public:
+    ~AndroidLogSink() override = default;
+
+    void Send(const absl::LogEntry& entry) override;
+
+    std::string m_android_native_tag = "native";
```
wangra-google: why is this member public?

capture_service/android_application.cc
```diff
@@ -151,8 +151,7 @@ absl::Status AndroidApplication::ParsePackage()
 
 absl::Status AndroidApplication::Cleanup()
 {
-    LOGI("%s AndroidApplication::Cleanup(): package %s\n", Dive::kLogPrefixCleanup,
-         m_package.c_str());
+    LOG(INFO) << "##### AndroidApplication::Cleanup() package " << m_package << " started";
```
hysw: Do we need `#####`?

src/dive/log/android_log_sink.h
angela28chen: The intent was for the other class `AbslLogger` to be able to set its value during `AbslLogger::Init()`

capture_service/dive_client_cli.cc
```diff
@@ -234,11 +237,11 @@ std::string GenerateAppTypeFlagHelp()
 // Waits for user input before exiting.
 absl::Status WaitForExitConfirmation()
 {
-    std::cout << "Press any key+enter to exit" << std::endl;
+    LOG(INFO) << "Press any key+enter to exit";
```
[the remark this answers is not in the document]
angela28chen: I'm not sure what determines if it should go to std::cout or the logger. I made some changes, PTAL and let me know if anything stands out to you as wrong.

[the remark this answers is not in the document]
hysw: Probably fine.

The main determination is if you pretend all the logger message does not exist (e.g. `2>/dev/null`), you can still interactive with the CLI.

hysw (approved):

PR #759, opened 2026-02-19
Adds What-If Dialogs

GrantComm:
Adds What-If Dialogs for setup and What-If modification configurations. Follow-up prs will introduce What-If tab view to view added modifications.

**- What-If Icon**

Screenshot 2026-02-19 at 12 34 36 PM

**- What-If Setup Dialog**

Screenshot 2026-02-19 at 12 34 52 PM

**- What-If Section in Main Window**

Screenshot 2026-02-19 at 12 35 33 PM

**- What-If Configure Dialog**

Screenshot 2026-02-19 at 12 36 01 PM

Manual Testing:
Linux:
- Click what-if button, connect to a device, select a package, close the dialog, ensure the dialog settings are preserved
- Click what-if button, connect to a device, select a package, close the dialog, disconnect the device, ensure the dialog settings are cleared
- Click what-if button, close the dialog ensure the start application button is still disabled
- Click what-if button, start application, ensure the what-if configure button is visible
- Click what-if button, start application, open the what-if modification configure dialog

### DIFF
--- a/ui/main_window.h
+++ b/ui/main_window.h
@@ -74,7 +74,7 @@ class MainWindow : public QMainWindow
     void OnPendingPerfCounterResults(const QString& file_name);
     void OnPendingGpuTimingResults(const QString& file_name);
     void OnPendingScreenshot(const QString& file_name);
-    void OnConfigureWhatIfModifcation();
+    void OnConfigureWhatIfModification();
     void OnWhatIfSetupTrigger();
     void OnWhatIfRuntimeEnabled(const QString& package_name, bool is_runtime_what_if_enabled);
 
@@ -155,6 +155,15 @@ class MainWindow : public QMainWindow
     void ResetVerticalScroll(const DiveTreeView& tree_view);
     void ClearViewModelSelection(DiveTreeView& tree_view, bool should_clear_tab);
     void CorrelateCounter(const QModelIndex& index, bool called_from_gfxr_view);
+    void SetupRuntimeWhatIfHeader();
+
+    struct RuntimeWhatIfHeader
+    {
+        QWidget* container = nullptr;
+        QPushButton* configure_button = nullptr;
+        QPushButton* stop_app_button = nullptr;
+        QLabel* app_name_label = nullptr;
+    };
 
     std::optional<uint64_t> GetDrawCallIndexFromProxyIndex(
         const QModelIndex& proxy_index, const QAbstractProxyModel& proxy_model,
@@ -207,10 +216,7 @@ class MainWindow : public QMainWindow
     CommandModel* m_command_hierarchy_model;
     QPushButton* m_search_trigger_button;
     SearchBar* m_event_search_bar = nullptr;
-    QWidget* m_what_if_container = nullptr;
-    QPushButton* m_what_if_configure_button = nullptr;
-    QPushButton* m_what_if_run_time_stop_application_button = nullptr;
-    QLabel* m_what_if_runtime_application_name_label = nullptr;
+    RuntimeWhatIfHeader m_runtime_what_if_header = {};
 
     TreeViewComboBox* m_view_mode_combo_box;
     TreeViewComboBox* m_filter_mode_combo_box;
--- a/ui/what_if_configure_dialog.cpp
+++ b/ui/what_if_configure_dialog.cpp
@@ -149,10 +149,10 @@ WhatIfConfigureDialog::WhatIfConfigureDialog(QWidget* parent) : DeviceDialog(par
     main_layout->addLayout(CreateButtonLayout());
     setLayout(main_layout);
 
-    m_specific_settings_container->hide();
     m_command_box->setEnabled(false);
 
     SetupConnections();
+    OnConfigValuesChanged();
 }
 
 QVBoxLayout* WhatIfConfigureDialog::CreateHeaderLayout()
@@ -352,7 +352,6 @@ QWidget* WhatIfConfigureDialog::SetupFlagContainer()
     flag_layout->addWidget(flag_label);
     flag_layout->addWidget(m_flag_box);
 
-    container->hide();
     return container;
 }
 
@@ -557,15 +556,13 @@ void WhatIfConfigureDialog::UpdateVisibility()
     // Based on the modification type, check if the required fields are filled. At least one filter
     // or flag should be selected.
     int type_index = m_type_box->currentIndex();
+    HideSpecificSettings();
     if (type_index <= 0)
     {
-        HideSpecificSettings();
         this->adjustSize();
         return;
     }
 
-    HideSpecificSettings();
-    m_specific_settings_container->show();
     const auto& modification_type_info = Dive::kWhatIfTypeInfos[type_index - 1];
 
     switch (modification_type_info.type)
@@ -579,24 +576,17 @@ void WhatIfConfigureDialog::UpdateVisibility()
             QString command_string = m_command_box->currentText();
             if (command_string.contains("Indirect", Qt::CaseInsensitive))
             {
-                m_draw_count_filter.label->show();
-                m_draw_count_filter.spin_box->show();
+                SetDrawCallFilterSpinnerVisibility(m_draw_count_filter, /*is_visible=*/true);
             }
             else if (command_string.contains("Indexed", Qt::CaseInsensitive))
             {
-                m_index_count_filter.label->show();
-                m_index_count_filter.spin_box->show();
-
-                m_instance_count_filter.label->show();
-                m_instance_count_filter.spin_box->show();
+                SetDrawCallFilterSpinnerVisibility(m_index_count_filter, /*is_visible=*/true);
+                SetDrawCallFilterSpinnerVisibility(m_instance_count_filter, /*is_visible=*/true);
             }
             else
             {
-                m_vertex_count_filter.label->show();
-                m_vertex_count_filter.spin_box->show();
-
-                m_instance_count_filter.label->show();
-                m_instance_count_filter.spin_box->show();
+                SetDrawCallFilterSpinnerVisibility(m_vertex_count_filter, /*is_visible=*/true);
+                SetDrawCallFilterSpinnerVisibility(m_instance_count_filter, /*is_visible=*/true);
             }
             this->adjustSize();
             break;
@@ -616,7 +606,6 @@ void WhatIfConfigureDialog::UpdateVisibility()
             m_modification_warning_label->show();
             break;
         default:
-            m_specific_settings_container->hide();
             qDebug() << absl::StrCat("The case ", modification_type_info.ui_name,
                                      " is unimplemented/unsupported")
                             .c_str();
@@ -660,8 +649,6 @@ void WhatIfConfigureDialog::OnWhatIfModificationTypeChanged(int index)
 //--------------------------------------------------------------------------------------------------
 void WhatIfConfigureDialog::HideSpecificSettings()
 {
-    m_specific_settings_container->hide();
-
     const auto& children = m_specific_settings_container->children();
     for (QObject* child : children)
     {
@@ -685,10 +672,10 @@ void WhatIfConfigureDialog::ResetDrawCallFilters()
 
 void WhatIfConfigureDialog::HideDrawCallFilters()
 {
-    HideDrawCallFilterSpinner(m_draw_count_filter);
-    HideDrawCallFilterSpinner(m_index_count_filter);
-    HideDrawCallFilterSpinner(m_instance_count_filter);
-    HideDrawCallFilterSpinner(m_vertex_count_filter);
+    SetDrawCallFilterSpinnerVisibility(m_draw_count_filter, /*is_visible=*/false);
+    SetDrawCallFilterSpinnerVisibility(m_index_count_filter, /*is_visible=*/false);
+    SetDrawCallFilterSpinnerVisibility(m_instance_count_filter, /*is_visible=*/false);
+    SetDrawCallFilterSpinnerVisibility(m_vertex_count_filter, /*is_visible=*/false);
 }
 
 void WhatIfConfigureDialog::CreateDrawCallFilterSpinner(DrawCallFilterSpinner& filter,
@@ -699,12 +686,13 @@ void WhatIfConfigureDialog::CreateDrawCallFilterSpinner(DrawCallFilterSpinner& f
     filter.spin_box->setRange(min, max);
 }
 
-void WhatIfConfigureDialog::HideDrawCallFilterSpinner(DrawCallFilterSpinner& filter)
+void WhatIfConfigureDialog::SetDrawCallFilterSpinnerVisibility(DrawCallFilterSpinner& filter,
+                                                               bool is_visible)
 {
-    if (filter.label) filter.label->hide();
+    if (filter.label) filter.label->setVisible(is_visible);
     if (filter.spin_box)
     {
-        filter.spin_box->hide();
+        filter.spin_box->setVisible(is_visible);
     }
 }
 
--- a/ui/what_if_configure_dialog.h
+++ b/ui/what_if_configure_dialog.h
@@ -86,7 +86,7 @@ class WhatIfConfigureDialog : public DeviceDialog
     void UpdateAddModificationButtonState();
     void CreateDrawCallFilterSpinner(DrawCallFilterSpinner& filter, const QString& label_text,
                                      int min, int max);
-    void HideDrawCallFilterSpinner(DrawCallFilterSpinner& filter);
+    void SetDrawCallFilterSpinnerVisibility(DrawCallFilterSpinner& filter, bool is_visible);
 
     // --- What-If Type Section ---
     QComboBox* m_type_box = nullptr;
--- a/ui/main_window.cpp
+++ b/ui/main_window.cpp
@@ -213,21 +213,7 @@ MainWindow::MainWindow(ApplicationController& controller) : m_controller(control
     {
         QVBoxLayout* left_vertical_layout = new QVBoxLayout();
 
-        QHBoxLayout* what_ifs_layout = new QHBoxLayout();
-        m_what_if_container = new QWidget();
-        m_what_if_container->setLayout(what_ifs_layout);
-        QLabel* what_if_info_label = new QLabel(tr("What-If Analysis:"));
-        m_what_if_configure_button = new QPushButton(tr("Configure"));
-        m_what_if_run_time_stop_application_button = new QPushButton(tr("Stop Application"));
-        QLabel* what_if_runtime_what_if_application_label = new QLabel(tr("Current Application: "));
-        m_what_if_runtime_application_name_label = new QLabel(tr("None"));
-        what_ifs_layout->addWidget(what_if_info_label);
-        what_ifs_layout->addWidget(m_what_if_configure_button);
-        what_ifs_layout->addWidget(m_what_if_run_time_stop_application_button);
-        what_ifs_layout->addStretch(1);
-        what_ifs_layout->addWidget(what_if_runtime_what_if_application_label);
-        what_ifs_layout->addWidget(m_what_if_runtime_application_name_label);
-        m_what_if_container->hide();
+        SetupRuntimeWhatIfHeader();
 
         QFrame* text_combo_box_frame = new QFrame();
 
@@ -344,7 +330,7 @@ MainWindow::MainWindow(ApplicationController& controller) : m_controller(control
             expand_to_lvl_layout->addWidget(expand_to_lvl_button);
         expand_to_lvl_layout->addStretch();
 
-        left_vertical_layout->addWidget(m_what_if_container);
+        left_vertical_layout->addWidget(m_runtime_what_if_header.container);
         left_vertical_layout->addWidget(m_event_search_bar);
         left_vertical_layout->addWidget(text_combo_box_frame);
         left_vertical_layout->addWidget(m_command_hierarchy_view);
@@ -581,10 +567,10 @@ MainWindow::MainWindow(ApplicationController& controller) : m_controller(control
     // What-If connections
     QObject::connect(m_what_if_setup_dig, &WhatIfSetupDialog::RuntimeWhatIfEnabled, this,
                      &MainWindow::OnWhatIfRuntimeEnabled);
-    QObject::connect(m_what_if_run_time_stop_application_button, &QPushButton::clicked,
+    QObject::connect(m_runtime_what_if_header.stop_app_button, &QPushButton::clicked,
                      m_what_if_setup_dig, &WhatIfSetupDialog::OnStopRuntimeWhatIf);
-    QObject::connect(m_what_if_configure_button, &QPushButton::clicked, this,
-                     &MainWindow::OnConfigureWhatIfModifcation);
+    QObject::connect(m_runtime_what_if_header.configure_button, &QPushButton::clicked, this,
+                     &MainWindow::OnConfigureWhatIfModification);
 
     CreateActions();
     CreateMenus();
@@ -1279,18 +1265,18 @@ void MainWindow::OnWhatIfRuntimeEnabled(const QString& package_name,
 {
     if (is_runtime_what_if_enabled)
     {
-        m_what_if_runtime_application_name_label->setText(package_name);
-        m_what_if_container->show();
+        m_runtime_what_if_header.app_name_label->setText(package_name);
+        m_runtime_what_if_header.container->show();
     }
     else
     {
-        m_what_if_runtime_application_name_label->setText("None");
-        m_what_if_container->hide();
+        m_runtime_what_if_header.app_name_label->setText("None");
+        m_runtime_what_if_header.container->hide();
     }
 }
 
 //--------------------------------------------------------------------------------------------------
-void MainWindow::OnConfigureWhatIfModifcation() { m_what_if_configure_dig->show(); }
+void MainWindow::OnConfigureWhatIfModification() { m_what_if_configure_dig->show(); }
 
 //--------------------------------------------------------------------------------------------------
 void MainWindow::OnExpandToLevel()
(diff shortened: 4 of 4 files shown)


src/dive/common/what_if_modification_types.h
```diff
@@ -0,0 +1,76 @@
...
+
+struct WhatIfModificationTypeInfo
+{
+    WhatModificationType type;
+    std::string_view ui_name;
+    std::string_view ui_name_short;
+    std::span<const std::string_view> supported_commands;
+};
+
+static constexpr std::string_view kDrawCmds[] = {"vkCmdDraw", "vkCmdDrawIndexed",
```
wangra-google: you should `inline constexpr` those global consts (instead of `static constexpr`) to avoid local copies.

src/dive/utils/device_resources_constants.h
```diff
@@ -41,6 +41,7 @@ Profiling resources dir:    DIVE_ROOT/build/pkg/CMAKE_GENERATED_PROFILING_PLUGIN
 
     // Device resources:
     static constexpr std::string_view kVkLayerLibName = "libVkLayer_dive.so";
+    static constexpr std::string_view kVkRuntimeLayerLibName = "libVkLayer_rt_dive.so";
```
wangra-google: better to change all "static constexpr" here to "inline constexpr" to avoid extra copies

ui/main_window.h
```diff
@@ -200,6 +207,12 @@ class MainWindow : public QMainWindow
     CommandModel* m_command_hierarchy_model;
     QPushButton* m_search_trigger_button;
     SearchBar* m_event_search_bar = nullptr;
+    QWidget* m_what_if_container;
```
wangra-google: i know this is a bit of a historical reason, but maybe at least for newly added members, we can do in-class member initialization in the headers. (we could probably have a cleanup pass later when we have time to change the existing ones)

ui/what_if_configure_dialog.cpp
```diff
@@ -0,0 +1,690 @@
...
+#include "application_controller.h"
+#include "capture_service/device_mgr.h"
+#include "dive/common/what_if_modification_types.h"
+
+namespace
+{
+constexpr int kNumImageCreationFlags = 2;
+constexpr int kNumRenderPassTypes = 4;
+
+constexpr const char* kFrameTitleStrings[kNumImageCreationFlags] = {
```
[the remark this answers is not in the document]
GrantComm: Done

ui/what_if_configure_dialog.h
```diff
@@ -0,0 +1,139 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+#include "device_dialog.h"
+#include "dive/ui/forward.h"
+
+#pragma once
```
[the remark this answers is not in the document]
GrantComm: Done

src/dive/utils/device_resources_constants.h
GrantComm: Per the conversation in #761, leaving this as is

src/dive/common/app_types.h
```diff
@@ -38,6 +38,7 @@ struct AppTypeInfo
     std::string_view ui_name;
     std::string_view cli_name;
     bool is_gfxr_capture_supported;
+    bool is_what_if_supported;
```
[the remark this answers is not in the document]
GrantComm: Set the default to true and set to false for the gles applications

capture_service/android_application.cc
```diff
@@ -355,6 +365,53 @@ absl::Status VulkanApplication::Cleanup()
...
+        /*package=*/m_package,
+        /*file_name=*/Dive::DeviceResourcesConstants::kVkRuntimeLayerLibName));
+
+    RETURN_IF_ERROR(m_dev.Adb().Run("shell settings put global enable_gpu_debug_layers 1"));
+    RETURN_IF_ERROR(
+        m_dev.Adb().Run(absl::StrFormat("shell settings put global gpu_debug_app %s", m_package)));
+    RETURN_IF_ERROR(m_dev.Adb().Run(
+        absl::StrFormat("shell settings put global gpu_debug_layer_app %s", m_package)));
+    RETURN_IF_ERROR(m_dev.Adb().Run(
+        absl::StrFormat("shell settings put global gpu_debug_layers %s", Dive::kVkLayerName)));
```
elviscapiaq: Thanks for fixing this, the runtime_layer was not loaded correctly in the previous PR.

capture_service/device_mgr.cc
```diff
@@ -513,7 +513,13 @@ absl::Status AndroidDevice::ForwardFirstAvailablePort()
 
 absl::Status AndroidDevice::SetupDevice()
 {
+    if (m_runtime_what_if_enabled)
```
elviscapiaq: I tested this part and in order to connect a Tcp Client with the server, we have to forward an available port to establish the connection between PC and device.
It can be fixed with:
```
if (m_runtime_what_if_enabled)
    {
        RETURN_IF_ERROR(ForwardFirstAvailablePort());
        return absl::OkStatus();
    }
```

src/dive/common/what_if_modification_types.h
```diff
@@ -0,0 +1,75 @@
...
+
+#pragma once
+
+#include <array>
+#include <string_view>
+
+namespace Dive
+{
+
+enum class WhatModificationType
```
angela28chen: suggestion: `WhatIfType`

src/dive/common/what_if_modification_types.h
```diff
@@ -0,0 +1,75 @@
...
+
+struct WhatIfModificationTypeInfo
+{
+    WhatModificationType type;
+    std::string_view ui_name;
+    std::string_view ui_name_short;
+    std::span<const std::string_view> supported_commands;
+};
+
+inline constexpr std::string_view kDrawCmds[] = {"vkCmdDraw", "vkCmdDrawIndexed",
```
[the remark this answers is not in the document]
wangra-google: Good point. Agree that kDrawCmds is too generic for the top-level dive namespace.
Maybe we could put all of those constexpr into a nested namespace, something like 

namespace dive::what_if {
    inline constexpr std::string_view kDrawCmds[] = { ... };
}

ui/main_window.h
```diff
@@ -200,6 +207,10 @@ class MainWindow : public QMainWindow
     CommandModel* m_command_hierarchy_model;
     QPushButton* m_search_trigger_button;
     SearchBar* m_event_search_bar = nullptr;
+    QWidget* m_what_if_container = nullptr;
+    QPushButton* m_what_if_configure_button = nullptr;
+    QPushButton* m_what_if_run_time_stop_application_button = nullptr;
+    QLabel* m_what_if_runtime_what_if_application_name_label = nullptr;
```
angela28chen: nit: was it intentional to have `what_if` substring twice in the name here?

ui/what_if_setup_dialog.cpp
```diff
@@ -0,0 +1,556 @@
...
+        {
+            return;
+        }
+    }
+
+    emit RuntimeWhatIfEnabled("None", false);
+    ResetDialog();
+}
+
+void WhatIfSetupDialog::ShowRuntimeWhatIfFields()
```
angela28chen: For this group of objects and the ones for replay, is it possible to make parent widgets for "runtime options" and "replay options" and then show/hide the parent widget?

ui/what_if_setup_dialog.cpp
```diff
@@ -0,0 +1,556 @@
...
+        ShowMessage(QString("Please select application type"));
+        return;
+    }
+
+    QModelIndex proxyModelIndex = m_app_type_filter_model->index(proxy_index, 0);
+    QModelIndex sourceModelIndex = m_app_type_filter_model->mapToSource(proxyModelIndex);
+    int source_row = sourceModelIndex.row();
+
+    std::string ty_str = Dive::kAppTypeInfos[source_row].ui_name.data();
+    if (m_start_application_button->text() == kStart_Application.data())
```
angela28chen: To reduce nesting with `if...else...`, make this a `!=` check and have an early return for if the stop application was clicked.

ui/what_if_setup_dialog.cpp
```diff
@@ -0,0 +1,556 @@
...
+    else
+    {
+        qDebug() << "Stop Application clicked for package: " << m_cur_pkg;
+        StopPackage();
+    }
+}
+
+void WhatIfSetupDialog::OnWhatIfTypeChanged(int button_id)
+{
+    m_runtime_what_if_enabled = (button_id == kRuntimeWhatIfButtonId);
```
angela28chen: private member only used in one method, could be local var instead?

src/dive/common/what_if_modification_types.h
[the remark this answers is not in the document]
angela28chen: Anon namespace is fine IMO, nested namespaces are hard to remember/find sometimes so I think thats why we tended to avoid them and use Dive namespace?

ui/what_if_setup_dialog.h
```diff
@@ -0,0 +1,109 @@
...
+    void StopPackage();
+    void ShowRuntimeWhatIfFields();
+    void ShowReplayWhatIfFields();
+
+    void OnDeviceSelected() override;
+    void OnDeviceSelectionCleared() override;
+
+    ApplicationController& m_controller;
+
+    static constexpr std::string_view kStart_Application = "&Start Application";
```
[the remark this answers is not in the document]
wangra-google: +1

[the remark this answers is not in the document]
wangra-google: could be put in anonymous namespace in the cpp file
also to be consistent with the other constexpr, better to remove the _ in the middle, something like kStartApplication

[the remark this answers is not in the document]
GrantComm: Done

ui/what_if_setup_dialog.h
```diff
@@ -0,0 +1,109 @@
...
+
+    QLabel* m_app_type_label = nullptr;
+    WhatIfAppTypeFilterModel* m_app_type_filter_model = nullptr;
+    QComboBox* m_app_type_box = nullptr;
+
+    QPushButton* m_start_application_button = nullptr;
+
+    std::vector<std::string> m_pkg_list;
+    QString m_cur_pkg;
+    std::string m_executable;
```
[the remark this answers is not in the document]
GrantComm: Removed

ui/what_if_setup_dialog.h
```diff
@@ -0,0 +1,109 @@
...
+    QLabel* m_args_label = nullptr;
+    QLineEdit* m_args_input_box = nullptr;
+
+    QLabel* m_app_type_label = nullptr;
+    WhatIfAppTypeFilterModel* m_app_type_filter_model = nullptr;
+    QComboBox* m_app_type_box = nullptr;
+
+    QPushButton* m_start_application_button = nullptr;
+
+    std::vector<std::string> m_pkg_list;
```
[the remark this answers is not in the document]
GrantComm: Created the RuntimeData struct

ui/what_if_setup_dialog.h
```diff
@@ -0,0 +1,111 @@
...
+    void ResetDialog();
+    bool StartPackage(Dive::AndroidDevice* device, const std::string& app_type);
+    void StopPackage();
+    void ShowRuntimeWhatIfFields();
+    void ShowReplayWhatIfFields();
+
+    void OnDeviceSelected() override;
+    void OnDeviceSelectionCleared() override;
+
+    ApplicationController& m_controller;
```
angela28chen: I don't quite understand the purpose for this, is it used anywhere?

ui/what_if_setup_dialog.h
```diff
@@ -0,0 +1,111 @@
...
+
+    QRadioButton* m_runtime_what_if_type_button = nullptr;
+    QRadioButton* m_replay_what_if_type_button = nullptr;
+
+    QWidget* m_runtime_options_widget = nullptr;
+    QWidget* m_replay_options_widget = nullptr;
+
+    QLabel* m_device_label = nullptr;
+    QPushButton* m_device_refresh_button = nullptr;
```
angela28chen: `m_device_refresh_button` seems to only be used in the constructor, does it need to be a private member?

ui/what_if_setup_dialog.h
GrantComm: Removed

ui/what_if_setup_dialog.cpp
```diff
@@ -0,0 +1,571 @@
...
+    {
+        std::string err_msg =
+            absl::StrFormat("Fail to setup for package %s, error: %s",
+                            m_runtime_data.m_cur_pkg.toStdString(), ret.message());
+        qDebug() << err_msg.c_str();
+        ShowMessage(QString::fromStdString(err_msg));
+        return false;
+    }
+    ret = device->StartApp();
+    if (!ret.ok())
```
angela28chen: nit: this part could be inlined
```
if (absl::Status ret = device->StartApp(); !ret.ok())
```

ui/what_if_setup_dialog.h
GrantComm: Fixed, sorry, missed this.

ui/what_if_configure_dialog.cpp
```diff
@@ -0,0 +1,712 @@
...
+    }
+    m_what_if_command_box->setModel(m_what_if_command_model);
+
+    if (ty_info.type == Dive::WhatIfType::kDrawCallDisabled)
+    {
+        ShowDrawCallFields();
+    }
+    else if (ty_info.type == Dive::WhatIfType::kRenderPassScissorOverridden)
+    {
+        ShowRenderPassFields();
```
angela28chen: I'm wondering if you can just avoid all the `Show*()` methods (and the hassle of having to create a new method every time there's a new enum field) by just having things inlined here. 

1. If you're worried about bloating this function, the fields will presumably be on a widget/layout that you can show/hide with one line, so I'm not concerned about bloat
2. I think it's still possible to handle cases like `kRenderPassScissorOverridden` and `kRenderPassLoadStoreOpOverridden` without duplicating code, just have something like this:

```
// I probably have the wrong syntax
case kRenderPassLoadStoreOpOverridden:
{
  m_what_if_modification_warning_label->show();
  [[fallthrough]];
}
case kRenderPassScissorOverridden:
{
  // contents of ShowRenderPassFields()
  break;
}
```
And it's also doable with the whole if/else if structure by using nesting, but personally I feel like the switch is neater.

ui/what_if_configure_dialog.cpp
```diff
@@ -0,0 +1,712 @@
...
+        m_what_if_modification_warning_label->show();
+    }
+    else if (ty_info.type == Dive::WhatIfType::kAnisotropicFilterDisabled)
+    {
+        m_what_if_modification_warning_label->show();
+    }
+}
+
+//--------------------------------------------------------------------------------------------------
+void WhatIfConfigureDialog::HideAllFields()
```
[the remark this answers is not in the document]
GrantComm: I changed it to HideSpecificWhatIfSettings. The container is hidden, but the container's children also need to be hidden. When the parent is shown again, the children are still visible to the parent, which is why the child widgets need to be hidden as well. Then depending on the modification type chosen, the appropriate children are shown

https://doc.qt.io/archives/qt-5.15/qwidget.html#visible-prop

ui/what_if_configure_dialog.h
```diff
@@ -0,0 +1,125 @@
...
+    void SetupDrawCallFiltersContainer();
+    void SetupRenderPassFiltersContainer();
+    void SetupFlagContainer();
+    QHBoxLayout* CreateButtonLayout();
+    void SetupConnections();
+
+ protected:
+    void closeEvent(QCloseEvent* event) override;
+
+ private slots:
```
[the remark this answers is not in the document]
GrantComm: We have a mix in the dive codebase

ui/what_if_setup_dialog.cpp
```diff
@@ -0,0 +1,571 @@
...
+
+void WhatIfSetupDialog::UpdatePackageList()
+{
+    auto device = Dive::GetDeviceManager().GetDevice();
+    if (device == nullptr)
+    {
+        return;
+    }
+
+    if (auto ret = device->ListPackage(m_runtime_data.m_pkg_list_options); !ret.ok())
```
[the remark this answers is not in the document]
GrantComm: It is auto since ListPackage is return type absl::StatusOr

ui/what_if_setup_dialog.cpp
```diff
@@ -0,0 +1,571 @@
...
+        std::string err_msg =
+            absl::StrFormat("Fail to start package %s, error: %s",
+                            m_runtime_data.m_cur_pkg.toStdString(), ret.message());
+        qDebug() << err_msg.c_str();
+        ShowMessage(QString::fromStdString(err_msg));
+        return false;
+    }
+    auto cur_app = device->GetCurrentApplication();
+
+    if (!cur_app->IsRunning())
```
angela28chen: Do we need to check if cur_app is not nullptr before using it?

Then the below `if (cur_app)` check can also be omitted

ui/what_if_configure_dialog.h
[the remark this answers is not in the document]
angela28chen: OK, fine to leave as-is

ui/what_if_configure_dialog.h
```diff
@@ -0,0 +1,121 @@
...
+    QComboBox* m_what_if_command_box = nullptr;
+    QStandardItemModel* m_what_if_command_model = nullptr;
+
+    // --- Specific Settings Container ---
+    QWidget* m_specific_what_if_settings_container = nullptr;
+
+    // --- Filter Section ---
+    QLabel* m_what_if_filter_label = nullptr;
+
+    // --- Draw Call Filters ---
```
angela28chen: This is really repetitive and unwieldy with the long member names.

I suggest you make a struct (`DrawCallFilter`) that's just going to hold a QLabel* and QSpinBox*, and provide some helper functions that take the struct as input. I think that would make the configure dialog code and this large list of members clearer.

Something like:
- `WhatIfConfigureDialog::CreateDrawCallFilter()`: input args: the struct, the label name, the spinner range. It does the initialization of widgets
- `WhatIfConfigureDialog::HideDrawCallFilter()`: input args: the struct. This is just for convenience
- You can access the label and spinner separately to:
  - add to the layout grid in `WhatIfConfigureDialog::SetupDrawCallFiltersContainer()`
  - show the widget

angela28chen: Maybe `DrawCallFilterSpinner` since there seems to be other combo box ones

ui/what_if_setup_dialog.cpp
```diff
@@ -0,0 +1,571 @@
...
+    EnableWhatIfTypeButtons(false);
+
+    absl::Status ret;
+    std::string device_serial = GetCurrentDeviceSerial();
+    qDebug() << "Start app on dev: " << device_serial.c_str()
+             << ", package: " << m_runtime_data.m_cur_pkg.toStdString().c_str()
+             << ", type: " << app_type.c_str()
+             << ", args: " << m_runtime_data.m_command_args.c_str();
+
+    if (app_type ==
```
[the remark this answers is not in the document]
angela28chen: Thanks

ui/what_if_configure_dialog.cpp
```diff
@@ -0,0 +1,685 @@
...
+            m_draw_call_pso_property_filter_box->currentIndex() == 0 &&
+            m_draw_call_render_pass_filter_box->currentIndex() == 0)
+        {
+            m_add_modification_button->setEnabled(false);
+            return;
+        }
+    }
+    else if (modification_type_info.type == Dive::WhatIfType::kImageCreationFlagRemoved)
+    {
+        if (m_flag_box->isVisible())
```
angela28chen: Is it possible for the modification type to be set to `kImageCreationFlagRemoved` and yet the flag box is not visible? I'm surprised we have this check here. I feel like if that's not a normal state we can just remove this check in the same way that the other if-else clauses in this section don't check if their related UI elements are visible.

ui/what_if_configure_dialog.cpp
```diff
@@ -0,0 +1,685 @@
...
+    filter.spin_box->setRange(min, max);
+}
+
+void WhatIfConfigureDialog::HideDrawCallFilterSpinner(DrawCallFilterSpinner& filter)
+{
+    if (filter.label) filter.label->hide();
+    if (filter.spin_box)
+    {
+        filter.spin_box->hide();
+        filter.spin_box->setValue(0);
```
angela28chen: I'm wondering, is it a good idea to be resetting the value here rather than letting it persist even when the spinner is hidden? We have all these spinners hooked up to slot `OnUpdateAddModificationButtonState()`. Does it mean that when there is a single `HideDrawCallFilterSpinner()` call, it will create up to N signals (for N nonzero spin boxes)? Do we need that?

What about just omitting the `setValue` call? After all, when the add modification button is clicked, it should only take into account the current modification type that is specified (which ought to have its widgets visible to the user). I don't think we risk the users adding unseen modifications.

ui/what_if_configure_dialog.cpp
```diff
@@ -0,0 +1,685 @@
...
+
+// =================================================================================================
+// MultiCheckComboBoxEventFilter
+// =================================================================================================
+MultiCheckComboBoxEventFilter::MultiCheckComboBoxEventFilter(QComboBox* parent)
+    : QObject(parent), combo_box(parent)
+{
+}
+
+bool MultiCheckComboBoxEventFilter::eventFilter(QObject* watched, QEvent* event)
```
angela28chen: This function contains deep nesting that makes it hard to read and error-prone. Maybe you can use a combination of having a helper function full of early returns (`abslStatusOr<bool> customEventFilter()`) and making this `eventFilter()` function call it. If there was an error, just report/suppress the error message and then fall back on `return QObject::eventFilter(watched, event);`. But if `customEventFilter()` returns a valid bool just immediately return that.

Also another opportunity to use a switch statement on `event->type()` enum, which could let you use some `break` to reduce nesting + handle the default cases and log debug messages for unexpected events.

Reference:
- https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

ui/what_if_setup_dialog.cpp
```diff
@@ -0,0 +1,582 @@
...
+    if (no_application_currently_running_on_device)
+    {
+        EnableWhatIfTypeButtons(true);
+        m_runtime_what_if_type_button->setChecked(true);
+        OnWhatIfTypeChanged(kRuntimeWhatIfButtonId);
+        m_start_application_button->setText(kStartApplication.data());
+    }
+
+    event->accept();
+    return;
```
angela28chen: nit: not necessary?

ui/what_if_setup_dialog.h
```diff
@@ -0,0 +1,115 @@
...
+    void OnStartClicked();
+    void OnWhatIfTypeChanged(int id);
+    void ShowMessage(const QString& message) override;
+
+ signals:
+    void RuntimeWhatIfEnabled(const QString& package_name, bool is_runtime_what_if_enabled);
+
+ private:
+    void ResetDialog();
+    bool StartPackage(Dive::AndroidDevice* device, const std::string& app_type);
```
angela28chen: not clear what this bool return value means. I suggest including the absl status header and making this return value into status instead.

ui/what_if_configure_dialog.h
```diff
@@ -0,0 +1,124 @@
...
+    MultiCheckComboBoxEventFilter& operator=(const MultiCheckComboBoxEventFilter&) = delete;
+
+ protected:
+    bool eventFilter(QObject* watched, QEvent* event) override;
+
+ private:
+    QComboBox* combo_box;
+};
+
+class WhatIfConfigureDialog : public DeviceDialog
```
angela28chen: Reading the implementation, a lot of the methods of this class do more than one task & it's not necessarily matching the name. I suggest:
- Adding comments in the line above the important/complex functions declarations here which gives the reader an idea of them without having to read the .cpp file. 
- Making sure each function's purpose is well-defined. For example, a function called `OnWhatIfModificationTypeChanged()` shouldn't be touching `m_add_modification_button` when you have `OnUpdateAddModificationButtonState()`. Either the first one signals the second, or (recommended) higher up in the flow the signal from `m_type_box` is also hooked up to `OnUpdateAddModificationButtonState()`.

ui/what_if_configure_dialog.cpp
```diff
@@ -0,0 +1,725 @@
...
+absl::StatusOr<bool> CustomEventFilter(QComboBox* combo_box, QObject* watched, QEvent* event)
+{
+    if (!combo_box || watched != combo_box->view()->viewport())
+    {
+        return absl::FailedPreconditionError("Not the watched viewport");
+    }
+
+    switch (event->type())
+    {
+        case QEvent::MouseButtonPress:
```
angela28chen: I'm a little confused, is the code now identical for the two cases? If so then you can stack it like this to avoid duplication:

```
  case QEvent::MouseButtonPress:
  case QEvent::MouseButtonRelease:
  {
    ...
  }
```

ui/what_if_configure_dialog.cpp
```diff
@@ -0,0 +1,716 @@
...
+        case Dive::WhatIfType::kDrawCallDisabled:
+        {
+            m_filter_label->show();
+            m_draw_call_filters_container->show();
+            HideDrawCallFilters();
+
+            QString command_string = m_command_box->currentText();
+            if (command_string.contains("Indirect", Qt::CaseInsensitive))
+            {
+                m_draw_count_filter.label->show();
```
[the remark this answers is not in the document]
GrantComm: Made SetDrawCallFilterSpinnerVisibility that takes in the filter and a bool for visibility

PR #796, opened 2026-03-19
Ensures spin box focus is maintained when values are changed

GrantComm:
Initially when entering values with more than one digit, focus was lost due to calls to UpdateVisibility(). It temporarily hides the specific configuration widgets using HideSpecificSettings() and HideDrawCallFilters() before selectively displaying the relevant ones again. Since changing the values inside the spin boxes or the filter combo boxes only changes whether the "Add Modification" button should be enabled, we should point those specific signals directly to UpdateAddModificationButtonState().


hysw:
Please add reviewer from runtime-if feature, I don't have sufficient context for this.

PR #813, opened 2026-03-27
Prototype for automated UI tests

footballhead:
UI tests can be written as gtests. This is made possible by Qt reflection and the many functions it exposes to query the object graph.

Tips for writing tests:

- Call setObjectName on things you want to find the test. These can be found with findChild
- Send key events using QTest::keyClick
- Manually process the event loop with QCoreApplication::sendPostedEvents
- Use QTest::qWaitFor when waiting for something to happen so that the main event loop is still processed.
- If something isn't working, add more asserts on state. E.g. Qt won't warn if you try to click a button that is disabled; either assert that the button is enabled for qWaitFor it to become enabled.
- I can't seem to get QTest::mouseClick to work so prefer keyboard navigaion, keyboard shortcuts, or directly emitting signals.

The tests are compiled into a plugin which is loaded into the Dive UI process. The plugin runs the tests after the UI has initialized

Test output is printed to stdout/stderr. In the future, I think it makes sense to write this to a file as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,11 +149,27 @@ if(NOT ANDROID)
     # Include gtest before Abseil so we can use ABSL_BUILD_TEST_HELPERS for status_matchers
     # Ask gtest to use /MD or /MDd since it matches the CMake defaults that we use.
     set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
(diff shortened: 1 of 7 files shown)


footballhead:
no plans on proceeding with this at the time being

PR #816, opened 2026-03-31
Prototype of performing post-processing of capture file

footballhead:
Before replay, Dive has the opportunity to modify the capture file to fix issues or perform optimizations. This proof of concept adds a vkQueueWaitIdle call to the end of the capture file before replay. In the future, much of what DiveVulkanReplayConsumer is doing at replay time could be moved to a single post-process step.


footballhead:
no plans on proceeding with this at the time being

PR #832, opened 2026-04-09
Enable GFXR capture between queue submits

footballhead:
Not every application has clear frame boundaries. For example,
benchmarks may not present and may not emit the frame end debug label.
While we don't want to require applications to provide clear frame
boundaries, it's difficult to know what to do when they are not present.

This change modifies the CLI and UI to toggle the GFXR frame end point setting added in #830.

The user can choose between the "frame-to-frame" and "queue-submit-to-queue-submit" behavior:

- In the CLI, provide --gfxr_capture_type queue_submit
- In the GUI, choose the Queue Submit Capture End Point in the Capture
  dialog.

image


capture_service/dive_client_cli.cc
```diff
@@ -978,6 +975,34 @@ std::string AbslUnparseFlag(GfxrReplayOptions run_type)
...
+}
+
+std::string AbslUnparseFlag(GfxrCaptureSettings::EndPoint end_point)
+{
+    switch (end_point)
+    {
+        case GfxrCaptureSettings::EndPoint::kFrame:
+            return "frame";
+        case GfxrCaptureSettings::EndPoint::kQueueSubmit:
+            return "kQueueSubmit";
```
wangra-google: should this one be `queue_submit`?

capture_service/dive_client_cli.cc
```diff
@@ -978,6 +975,34 @@ std::string AbslUnparseFlag(GfxrReplayOptions run_type)
...
+    if (text == "queue_submit")
+    {
+        *end_point = GfxrCaptureSettings::EndPoint::kQueueSubmit;
+        return true;
+    }
+    *error = "unknown value for enumeration";
+    return false;
+}
+
+std::string AbslUnparseFlag(GfxrCaptureSettings::EndPoint end_point)
```
footballhead, line 999: `error: control reaches end of non-void function [-Werror=return-type]`

Despite being an exhaustive switch, anyone can cast an invalid value into an endpoint for any number of reasons. So this still requires a return at the end

capture_service/dive_client_cli.cc
wangra-google: maybe better with some constexpr since this is used on 2 places, just in case we need to change the string for whatever reason in the future.

GrantComm (dismissed):

wangra-google (dismissed):

wangra-google (approved):

GrantComm (approved):

PR #835, opened 2026-04-10
Reorders resource directories

GrantComm:
Reordered the search_dirs arrays in ResolvePluginsDir, ResolveDeviceResourcesLocalPath, and ResolveProfilingResourcesLocalPath to prioritize the Apple bundle specific paths. Currently the generic "Most platforms" fallback path is listed first. If using the local build directory structure, the app would match that first and completely ignore the .app bundle's Resources directory, causing failures. Prioritizing the Apple bundle paths ensures the app will correctly look inside its own packaged Resources directory first. If it doesn't exist (running on Linux/Windows), it falls back to the standard generic paths.


wangra-google (approved):

angela28chen:
Thanks for the PR and the detailed PR description is appreciated!

angela28chen (approved):

PR #838, opened 2026-04-10
Updates assert for subsampled captures

GrantComm:
Supports PM4 capture of applications that use subsampled images with and without course reconstruction.

When MSAA or subsampling is active, the driver batches multiple rectangles into a single draw packet to improve performance. This exceeded the expected index count and triggered an assertion.

Manual Testing:
Linux

Load subsampled demeo rd capture, no crash occurs.


footballhead (approved):
I'd reword the description to be something like: "support PM4 capture of applications that use subsampled images with and without course reconstruction". `--subsampled` is only relevant to an internal benchmark.

wangra-google (approved):

PR #840, opened 2026-04-13
More updates for build script focused on plugins and clean build

angela28chen:
* Add copy_plugins action (for external plugins) to the build script
* `--clean-build` now applies at the beginning of the script and will clear the entire build folder

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -165,15 +165,13 @@ def check_environment(args):
 
 
 def clean_build(args):
-    """Delete args.root_build_dir
-    """
-    if os.path.exists(f"{args.root_build_dir}"):
+    if os.path.exists(args.root_build_dir):
         print("\nClearing build folder...")
-        shutil.rmtree(f"{args.root_build_dir}")
+        shutil.rmtree(args.root_build_dir)
 
 
 def copy_plugins(args):
-    """Implementing ActionType.COPY_PLUGINS stage
+    """Implements ActionType.COPY_PLUGINS stage
     """
     if not os.path.exists(EXTERNAL_PLUGINS_DIR):
         print(f"\nDir {EXTERNAL_PLUGINS_DIR} does not exist, skipping")
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -31,11 +31,11 @@ NINJA_MULTI_CONFIG_NAME = "Ninja Multi-Config"
 
 # TODO: b/484082504 - Add more stages for packaging and deploying (incorporate scripts/deploy_mac_bundle.py)
 class ActionType(enum.StrEnum):
+    COPY_PLUGINS = "copy_plugins"       # copy external plugins into pkg dir
     CONFIGURE_HOST = "configure_host"
     BUILD_HOST = "build_host"           # separated to make VS builds using the UI easy
     INSTALL_HOST = "install_host"
     ALL_DEVICE = "all_device"           # configure, build, and install device libraries
-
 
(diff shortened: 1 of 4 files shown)


scripts/build.py
```diff
@@ -160,6 +160,18 @@ def check_environment(args):
         dive.echo_and_run(cmd)
 
 
+def copy_plugins(args):
+    """Implementing ActionType.COPY_PLUGINS stage
+    """
+    if args.clean_build:
+        if os.path.exists(f"{args.root_build_dir}/pkg/plugins"):
```
angela28chen: There's a minor issue here where it will remove ALL plugins, so when the `clean_build` flag is being used with step `copy_plugins` then we need to rerun `install_host` after to recopy `plugin_sample`. I'm not sure what a good way to address this is, I don't think it's super important but it could be confusing to encounter.

I think one possibility is to make `build/pkg/plugins/external` and have that be cleaned, but then the plugin loading logic needs a small tweak. IMO a good approach since the file structure will become more similar between `root/plugins/` and `build/pkg/plugins`. 

Suggestions welcome!

footballhead: What's the intent build `--clean-build` being orthogonal to the build actions? Why not make it an action and run it first and remove the entire build directory? I doubt one would want a clean build often; if that's the case then they can just build the entire thing again. It's simple to reason about and maintain

angela28chen: I set it up thinking that people may want to have a clean build for one portion (device vs host) while not touching the other portion, but I'm fine with changing it like you suggested if that's not a common workflow.

angela28chen: OK, as discussed during the meeting, I'm going to change --clean-build into another step/action that's before the rest

angela28chen:
Ready for review

scripts/build.py
```diff
@@ -162,14 +164,29 @@ def check_environment(args):
         dive.echo_and_run(cmd)
 
 
+def clean_build(args):
+    """Delete args.root_build_dir
+    """
+    if os.path.exists(f"{args.root_build_dir}"):
+        print("\nClearing build folder...")
+        shutil.rmtree(f"{args.root_build_dir}")
```
footballhead: args.root_build_dir is a string so the format string interpolation is redundant

```suggestion
    if os.path.exists(args.root_build_dir):
        print("\nClearing build folder...")
        shutil.rmtree(args.root_build_dir)
```

scripts/build.py
```diff
@@ -162,14 +164,29 @@ def check_environment(args):
         dive.echo_and_run(cmd)
 
 
+def clean_build(args):
+    """Delete args.root_build_dir
+    """
```
footballhead: Nit: A couple of thoughts on comments based on [Google Python style](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods):

- Be consistent with choice of descriptive ("deletes") or imperative ("delete") within the file. I don't know which one "implementing" falls into
- For single-line docstrings, can put all the closing """ on the same line. Saves space
- Since this function is relatively small, the comment many not be necessary

```suggestion
```

hysw (dismissed):

footballhead (dismissed):

scripts/build.py
angela28chen: OK thanks, done

scripts/build.py
angela28chen: Done

hysw (approved):

footballhead (approved):

PR #841, opened 2026-04-13
Make windows presubmit build Dive host tools faster

angela28chen:
Changes:
- Windows presubmit uses `msbuild` to build host tools rather than initiating the build with `cmake`
- Added github action `microsoft/setup-msbuild` v3, so that `msbuild` is on the path of the runner
- Previous presubmit bottleneck was the Windows builds (~20min), now it runs faster thanks to msbuild being able to use multithreading (~16min) 
  - This was not possible when build is initiated via cmake because there is difficulty in passing down flags to the VS generator

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -23,13 +23,12 @@ jobs:
       - name: Install Dive dependencies
         run: |
           python3 -m pip install mako
-      # TODO: b/484082504 - Remove --build-with-cmake, need to ensure msbuild is on the Path on this runner first
+      - uses: microsoft/setup-msbuild@30375c66a4eea26614e0d39710365f22f8b0af57 # v3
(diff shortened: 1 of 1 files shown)


elviscapiaq (approved):
LGTM. Just by curiosity, what is `30375c66a4eea26614e0d39710365f22f8b0af57` suffix?

footballhead (approved):

angela28chen:
> LGTM. Just by curiosity, what is `30375c66a4eea26614e0d39710365f22f8b0af57` suffix?

This is which commit we're using for the `microsoft/setup-msbuild` github action. It corresponds to the current latest version, v3. The reason that I'm using a commit hash instead of a shorter tag is because this is more secure. In the event that the action is compromised, the tag can be changed which would affect our presubmits.

PR #843, opened 2026-04-13
Exports main window and trace dialog

GrantComm:
Declares export macro for MainWindow and TraceDialog so that they can be accessed from across the DLL boundary for plugin use.

### DIFF
--- a/ui/main_window.h
+++ b/ui/main_window.h
@@ -26,9 +26,10 @@
 #include "dive/ui/forward.h"
 #include "dive_core/cross_ref.h"
 #include "dive_core/log.h"
+#include "dive_ui_lib_export.h"
 #include "ui/progress_tracker_callback.h"
 
-class MainWindow : public QMainWindow
+class DIVE_UI_LIB_EXPORT MainWindow : public QMainWindow
 {
     Q_OBJECT
 
--- a/ui/trace_window.h
+++ b/ui/trace_window.h
@@ -23,6 +23,7 @@
 
 #include "capture_service/device_mgr.h"
 #include "device_dialog.h"
+#include "dive_ui_lib_export.h"
 #include "package_filter.h"
 
 #pragma once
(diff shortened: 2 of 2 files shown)


hysw:
I'd put separate interface in `src/dive/plugin/abi`, export unstable abi is not really a good idea.

wangra-google (approved):

wangra-google:
> I'd put separate interface in `src/dive/plugin/abi`, export unstable abi is not really a good idea.

Thanks Peter for bringing this up. Just to add why this is important:
- With current implementation, the compiler exports the exact memory layout MainWindow. (member variable, layout of the vtable). If anyone modifies the MainWindow class, the memory layout would change, which requires recompilation of all plugins. (by not doing so, you will have weird crashes)
- MainWindow includes Qt headers and other headers, which means the plugin is forced to parse those dependencies which makes the plugins tightly coupled to the internal impl details. (also slowing down compiling time)

i think a standard solution (@hysw please correct me if i am wrong) is to use an abstract base class inside src/dive/plugin/abi (as Peter suggested), and a pure interface has no state and no mem var, its memory layout never changes. (The interface header only needs forward declarations or standard library types. It doesn't need to include any Qt headers or heavy internal headers.) MainWindow & Plugins implements this interface, and the plugin only sees the interface pointer.

hysw:
> i think a standard solution (@hysw please correct me if i am wrong) is to use an abstract base class inside src/dive/plugin/abi (as Peter suggested), and a pure interface has no state and no mem var, its memory layout never changes. (The interface header only needs forward declarations or standard library types. It doesn't need to include any Qt headers or heavy internal headers.) MainWindow & Plugins implements this interface, and the plugin only sees the interface pointer.

The standard qt solution is not abstract base class, but pImpl pointer + signal/slot.

vtable is still exported by the compiler, so you can't change that (i.e. adding virtual function).

See: https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B

hysw (approved):
Approve for now.

Given the plugins are compiled with the main binary anyway, however we need to reconsider if we ever think about having a standalone plugin.

PR #845, opened 2026-04-14
Fix errors while running build.py with configure_host in regular cmd

wangra-google:
This is to fix errors while running `python scripts/build.py --actions configure_host`. (I am having this error even in `Developer Command Prompt`).

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -20,6 +20,7 @@ import enum
 import os
 import platform
 import shutil
+import subprocess
 
 
 # GLOBAL CONSTANTS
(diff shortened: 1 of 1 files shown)


angela28chen:
@wangra-google I'm using VS2022. Are you using a different version? I believe if VS2022 was properly installed then `msbuild` should be on your path inside the Developer Command Prompt (https://learn.microsoft.com/en-us/visualstudio/msbuild/whats-new-msbuild-17-0?view=visualstudio). I think it's also put on the path if you installed VS build tools (https://visualstudio.microsoft.com/visual-cpp-build-tools/)

```bat
REM Inside developer command prompt
C:\src\git\dive>where msbuild
C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin\amd64\MSBuild.exe
C:\Windows\Microsoft.NET\Framework64\v4.0.30319\MSBuild.exe
```

Is that not the case for your setup?

If "msbuild" is not on the path and you are not supplying the exact location with "--msbuild-exec" then you are supposed to be hitting a scary-looking python error at the `msbuild --version` line inside `check_environment()`. I think adding logic to locate the `MsBuild.exe` is kind of outside the scope of this build script.

My intent is that `BUILD.md` guides the user on how to set up the appropriate build environment, and then `build.py` does a cursory check to ensure the environment is set up as expected. IMO you can instead add more details to `BUILD.md` VS prerequisite section about checking if `msbuild` is on the path

wangra-google:
> ```batchfile
> where msbuild
> ```

i am using vs2022 (Dive is in `D` drive, vs2022 is installed on `C` drive)
D:\GitHub\dive>where msbuild
C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin\amd64\MSBuild.exe
C:\Windows\Microsoft.NET\Framework\v4.0.30319\MSBuild.exe

PR #849, opened 2026-04-15
Updates location of dive plugins for macOS

GrantComm:
Updates location of dive plugins for macOS to  **dive.app/Contents/PlugIns/dive_plugins**. macOS requires .dylib files to be either under dive.app/Contents/PlugIns or dive.app/Contents/Resources for notarization.


hysw (dismissed):

hysw (dismissed):

wangra-google (dismissed):

scripts/deploy_mac_bundle.py
```diff
@@ -76,13 +73,25 @@ def main(args):
...
     if args.device_libraries:
         shutil.copytree("device", "dive.app/Contents/Resources", dirs_exist_ok=True)
     else:
         print("\nSkipping copying device libraries because --no-device-libraries...")
 
+    # Find all plugins and add them via the -executable flag so that internal paths pointing to Homebrew are replaced with the bundle-relative paths.
+    cmd = [macdeployqt_exec, "dive.app"]
+    for root, dirs, files in os.walk(plugin_directory):
+        for file in files:
+            if file.endswith(".dylib") or file.endswith(".so"):
```
hysw, line 89: Does macOS uses .so file?

hysw (approved):

scripts/deploy_mac_bundle.py
GrantComm: I think technically it can, provided that the .so file is in macOS format. If someone were to compile a .so on linux and try to run it on mac it wouldn't work.

hysw: We might not want to do recursive walk due to existence of profiling plugin, which has android `.so` file

hysw: Actually, we probably need to split this part into host plugins and device plugins and put them into correct location. Probably for the next PR.

wangra-google (approved):

PR #850, opened 2026-04-15
Disable clang-tidy presubmit check for post-submit.

hysw:


wangra-google (approved):

footballhead (changes requested):
I would suggest fixing errors rather than ignoring them

wangra-google:
> I would suggest fixing errors rather than ignoring them

just out of curiosity, why do we need clang-tidy in addition to presubmit?

.github/workflows/build_mac.yml
footballhead, line 37: This step is only required for clang-tidy. If we think that this should only be run in presubmit then I'd consider merging this step with the "run clang-tidy" step so that the related steps are logically located together and will both be disabled by the `if`. This would save ~45s

footballhead (changes requested):
I've thought about it some more and I'm OK with this change from a technical level. `github.event.pull_request.base.sha ` is only available for merge requests. On post-submit, it's undefined. So the entire repo is scanned, which takes 8minutes.

Although maybe we should have a weekly job that analyzses the entire repo

footballhead:
I fixed the failure in https://github.com/google/dive/pull/851

footballhead:
> > I would suggest fixing errors rather than ignoring them
> 
> just out of curiosity, why do we need clang-tidy in addition to presubmit?

Since we serialize all commits, there's not a whole lot of benefit other than piece of mind: if main is green then I have confidence that it's free of issues.

It would make sense in a more complicated build environment where we didn't have strict requirements for a PR to build against main in order to be submitted, or if we allowed two commits to be merged at once. Then, there would be a discrepancy between what is being run in presubmit compared to postsubmit; this would warrant the additional check

PR #852, opened 2026-04-16
Change the kDeviceDownloadPath to fix some issues on certain devices

wangra-google:
- change "/sdcard/Download" to "/data/local/tmp/dive" as libwrap.so doesn't have write permission to folder "/sdcard/Download" on some devices/OS, also Download folder is deleted after reboot


RenfengLiu (approved):

src/dive/utils/device_resources_constants.h
```diff
@@ -89,7 +89,7 @@ Profiling resources dir:    DIVE_ROOT/build/pkg/CMAKE_GENERATED_PROFILING_PLUGIN
     // -------------------------------------------------------------------------
 
     // The absolute path to the standard Download directory on Android.
-    static constexpr char kDeviceDownloadPath[] = "/sdcard/Download";
+    static constexpr char kDeviceDownloadPath[] = "/data/local/tmp/dive";
```
footballhead, line 96: This is also used by GFXR capture, and apps don't have permission to write to /data/local/tmp. Do you know what's deleting /sdcard/Download? Not all devices do this. Does /sdcard work?

footballhead: So /data/local/tmp doesn't work but subdirectories do. That's where my confusion was coming from.

footballhead, line 96: I ran into this problem with GFXR capture

> Fail to setup for package com.google.bigwheels.project_sample_04_cube.debug error: Command `adb -s R32X80012TL shell setprop debug.gfxrecon.capture_file /data/local/tmp/dive/dive_capture_files/com.google.bigwheels.project_sample_04_cube.debug.gfxr` failed with return code 256, error: Value '/data/local/tmp/dive/dive_capture_files/com.google.bigwheels.project_sample_04_cube.debug.gfxr' is too long, 94 bytes vs a max of 92

footballhead: We'd have to look into truncation of the file name. I'd say maybe a shorter file but the user can set this to be howoever long they want. Unless we take away that configuration.

footballhead (approved):
Truncation is an issue outside of this PR, this just exacerbates it.

src/dive/utils/device_resources_constants.h
wangra-google: should be fixed by https://github.com/google/dive/pull/856

PR #853, opened 2026-04-16
Add AndroidGlesEarlyInit to fix infinite loop on some devices/os

wangra-google:
this is causing some issues on some devices/os.


layer/layer_common.cc
```diff
@@ -71,16 +71,7 @@ struct InitServer
         if (is_libwrap_loaded)
         {
             server_thread = std::thread([]() {
-                // This is to make sure libraries are loaded before running gRPC server. This is a
```
elviscapiaq, line 74: It seems the issue on GLES  apps is that the layer library is loaded so early that there is no chance to mark  `g_layer_load_status` as `true`. Therefore, the loop is infinite.
While I think removing this part is safe for most of the cases, I am concern that for some vulkan or openxr apps, we will not be able to load the library properly. I think a better solution is to mark the library as loaded for GLES specifically without removing this code or we can just sleep for 2 seconds. 
First solution should looks like at the GLES layer (gles_layer.cc):
```
__attribute__((constructor)) void AndroidGlesEarlyInit() {
    SetLayerStatusLoaded();
}
```

layer/layer_common.cc
```diff
@@ -71,16 +71,7 @@ struct InitServer
...
-                // workaround to make sure that gRPC dependencies are loaded before we start the
-                // service. we can remove this check once we transit to our own socket
-                // implementation.
-                while (!IsLayerLoaded())
```
footballhead, line 78: This is the only user of IsLayerLoaded, so the supporting code can also be removed

wangra-google: i went with the suggestion from Elvis, so we are keeping this IsLayerLoaded

layer/layer_common.cc
wangra-google: good point, updated the code

elviscapiaq (approved):

RenfengLiu (approved):

PR #854, opened 2026-04-16
Add Windows release process to the build script, also some refactoring

angela28chen:
Changes:
* Added new stages at the end of the build script, including linux/mac placeholders:
  * DEPLOY_QT : Runs `*deployqt` to put QT dependencies into the appropriate folder
  * PACKAGE : Removes "unnecessary" files in the deployment folder `build/pkg`, before making it into an archive in `build/`

Some minor changes:
* Renamed some of the flags `*_exec` to `exec_*` just so that it's more organized alphabetically
* Boolean flags now support `--no-` prefix
* Fixed some mistyped help text
* Removed `--clean-build` since we can just delete the build folder before running the script

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -18,7 +18,6 @@ import argparse
 import common_dive_utils as dive
 import enum
 import os
-import pathlib
 import platform
 import shutil
 
@@ -57,6 +56,12 @@ def get_default_deployqt():
         return "windeployqt"
     # TODO: b/484082504 - Support other platforms
     return None
+
+
+def rmtree_if_exists(path):
+    if os.path.exists(path):
+        shutil.rmtree(path)
+        print(f"\nDeleted existing dir: {path}")
 
 
 def parse_args():
@@ -123,6 +128,8 @@ def parse_args():
         "--root-build-dir",
         default="build",
         help="The build directory, relative to Dive root directory")
+    # TODO: b/504654590 - Make a platform-specific default so that users can
+    # have more flexibility with choosing a generator
     parser.add_argument(
         "--visual-studio-name",
         default="Visual Studio 17 2022",
@@ -202,7 +209,7 @@ def copy_plugins(args):
 
 
 def configure_host(args):
-    """Implementing ActionType.CONFIGURE_HOST stage, uses cmake to generate
+    """Implements ActionType.CONFIGURE_HOST stage, uses cmake to generate
     build files targeting the host
     """
     print("\nGenerating build files with cmake...")
@@ -261,9 +268,7 @@ def install_host(args):
     # TODO: b/482108095 - This must be done regardless of clean build because
     # install folders are re-used between different build types and the
     # timestamps could affect whether files within are correctly replaced or not
-    if os.path.exists(f"{args.root_build_dir}/{PKG_DIR}/host"):
-        print("\nClearing host pkg folders...")
-        shutil.rmtree(f"{args.root_build_dir}/{PKG_DIR}/host")
+    rmtree_if_exists(f"{args.root_build_dir}/{PKG_DIR}/host")
 
     print("\nInstalling with cmake...")
     cmd = [args.exec_cmake,
@@ -304,9 +309,7 @@ def all_device(args):
     # TODO: b/482108095 - This must be done regardless of clean build because
     # install folders are re-used between different build types and the
     # timestamps could affect whether files within are correctly replaced or not
-    if os.path.exists(f"{args.root_build_dir}/{PKG_DIR}/device"):
-        print("\nClearing install folder...")
-        shutil.rmtree(f"{args.root_build_dir}/{PKG_DIR}/device")
+    rmtree_if_exists(f"{args.root_build_dir}/{PKG_DIR}/device")
 
     print("\nInstalling with cmake...")
     cmd = [args.exec_cmake,
@@ -361,17 +364,15 @@ def package(args):
 
         case "Windows":
             for dir_name in WINDOWS_UNNECESSARY_DIRS:
-                if os.path.exists(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}"):
-                    print(f"\nClearing unnecessary {dir_name} folder...")
-                    shutil.rmtree(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}")
+                rmtree_if_exists(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}")
         
             print(f"\nZipping into '{APP_DIR}.zip'...")
             shutil.make_archive(f"{APP_DIR}", "zip", f"{args.root_build_dir}/{PKG_DIR}")
 
-            final_location_zip = pathlib.Path(f"{args.root_build_dir}/{APP_DIR}.zip")
-            if final_location_zip.exists():
+            final_location_zip = f"{args.root_build_dir}/{APP_DIR}.zip"
+            if os.path.exists(final_location_zip):
                 print("\nClearing previous archive...")
-                final_location_zip.unlink()
+                os.unlink(final_location_zip)
 
             print(f"\nMoving zip archive to {os.getcwd()}/{final_location_zip}...")
             shutil.move(f"{APP_DIR}.zip", final_location_zip)
@@ -388,6 +389,8 @@ def main():
         list_actions()
         return
 
+    # TODO: b/504654590 - Use parser.add_argument() functionality to avoid
+    # having to call parse_actions() separately
     parse_actions(args)
 
     if not args.ci:
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -43,7 +43,7 @@ class ActionType(enum.StrEnum):
     INSTALL_HOST = "install_host"
     ALL_DEVICE = "all_device"           # configure, build, and install device libraries
     DEPLOY_QT = "deploy_qt"
-    CLEAN_AND_ZIP = "clean_and_zip"     # removes some extraneous files and compresses the release
+    PACKAGE = "package"                 # removes some extraneous files and packages the release
 
 
 class BuildType(enum.StrEnum):
@@ -147,78 +147,56 @@ def parse_actions(args):
                                 "to --list-actions for the valid list OR leave blank to run all")
             actions.append(action)
     else:
-        actions = []
-        for action in ActionType:
-            actions.append(action)
-            if (args.build_type == BuildType.DEBUG) and (action == ActionType.ALL_DEVICE) and (platform.system() != "Darwin"):
-                # For Windows and Linux Debug builds, by default the last action is ALL_DEVICE
-                break
-            if (args.build_type == BuildType.DEBUG) and (action == ActionType.DEPLOY_QT) and (platform.system() == "Darwin"):
-                # For macOS Debug builds, by default the last action is DEPLOY_QT
-                break
+        actions = [x for x in ActionType]
 
     str_actions = ", ".join(actions)
     print(f"\nQueued actions, not necessarily listed in order: {str_actions}")
 
+    if (not args.prereq_checks) and (ActionType.package in actions) and (not args.exec_deployqt):
+        raise Exception("Must pass value for --exec-deployqt when bypassing prereq checks for action DEPLOY_QT")
+
     args.actions = actions
-
-
-def check_intersect(lis1, lis2):
-    intersect = set(lis1) & set(lis2)
-    return len(intersect) > 0
 
 
 def check_environment(args):
     """Checks for env vars and executables that will be used by this script.
-    Overwrites some args.exec_* values in-place with valid ones.
+    Overwrites args.exec_deployqt value in-place with platform-specific valid name.
     Can be disabled with --no-prereq-checks
     """
-    if check_intersect([ActionType.ALL_DEVICE], args.actions):
-        if args.build_type != BuildType.DEBUG:
-            print("WARNING: GFXR gradle build is hardcoded to Debug regardless of --build_type")
 
-        print("\nChecking ANDROID_NDK_HOME...")
-        if "ANDROID_NDK_HOME" not in os.environ:
-            raise Exception("ANDROID_NDK_HOME env var must be set")
+    if args.build_type != BuildType.DEBUG:
+        print("WARNING: GFXR gradle build is hardcoded to Debug regardless of --build_type")
 
-    cmake_actions = [ActionType.CONFIGURE_HOST, ActionType.BUILD_HOST, ActionType.INSTALL_HOST, ActionType.ALL_DEVICE]
-    if check_intersect(cmake_actions, args.actions):
-        cmd = [args.exec_cmake, "--version"]
-        dive.echo_and_run(cmd)
+    if "ANDROID_NDK_HOME" not in os.environ:
+        raise Exception("ANDROID_NDK_HOME env var must be set")
 
-    check_ninja = False
-    if ActionType.ALL_DEVICE in args.actions:
-        check_ninja = True
-    elif platform.system() in ["Linux", "Darwin"]:
-        if (ActionType.BUILD_HOST in args.actions) and args.build_via_generator:
-            check_ninja = True
-    if check_ninja:
-        cmd = [args.exec_ninja, "--version"]
-        dive.echo_and_run(cmd)
+    cmd = [args.exec_cmake, "--version"]
+    dive.echo_and_run(cmd)
 
-    if (platform.system() == "Windows") and (args.build_via_generator) and (ActionType.BUILD_HOST in args.actions):
+    cmd = [args.exec_ninja, "--version"]
+    dive.echo_and_run(cmd)
+
+    if platform.system() == "Windows":
         cmd = [args.exec_msbuild, "--version"]
         dive.echo_and_run(cmd)
 
-    if ActionType.DEPLOY_QT in args.actions:
-        deploy_qt_exec = args.exec_deployqt
-        if not deploy_qt_exec:
-            if (platform.system() == "Darwin"):
-                deploy_qt_exec = "macdeployqt"
+    deploy_qt_exec = args.exec_deployqt
+    if not deploy_qt_exec:
+        if platform.system() == "Windows":
+            deploy_qt_exec = "windeployqt"
+        else:
+            # TODO: b/484082504 - Support other platforms
+            print("\nDefault *deployqt is not implemented")
 
-            elif (platform.system() == "Windows"):
-                deploy_qt_exec = "windeployqt"
+    if deploy_qt_exec:
+        # Cannot check the --version because for some reason the return code is nonzero
+        deployqt_found = shutil.which(deploy_qt_exec)
+        if not deployqt_found:
+            raise Exception(f"{deploy_qt_exec} not found on the Path")
+        args.exec_deployqt = deploy_qt_exec
 
-        if deploy_qt_exec:
-            # Cannot check the --version because for some reason the return code is nonzero
-            deployqt_found = shutil.which(deploy_qt_exec)
-            if not deployqt_found:
-                raise Exception(f"{deploy_qt_exec} not found on the Path")
-            args.exec_deployqt = deploy_qt_exec
-
-    if (platform.system() == "Windows") and (ActionType.CLEAN_AND_ZIP in args.actions):
-        cmd = [args.exec_tar, "--version"]
-        dive.echo_and_run(cmd)
+    cmd = [args.exec_tar, "--version"]
+    dive.echo_and_run(cmd)
 
 
 def copy_plugins(args):
@@ -348,18 +326,21 @@ def all_device(args):
 
 
 def deploy_qt(args):
-    """Implements ActionType.DEPLOY_QT stage
+    """Implements ActionType.DEPLOY_QT stage, which uses a QT tool to add QT
+    libraries to the same dir as the binary that depends on them in preparation
+    for packaging and distribution
     """
     # TODO: b/484082504 - Support other platforms
-    if (platform.system() == "Linux"):
+    # TODO: b/504654590 - Investigate if there's a way to identify the binaries without hardcoding them
+    if platform.system() == "Linux":
         print("\ndeploy_qt() is UNIMPLEMENTED for Linux...")
         return
 
-    elif (platform.system() == "Darwin"):
+    elif platform.system() == "Darwin":
         print("\ndeploy_qt() is UNIMPLEMENTED for macOS...")
         return
 
-    elif (platform.system() == "Windows"):
+    elif platform.system() == "Windows":
         print(f"\nDeploying with {args.exec_deployqt}...")
         cmd = [args.exec_deployqt, 
                f"{args.root_build_dir}/{PKG_DIR}/host/{WINDOWS_QT_DEPENDENCIES_BINARY}"
@@ -370,19 +351,21 @@ def deploy_qt(args):
         raise Exception(f"Unrecognized platform: {platform.system()}")
 
 
-def clean_and_zip(args):
-    """Implements ActionType.CLEAN_AND_ZIP stage
+def package(args):
+    """Implements ActionType.PACKAGE stage, which will package this Dive build
+    into a single file that is easy to transfer
     """
     # TODO: b/484082504 - Support other platforms
-    if (platform.system() == "Linux"):
-        print("\nclean_and_zip() is UNIMPLEMENTED for Linux...")
+    # TODO: b/504654590 - Investigate if there's a way to identify extraneous files
+    if platform.system() == "Linux":
+        print("\npackage() is UNIMPLEMENTED for Linux...")
         return
 
-    elif (platform.system() == "Darwin"):
-        print("\nclean_and_zip() is UNIMPLEMENTED for macOS...")
+    elif platform.system() == "Darwin":
+        print("\npackage() is UNIMPLEMENTED for macOS...")
         return
 
-    elif (platform.system() == "Windows"):
+    elif platform.system() == "Windows":
         for dir_name in WINDOWS_UNNECESSARY_DIRS:
             if os.path.exists(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}"):
                 print(f"\nClearing unnecessary {dir_name} folder...")
@@ -443,9 +426,9 @@ def main():
         if ActionType.DEPLOY_QT in args.actions:
             with dive.Timer(ActionType.DEPLOY_QT):
                 deploy_qt(args)
-        if ActionType.CLEAN_AND_ZIP in args.actions:
-            with dive.Timer(ActionType.CLEAN_AND_ZIP):
-                clean_and_zip(args)
+        if ActionType.PACKAGE in args.actions:
+            with dive.Timer(ActionType.PACKAGE):
+                package(args)
 
 
 if __name__ == "__main__":
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -36,13 +36,13 @@ WINDOWS_QT_DEPENDENCIES_BINARY = "dive_gui_lib.dll"
 
 
 class ActionType(enum.StrEnum):
-    COPY_PLUGINS = "copy_plugins"       # copy external plugins into pkg dir
-    CONFIGURE_HOST = "configure_host"
-    BUILD_HOST = "build_host"           # separated to make VS builds using the UI easy
-    INSTALL_HOST = "install_host"
-    ALL_DEVICE = "all_device"           # configure, build, and install device libraries
-    DEPLOY_QT = "deploy_qt"
-    PACKAGE = "package"                 # removes some extraneous files and packages the release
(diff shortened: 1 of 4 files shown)


angela28chen:
Manual Testing:
* Windows: I'm able to unzip `dive_app.zip` and confirm that dive UI launches without issue, "About Dive" shows correct info including external plugin

angela28chen:
Maybe in a following PR I can try hooking up these extra two steps for the nightly windows release build so that it can produce zip files? I'm not sure if the test artifacts are downloadable from github but I think it ought to be. I can call them "nightly" builds or something

scripts/build.py
```diff
@@ -27,17 +28,23 @@
...
 
-# TODO: b/484082504 - Add more stages for packaging and deploying (incorporate scripts/deploy_mac_bundle.py)
 class ActionType(enum.StrEnum):
     COPY_PLUGINS = "copy_plugins"       # copy external plugins into pkg dir
     CONFIGURE_HOST = "configure_host"
     BUILD_HOST = "build_host"           # separated to make VS builds using the UI easy
     INSTALL_HOST = "install_host"
     ALL_DEVICE = "all_device"           # configure, build, and install device libraries
+    DEPLOY_QT = "deploy_qt"
+    CLEAN_AND_ZIP = "clean_and_zip"     # removes some extraneous files and compresses the release
```
footballhead: I'd avoid using the term `clean` here since it already means something different in `--clean-build`. Maybe "make_releasable_zip" or just "package" and leave the file removal part as an implementation detail

scripts/build.py
```diff
@@ -163,6 +178,10 @@ def check_environment(args):
         cmd = [args.msbuild_exec, "--version"]
         dive.echo_and_run(cmd)
 
+    if (platform.system() == "Windows"):
```
footballhead: nit: parans are unnecessary

```suggestion
    if platform.system() == "Windows":
```

scripts/build.py
```diff
@@ -296,6 +330,105 @@ def all_device(args):
...
+            for file in files:
+                if file.endswith(".dylib") or file.endswith(".so"):
+                    plugin_path = os.path.join(root, file)
+                    cmd.append(f"-executable={plugin_path}")
+        dive.echo_and_run(cmd)
+
+        if args.mac_local_sign:
+            print(f"\nSign bundle using codesign...")
+            cmd = ["codesign", "--force", "--deep", "--sign", "-", bundle_path]
+        dive.echo_and_run(cmd)
```
elviscapiaq: Do we need to put this inside the if? 
otherwise we are going to execute the same command twice if args.mac_local_sign is False.

scripts/build.py
```diff
@@ -296,6 +330,105 @@ def all_device(args):
...
+    
+        print("\nZipping with tar...")
+        former_dir = os.getcwd()
+        os.chdir(f"{args.root_build_dir}/{PKG_DIR}")
+        cmd = ["tar", "-a", "-c", "-v", "-f", f"{APP_DIR}.zip", "--exclude", f"{APP_DIR}.zip", "*"]
+        dive.echo_and_run(cmd)
+        os.chdir(former_dir)
+
+        final_location_zip = pathlib.Path(f"{args.root_build_dir}/{APP_DIR}.zip")
+        if os.path.exists(final_location_zip):
```
elviscapiaq: Maybe we can use here `if final_location_zip.exists():` to avoid mixing os.path and pathlib?

scripts/build.py
```diff
@@ -120,30 +141,39 @@ def list_actions():
...
     return actions
 
 
-def check_environment(args):
+def check_environment(args, actions):
     """Checks for env vars and executables that will be used by this script.
-    Can be disabled with --bypass-prereq-checks
+    Can be disabled with --no-prereq-checks
     """
     if args.build_type != "Debug":
```
elviscapiaq: should we use `if args.build_type != BuildType.DEBUG:` here for better consistency?

scripts/build.py
angela28chen: good catch, thanks!

angela28chen: Ended up removing the mac portion from this PR

scripts/build.py
```diff
@@ -63,44 +70,51 @@ def parse_args():
...
         "--build-via-generator",
-        action="store_true",
+        action=argparse.BooleanOptionalAction,
+        default=False,
         help="Perform the build step by invoking the generator rather than cmake, potentially faster on Windows platform")
     parser.add_argument(
-        "--bypass-prereq-checks",
-        action="store_true",
-        help="Bypass the check for if prerequisites are located as expected")
+        "--prereq-checks",
```
footballhead, line 89: +1 positive flags are easier to reason about

scripts/build.py
```diff
@@ -27,17 +28,23 @@
 EXTERNAL_PLUGINS_DIR = "plugins/external"
 # Build directory structure relative to --root-build-dir
 PKG_DIR = "pkg"
+APP_DIR = "dive_app"
 # CMake generator names
 NINJA_MULTI_CONFIG_NAME = "Ninja Multi-Config"
+# For release
+WINDOWS_UNNECESSARY_DIRS = ["translations"]
+WINDOWS_QT_DEPENDENCIES_BINARY = "dive_gui_lib.dll"
```
footballhead, line 35: Is it worth adding a comment about why this binary in particular, or maybe a TODO bug to look at all releasable binaries

scripts/build.py
```diff
@@ -120,54 +134,91 @@ def list_actions():
...
-    if args.build_type != "Debug":
-        print("WARNING: GFXR gradle build is hardcoded to Debug regardless of --build_type")
-    print("\nChecking env vars...")
-    if "ANDROID_NDK_HOME" not in os.environ:
-        raise Exception("ANDROID_NDK_HOME env var must be set")
-
-    print("\nChecking cmake...")
-    cmd = [args.cmake_exec, "--version"]
-    dive.echo_and_run(cmd)
+    if check_intersect([ActionType.ALL_DEVICE], args.actions):
```
footballhead: I worry about adding all these extra checks because it makes this function hard to reason about and it could change if new steps are added into the future. How I think about this is: if I asked someone else to implement a new ActionType, how many lines of code would it take? Which functions would need to be modified? Would they find all the functions or would they miss one and cause a bug? Ideally, all the code is collocated so that it can be easily found

scripts/build.py
```diff
@@ -288,14 +339,73 @@ def all_device(args):
...
            "--config", f"{args.build_type}"
            ]
     dive.echo_and_run(cmd)
 
 
+def deploy_qt(args):
+    """Implements ActionType.DEPLOY_QT stage
+    """
+    # TODO: b/484082504 - Support other platforms
+    if (platform.system() == "Linux"):
```
footballhead: nit: brackets are unnecessary

```suggestion
    if platform.system() == "Linux":
```

scripts/build.py
```diff
@@ -288,14 +339,73 @@ def all_device(args):
...
+        print("\ndeploy_qt() is UNIMPLEMENTED for Linux...")
+        return
+
+    elif (platform.system() == "Darwin"):
```
footballhead: nit brackets are unnecessary

```suggestion
    elif platform.system() == "Darwin":
```

scripts/build.py
```diff
@@ -288,14 +339,73 @@ def all_device(args):
...
+        print("\ndeploy_qt() is UNIMPLEMENTED for macOS...")
+        return
+
+    elif (platform.system() == "Windows"):
```
footballhead: nit: brackets are unnecessary
```suggestion
    elif platform.system() == "Windows":
```

scripts/build.py
```diff
@@ -288,14 +339,73 @@ def all_device(args):
...
            "--install", f"{args.root_build_dir}/device",
            "--prefix", f"{args.root_build_dir}/{PKG_DIR}",
            "--config", f"{args.build_type}"
            ]
     dive.echo_and_run(cmd)
 
 
+def deploy_qt(args):
+    """Implements ActionType.DEPLOY_QT stage
+    """
```
footballhead: This comment doesn't add any value since I could guess it from the function name. Consider removing it or writing down what behavior the caller can expect

scripts/build.py
```diff
@@ -288,14 +339,73 @@ def all_device(args):
...
+
+        final_location_zip = pathlib.Path(f"{args.root_build_dir}/{APP_DIR}.zip")
+        if final_location_zip.exists():
+            print("\nClearing previous archive...")
+            final_location_zip.unlink()
+        print(f"\nMoving zip archive to {os.getcwd()}/{args.root_build_dir}/{APP_DIR}.zip...")
+        shutil.move(f"{args.root_build_dir}/{PKG_DIR}/{APP_DIR}.zip", final_location_zip)
+
+    else:
+        raise Exception(f"Unrecognized platform: {platform.system()}")
```
footballhead: Same comments from deploy_qt apply here

scripts/build.py
```diff
@@ -288,14 +339,73 @@ def all_device(args):
...
+    elif (platform.system() == "Windows"):
+        for dir_name in WINDOWS_UNNECESSARY_DIRS:
+            if os.path.exists(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}"):
+                print(f"\nClearing unnecessary {dir_name} folder...")
+                shutil.rmtree(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}")
+    
+        print("\nZipping with tar...")
+        former_dir = os.getcwd()
+        os.chdir(f"{args.root_build_dir}/{PKG_DIR}")
+        cmd = [args.exec_tar, "-a", "-c", "-v", "-f", f"{APP_DIR}.zip", "--exclude", f"{APP_DIR}.zip", "*"]
```
footballhead: Is dive_app.zip compressed? I don't know about windows tar but on Linux I just get an uncompressed tarball

scripts/build.py
```diff
@@ -120,54 +134,91 @@ def list_actions():
...
-        actions = [x for x in ActionType]
+        actions = []
+        for action in ActionType:
+            actions.append(action)
+            if (args.build_type == BuildType.DEBUG) and (action == ActionType.ALL_DEVICE) and (platform.system() != "Darwin"):
+                # For Windows and Linux Debug builds, by default the last action is ALL_DEVICE
+                break
+            if (args.build_type == BuildType.DEBUG) and (action == ActionType.DEPLOY_QT) and (platform.system() == "Darwin"):
+                # For macOS Debug builds, by default the last action is DEPLOY_QT
+                break
```
footballhead: Is this only temporary due to b/484082504? If so, I would just let those steps run as noops since I don't think the extra complexity here is warranted

scripts/build.py
angela28chen: Added a TODO inside `deploy_qt()`

scripts/build.py
angela28chen: expanded comment

scripts/build.py
angela28chen: I guess I meant "package" and not "compress", reworded comment elsewhere

scripts/build.py
angela28chen: fair

scripts/build.py
angela28chen: OK, I will just make it check everything unless the user passes `--no-prereq-checks`

angela28chen: I added a check inside parse_actions() for an edge case

scripts/build.py
angela28chen: Originally it wasn't supposed to be temporary, but thinking about it now, it's probably OK to have these stages run by default... I've changed it

scripts/build.py
angela28chen: I think I've addressed the function description and bracket stuff

scripts/build.py
footballhead: If we want a tar file I'd change the file extension to .tar so that someone doesn't try to run `unzip` on it and get confused when it doesn't work.
```suggestion
        cmd = [args.exec_tar, "-a", "-c", "-v", "-f", f"{APP_DIR}.tar", "--exclude", f"{APP_DIR}.tar", "*"]
```
Then if we want compression we could add it to the suffix, e.g. to use gzip:
```suggestion
        cmd = [args.exec_tar, "-a", "-c", "-v", "-f", f"{APP_DIR}.tgz", "--exclude", f"{APP_DIR}.tgz", "*"]
```
Alternatively, python offers standard libraries for dealing with archives that we could consider using here instead, like [tarfile](https://docs.python.org/3/library/tarfile.html) and [zipfile](https://docs.python.org/3/library/zipfile.html). If these come standard then we don't need `--exec-tar`

scripts/build.py
```diff
@@ -288,14 +317,78 @@ def all_device(args):
...
+
+    elif platform.system() == "Windows":
+        print(f"\nDeploying with {args.exec_deployqt}...")
+        cmd = [args.exec_deployqt, 
+               f"{args.root_build_dir}/{PKG_DIR}/host/{WINDOWS_QT_DEPENDENCIES_BINARY}"
+               ]
+        dive.echo_and_run(cmd)
+
+    else:
+        raise Exception(f"Unrecognized platform: {platform.system()}")
```
footballhead: (optional) using `match` here could make this a little more succinct:


```suggestion
    match platform.system():
        case "Linux":
            print("\ndeploy_qt() is UNIMPLEMENTED for Linux...")

        case "Darwin":
            print("\ndeploy_qt() is UNIMPLEMENTED for macOS...")

        case "Windows":
            print(f"\nDeploying with {args.exec_deployqt}...")
            cmd = [
                args.exec_deployqt, 
                f"{args.root_build_dir}/{PKG_DIR}/host/{WINDOWS_QT_DEPENDENCIES_BINARY}"
            ]
            dive.echo_and_run(cmd)

        case unknown:
            raise Exception(f"Unrecognized platform: {unknown}")
```

scripts/build.py
```diff
@@ -63,44 +70,51 @@ def parse_args():
...
-        "configure_host stage.")
+        "--exec-cmake",
+        default="cmake",
+        help="The name of the CMake executable on the path")
     parser.add_argument(
-        "--list-actions",
-        action="store_true",
-        help="Build is part of continuous integration")
+        "--exec-deployqt",
+        help="The path to the platform-specific deployqt executable, if blank will use windeployqt/macdeployqt")
```
footballhead: Could consider adding a platform-specific default here. That way, check_environment() or other users don't need to do as much validation. Additionally, `check_environment()` currently unceremoniously overrides the value provided by the user; since the override could be confusing, preferring the default over the override is easier to understand:

```suggestion
        help="The path to the platform-specific deployqt executable",
        default="windeployqt" if platform.system() == "Windows" else None)
```

This could be extended to other platforms in the future, either by ternary or osmething more readable:

```suggestion
        help="The path to the platform-specific deployqt executable",
        default="windeployqt" if platform.system() == "Windows" else ("macdeployqt" if platform.system() == "Darwin" else (...)))
```

angela28chen: I see, thanks for the suggestion, I've added a function to get the default value just because I prefer reading it over everything in one line

scripts/build.py
angela28chen: Neat, I didn't realize python had a switch statement that also handles strings. Changed but I kept the if/else pattern in build_host()

scripts/build.py
angela28chen: I see. I'm not sure if we have a strong preference for tar format vs zip format... 

I've decided to use `shutil.make_archive()` and make a zip file for windows build. I chose `shutil` since it's already used a lot in this script and it's part of the standard python library. It also has some flexibility so if we choose to do another archive format for linux/mac, it should be easy.

scripts/build.py
```diff
@@ -288,14 +309,77 @@ def all_device(args):
...
+                    print(f"\nClearing unnecessary {dir_name} folder...")
+                    shutil.rmtree(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}")
+        
+            print(f"\nZipping into '{APP_DIR}.zip'...")
+            shutil.make_archive(f"{APP_DIR}", "zip", f"{args.root_build_dir}/{PKG_DIR}")
+
+            final_location_zip = pathlib.Path(f"{args.root_build_dir}/{APP_DIR}.zip")
+            if final_location_zip.exists():
+                print("\nClearing previous archive...")
+                final_location_zip.unlink()
```
footballhead: nit: While I prefer Pathlib, I'd recommend `os`/`os.path` to remain consistent with the rest of the file.

```suggestion
            final_location_zip = f"{args.root_build_dir}/{APP_DIR}.zip"
            if os.path.exists(final_location_zip):
                print("\nClearing previous archive...")
                os.unlink(final_location_zip)
```

scripts/build.py
```diff
@@ -288,14 +309,77 @@ def all_device(args):
...
+
+        case "Darwin":
+            print("\npackage() is UNIMPLEMENTED for macOS...")
+            return
+
+        case "Windows":
+            for dir_name in WINDOWS_UNNECESSARY_DIRS:
+                if os.path.exists(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}"):
+                    print(f"\nClearing unnecessary {dir_name} folder...")
+                    shutil.rmtree(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}")
```
footballhead: nit: i see this "remove X if exists" in several places, would it make sense to combine it into one function? It's more more succinct, reduces indendation, and less bug prone since the file is only spelled once (instead of twice).

```py
def rmtree_if_exists(path: str):
    if os.path.exists(path):
        shutil.rmtree(path)
    # There are several single-line alternatives too
```

It would be used like:

```suggestion    
                rmtree_if_exists(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}")
```

scripts/build.py
```diff
@@ -183,23 +202,25 @@ def copy_plugins(args):
...
+                   f"-DDIVE_RELEASE_TYPE={args.dive_release_type}"
+                   ]
+        case "Windows":
+            cmd = [args.exec_cmake, ".",
+                   f"-G{args.visual_studio_name}",
+                   f"-B{args.root_build_dir}/host",
+                   f"-DDIVE_RELEASE_TYPE={args.dive_release_type}"
+                   ]
+        case _:
+            raise Exception(f"Unrecognized platform: {platform.system()}")
```
footballhead: optional for this PR, but I would recommend replacing --visual-studio-name with --host-cmake-generator using a platform-specific default. I like using ninja on Windows and I currently can't use it; implementing --host-cmake-generator in this way would keep the current behavior and allow the user to specify different generators if they desire. It would also simpify this logic:


```suggestion
    cmd = [args.exec_cmake, ".",
           f"-G{args.host_cmake_generator}",
           f"-B{args.root_build_dir}/host",
           f"-DDIVE_RELEASE_TYPE={args.dive_release_type}"
           ]
```

scripts/build.py
```diff
@@ -183,23 +202,25 @@ def copy_plugins(args):
 
 
 def configure_host(args):
-    """Implementing ActionType.CONFIGURE_HOST stage
+    """Implementing ActionType.CONFIGURE_HOST stage, uses cmake to generate
```
footballhead: nit: remain consistent with other docs

```suggestion
    """Implements ActionType.CONFIGURE_HOST stage, uses cmake to generate
```

scripts/build.py
```diff
@@ -304,33 +388,38 @@ def main():
         list_actions()
         return
 
-    actions = parse_actions(args.actions)
+    parse_actions(args)
```
footballhead, line 411: `add_argument()` has a `type` parameter which can be a class with a single argument constructor or a callable with a single argument. That way, the caller of `parse_args()` can assume that `args.actions` is valid and doesn't need to (forget to) perform additional validation

scripts/build.py
angela28chen: Thank you for the suggestion

scripts/build.py
angela28chen: Noted, I've added it to the internal cleanup bug and won't address it in this PR

scripts/build.py
angela28chen: Added a todo for this

angela28chen:
Actually since we aren't looking at making github actions produce builds to download anytime soon, I'm going to remove the non- build.py changes from this PR

scripts/build.py
```diff
@@ -27,24 +27,43 @@
...
 
-# TODO: b/484082504 - Add more stages for packaging and deploying (incorporate scripts/deploy_mac_bundle.py)
 class ActionType(enum.StrEnum):
     COPY_PLUGINS = "copy_plugins"       # copy external plugins into pkg dir
     CONFIGURE_HOST = "configure_host"
     BUILD_HOST = "build_host"           # separated to make VS builds using the UI easy
     INSTALL_HOST = "install_host"
     ALL_DEVICE = "all_device"           # configure, build, and install device libraries
+    DEPLOY_QT = "deploy_qt"
+    PACKAGE = "package"                 # removes some extraneous files and packages the release
```
footballhead: optional: you can use enum.auto() here and python will generate lowercase values based on the identifier. prevents accidental name/value mismatch since the value is autogenerated from the name instead of manually typed. Also, slightly less work to add new enums

```suggestion
    # ActionType.DEPLOY_QT == "deploy_qt"
    DEPLOY_QT = enum.auto()
    # ActionType.PACKAGE == "package"
    PACKAGE = enum.auto()               # removes some extraneous files and packages the release
```

scripts/build.py
```diff
@@ -136,38 +161,39 @@ def parse_actions(lis):
...
-    cmd = [args.cmake_exec, "--version"]
+    cmd = [args.exec_cmake, "--version"]
     dive.echo_and_run(cmd)
 
-    print("\nChecking ninja...")
-    cmd = [args.ninja_exec, "--version"]
+    cmd = [args.exec_ninja, "--version"]
     dive.echo_and_run(cmd)
 
-    if (platform.system() == "Windows") and (args.build_via_generator):
```
footballhead, line 162: Did we want to keep the `args.build_via_generator` check? Otherwise we require people to use the Developer Command Prompt when they don't need to; Ran was asking recently if we could avoid requiring the developer command prompt.

scripts/build.py
```diff
@@ -136,38 +161,39 @@ def parse_actions(lis):
...
+    print(f"\nANDROID_NDK_HOME found: {os.environ["ANDROID_NDK_HOME"]}")
 
-    print("\nChecking cmake...")
-    cmd = [args.cmake_exec, "--version"]
+    cmd = [args.exec_cmake, "--version"]
     dive.echo_and_run(cmd)
 
-    print("\nChecking ninja...")
-    cmd = [args.ninja_exec, "--version"]
+    cmd = [args.exec_ninja, "--version"]
```
footballhead: nit: [For maximum reliability, use a fully qualified path for the executable](https://docs.python.org/3/library/subprocess.html#popen-constructor)


```suggestion
    cmd = [shutil.which(args.exec_ninja), "--version"]
```

scripts/build.py
```diff
@@ -136,38 +161,39 @@ def parse_actions(lis):
...
-def clean_build(args):
-    if os.path.exists(args.root_build_dir):
-        print("\nClearing build folder...")
-        shutil.rmtree(args.root_build_dir)
+    # Cannot check the flag --version because for some reason the return code is nonzero
+    if args.exec_deployqt:
+        deployqt_found = shutil.which(args.exec_deployqt)
+        if not deployqt_found:
+            raise Exception(f"{args.exec_deployqt} not found on the Path")
+        print(f"\n{args.exec_deployqt} found on the Path")
```
footballhead: Experimentally (I can't actually find any docs for this), `subprocess.run()` will raise `FileNotFoundError` if the program doesn't exist or `PermissionError` if the file isn't executable. So if we just want to verify that the program exists and is executable (regardless of what that program returns) we could still use `dive.echo_and_run` and swallow the `subprocess.CalledProcessError`

```suggestion
    if args.exec_deployqt:
        deployqt_found = shutil.which(args.exec_deployqt)
        if not deployqt_found:
            raise Exception(f"{args.exec_deployqt} not found on the Path")
        try:
            dive.echo_and_run([deployqt_found, "--version"])
        except subprocess.CalledProcessError:
            # --version for some reason returns non-zero
            pass
        # Other exceptions like FileNotFoundError and PermissionError will continue to propagate
```

scripts/build.py
```diff
@@ -283,19 +309,78 @@ def all_device(args):
...
+
+        case "Windows":
+            print(f"\nDeploying with {args.exec_deployqt}...")
+            cmd = [args.exec_deployqt, 
+                f"{args.root_build_dir}/{PKG_DIR}/host/{WINDOWS_QT_DEPENDENCIES_BINARY}"
+                ]
+            dive.echo_and_run(cmd)
+
+        case _:
+            raise Exception(f"Unrecognized platform: {platform.system()}")
```
footballhead: can store the result to avoid running the same function twice

```suggestion
        case unsupported_platform:
            raise Exception(f"Unrecognized platform: {unsupported_platform}")
```

scripts/build.py
```diff
@@ -283,19 +309,78 @@ def all_device(args):
...
+            final_location_zip = f"{args.root_build_dir}/{APP_DIR}.zip"
+            if os.path.exists(final_location_zip):
+                print("\nClearing previous archive...")
+                os.unlink(final_location_zip)
+
+            print(f"\nMoving zip archive to {os.getcwd()}/{final_location_zip}...")
+            shutil.move(f"{APP_DIR}.zip", final_location_zip)
+
+        case _:
+            raise Exception(f"Unrecognized platform: {platform.system()}")
```
footballhead:
```suggestion
        case unsupported_platform:
            raise Exception(f"Unrecognized platform: {unsupported_platform}")
```

scripts/build.py
angela28chen: This seems like an improvement that can be made later, I will note it on the bug

scripts/build.py
angela28chen: OK thanks, changed

scripts/build.py
angela28chen: oops, fixed

scripts/build.py
angela28chen: done

scripts/build.py
angela28chen: I changed it but not exactly like this. Is there a reason to specifically avoid the wildcard?

scripts/build.py
angela28chen: OK, added a check before using the exec in a command

scripts/build.py
footballhead: To avoid doing the same thing twice. Both wildcard and capture patterns are [irrefutable](https://docs.python.org/3/reference/compound_stmts.html#irrefutable-case-blocks) but capture patterns contain the result so you don't need to type it out again. If you change the match subject expr then this will also change

scripts/build.py
```diff
@@ -136,38 +161,45 @@ def parse_actions(lis):
...
+
     if "ANDROID_NDK_HOME" not in os.environ:
         raise Exception("ANDROID_NDK_HOME env var must be set")
+    print(f"\nANDROID_NDK_HOME found: {os.environ["ANDROID_NDK_HOME"]}")
 
-    print("\nChecking cmake...")
-    cmd = [args.cmake_exec, "--version"]
+    if not shutil.which(args.exec_cmake):
+        raise Exception(f"Cannot find cmake exec {args.exec_cmake}")
+    cmd = [args.exec_cmake, "--version"]
```
footballhead: Use the value of `which` in the cmd, not just as another check. This resolves the relative name to absolute which satisfies the warning for [subprocess.Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) (which subprocess.run uses)

```suggestion
    cmake = shutil.which(args.exec_cmake)
    if not cmake:
        raise Exception(f"Cannot find cmake exec {cmake}")
    cmd = [cmake, "--version"]
```

scripts/build.py
footballhead: The alternative works but my suggestion is more concise

scripts/build.py
angela28chen: OK, done

scripts/build.py
```diff
@@ -136,38 +161,53 @@ def parse_actions(lis):
...
+    print(f"\nANDROID_NDK_HOME found: {os.environ["ANDROID_NDK_HOME"]}")
 
-    print("\nChecking cmake...")
-    cmd = [args.cmake_exec, "--version"]
+    cmake = shutil.which(args.exec_cmake)
+    if not cmake:
+        raise Exception(f"Cannot find cmake exec {args.exec_cmake}")
+    cmd = [cmake, "--version"]
     dive.echo_and_run(cmd)
+    args.exec_cmake = cmake
```
footballhead, line 187: optional: since this is repeated several times, could consider refactoring it into a function

footballhead (approved):

scripts/build.py
angela28chen: I added this suggestion to the further improvements bug

elviscapiaq (approved):

PR #857, opened 2026-04-20
Upgrade NDK to r29

footballhead:
I had a choice between r27 (LTS), r29 (stable), and r30 (beta, soon-to-be LTS). I chose r29 since it will be closer to r30 when it's released. r30 is currently in beta so it's likely to release sooner than later.

Additionally, update the indentation for fenced code inside a list to align with examples provided by GitHub for GitHub Flavored Markdown: https://github.github.com/gfm/#example-298 . Notably, it uses a "hanging indent" of 2 spaces. This is rendered correctly in:

- the text diff viewer
- the rich diff viewer
- the final rendered file

Note that this doesn't impact the GFXR build. All GFXR build.gradle files use 26.3.11579264. This should only be an issue for Dive code that is built as part of the GFXR build (e.g. gfxr_decode_ext_lib, gpu_time).

### DIFF
--- a/.github/workflows/build_android.yml
+++ b/.github/workflows/build_android.yml
@@ -44,7 +44,8 @@ jobs:
       - uses: nttld/setup-ndk@ed92fe6cadad69be94a966a7ee3271275e62f779 # v1.6.0
         id: setup-ndk
         with:
-          ndk-version: r25
+          # TODO: b/504651150 - Prefer r30 (LTS) when it is relesed
+          ndk-version: r29
           add-to-path: true
       - name: (ccache) Clear stats
         run: ccache -z
--- a/BUILD.md
+++ b/BUILD.md
@@ -3,12 +3,15 @@
 - CMake
 - Ninja
 - The QT framework, can be installed from [QT online installer](https://download.qt.io/archive/online_installers/4.6/). We are currently using QT 5.15.2. Note that to install QT 5.15.2 from the online installer, you have to enable (turn on) the `archived` versions and then click on `filter`.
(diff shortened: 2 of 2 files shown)


angela28chen (approved):
This is interesting, I didn't realize you can build the GFXR stuff and Dive stuff with different versions of the android NDK. So you'd need both of them (29... 26...) installed to build Dive?

footballhead:
> This is interesting, I didn't realize you can build the GFXR stuff and Dive stuff with different versions of the android NDK. So you'd need both of them (29... 26...) installed to build Dive?

Yes both need to be installed. GFXR uses Gradle which will automatically download 26 (see [ndkVersion](https://developer.android.com/reference/tools/gradle-api/9.1/com/android/build/api/dsl/CommonExtension?_gl=1*1d99gl*_up*MQ..*_ga*MTAwMTA4MTA0OS4xNzc2NzE1ODk2*_ga_6HH9YJMN9M*czE3NzY3MTU4OTYkbzEkZzAkdDE3NzY3MTU4OTYkajYwJGwwJGgxMjE0OTMxMjYw#ndkVersion())) for GFXR. Since Dive doesn't use Gradle, you need to manually install it.

This is technically the case today as well, it's just that Dive uses 25 and GFXR uses 26. We could look at consolidating the versions but I'm not sure it matters a lot since it seems to build and run fine. And recent NDKs are built using the same clang tooling so it should all be compatible. That said, it should be as trivial as updating `ndkVersion` in the `build.gradle` file.

hysw (approved):

PR #862, opened 2026-04-23
Fix PM4 capture path

footballhead:
PM4 analysis of a capture is failing since the UI can't find the PM4 file.

PR #852 changed the path that the UI expects without changing the path that libwrap makes. This harmonizes the two.

This is a quick fix for the immediate issue. I'm going to spend more time on this issue and put up a better fix later.


angela28chen (approved):

hysw (approved):
Maybe put this in a variable somewhere?

footballhead:
> Maybe put this in a variable somewhere?

That's the idea. I have a plan for this in a future PR.

PR #864, opened 2026-04-23
[libwrap] Remove ineffectual build type override

footballhead:
CMAKE_BUILD_TYPE only applies to single-configuration generators (https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) so is being ignored by our use of multi-config generators.

Additionally, I may have solved the underlying issue in #863


GrantComm (approved):

footballhead:
Review https://github.com/google/dive/pull/867 instead

PR #865, opened 2026-04-23
Adds More Treeview Vulkan Command Info

GrantComm:
Adds more vulkan command info visible in the gfxr tree view. For draw calls, the instance count is displayed when the count is greater than 1. 
Screenshot 2026-04-28 at 2 56 37 PM

Screenshot 2026-04-30 at 9 23 31 AM

### DIFF
--- a/dive_core/gfxr_vulkan_command_hierarchy.cpp
+++ b/dive_core/gfxr_vulkan_command_hierarchy.cpp
@@ -49,29 +49,147 @@ void GfxrVulkanCommandHierarchyCreator::ConditionallyAddChild(uint64_t node_inde
     }
 }
 
-//--------------------------------------------------------------------------------------------------
-std::string GfxrVulkanCommandHierarchyCreator::GetCurrDrawCallString()
+std::string GfxrVulkanCommandHierarchyCreator::GetValueStr(const nlohmann::ordered_json& node,
+                                                           const std::string& key)
 {
-    // Early return for zero instance count
-    if (m_cur_draw_call_info.instance_count == 0)
+    if (node.contains(key))
     {
-        return "";
+        const auto& val = node[key];
+        if (val.is_string())
+        {
+            return val.get<std::string>();
+        }
+        std::ostringstream s;
+        s << val;
+        return s.str();
     }
+    return "0";
+}
 
+void GfxrVulkanCommandHierarchyCreator::AppendDrawCallSummary(std::ostringstream& s,
+                                                              const nlohmann::ordered_json& args,
+                                                              const std::string& primary_key)
+{
+    if (args.contains(primary_key))
+    {
+        s << "(" << primary_key << ": " << GetValueStr(args, primary_key);
+        if (args.contains("instanceCount"))
+        {
+            std::string instance_count = GetValueStr(args, "instanceCount");
+            if (instance_count != "1")
+            {
+                s << ", instanceCount: " << instance_count;
+            }
+        }
+        s << ")";
+    }
+}
+
+void GfxrVulkanCommandHierarchyCreator::AppendEnumSummary(std::ostringstream& s,
+                                                          const nlohmann::ordered_json& args,
+                                                          const std::string& key,
+                                                          const std::string& prefix)
+{
+    if (args.contains(key))
+    {
+        std::string val = GetValueStr(args, key);
+        size_t prefix_pos = val.find(prefix);
+        if (prefix_pos != std::string::npos)
+        {
+            val.erase(prefix_pos, prefix.length());
+        }
+        s << "(" << val << ")";
+    }
+}
+
+std::string GfxrVulkanCommandHierarchyCreator::GetCommandSummary(const std::string& cmd_name,
+                                                                 const nlohmann::ordered_json& args)
+{
     std::ostringstream s;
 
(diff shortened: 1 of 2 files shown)


wangra-google:
> Adds more vulkan command info visible in the gfxr tree view. For draw calls, the instance count is displayed when the count is greater than 1. Screenshot 2026-04-28 at 2 56 37 PM

sorry i dont see instance count in this screenshot. wrong screenshot?

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -49,29 +49,147 @@ void GfxrVulkanCommandHierarchyCreator::ConditionallyAddChild(uint64_t node_inde
...
+    }
+}
+
+std::string GfxrVulkanCommandHierarchyCreator::GetCommandSummary(const std::string& cmd_name,
+                                                                 const nlohmann::ordered_json& args)
+{
+    std::ostringstream s;
 
-    if (m_cur_draw_call_info.index_count > 0)
+    if (cmd_name == "vkCmdDrawIndexed")
```
wangra-google: i am slightly worried about the cpu cost on doing the compare to strings for each vk command. maybe better with std::unordered_map<std::string_view, std::function<...>> with lambda?

Something like:
```
using SummaryHandler = std::function<void(GfxrVulkanCommandHierarchyCreator*, 
                                          std::ostringstream&, 
                                          const nlohmann::ordered_json&)>;
static const std::unordered_map<std::string_view, SummaryHandler> handlers = {
        
        {"vkCmdDrawIndexed", [](auto* self, auto& s, const auto& args) {
            self->AppendDrawCallSummary(s, args, "indexCount");
        }},
...
};
...
auto it = handlers.find(std::string_view(cmd_name));
if (it != handlers.end())
{
    it->second(this, s, args);
}
```

dive_core/gfxr_vulkan_command_hierarchy.cpp
```diff
@@ -49,29 +49,147 @@ void GfxrVulkanCommandHierarchyCreator::ConditionallyAddChild(uint64_t node_inde
...
-    if (m_cur_draw_call_info.instance_count == 0)
+    if (node.contains(key))
     {
-        return "";
+        const auto& val = node[key];
+        if (val.is_string())
+        {
+            return val.get<std::string>();
+        }
+        std::ostringstream s;
```
wangra-google: would std::ostringstream allocate buffers internally which could be costly if we do that too many times? i am wondering if we could use `nlohmann::json::dump()` (just `return val.dump();` at the end)?

GrantComm: Done, also applied this change to the other sections of the file that used std::ostringstream

dive_core/gfxr_vulkan_command_hierarchy.cpp
GrantComm: Added the unordered map mapped to stringview keys

GrantComm:
> > Adds more vulkan command info visible in the gfxr tree view. For draw calls, the instance count is displayed when the count is greater than 1. Screenshot 2026-04-28 at 2 56 37 PM
> 
> sorry i dont see instance count in this screenshot. wrong screenshot?

Added another screenshot that shows an example with instance count

wangra-google (dismissed):

dive_core/gfxr_vulkan_command_hierarchy.h
```diff
@@ -77,6 +77,23 @@ class GfxrVulkanCommandHierarchyCreator
                   uint64_t child_node_index);
     void ConditionallyAddChild(uint64_t node_index);
 
+    // Helper function to extract and convert a value associated with a specific
+    // key from a JSON node into a string representation.
+    std::string GetValueStr(const nlohmann::ordered_json& node, const std::string& key);
```
angela28chen: Optional: I feel like these functions can just be at the top of gfxr_vulkan_command_hierarchy.cpp and not be a class method. They don't affect any of the class members or use any of the class structs, right?

angela28chen: You can also consider making a separate helper class whose purpose is just to parse JSON of a vulkan command into a string summary (`VulkanCommandJsonParser` or something)? It can have `std::ostringstream` as a class member, and then private methods `AppendDrawCallSummary()` and `AppendEnumSummary()`, with some public methods to parse JSON args and retrieve the result string, and/or clear the stringstream.

I'm mainly just suggesting this because if methods don't necessitate being in a class, it's a good idea to separate it out for more modular code.

angela28chen: Otherwise if you're keeping them here they should probably be marked with `const`

angela28chen (dismissed):

dive_core/gfxr_vulkan_command_hierarchy.h
GrantComm: Correct, they do not modify the class members, or structs. I moved them to the top of gfxr_vulkan_command_hierarchy.cpp and removed them from the class.

angela28chen (approved):

wangra-google (approved):

PR #866, opened 2026-04-23
Add --archive-name flag to build script for Windows build

angela28chen:
Archive name can be specified, and the default value uses the host tools version string

e.g. "dive-1.2.0-dev-win-f45a293.zip"

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -183,7 +183,7 @@ def check_environment(args):
 
     if "ANDROID_NDK_HOME" not in os.environ:
         raise Exception("ANDROID_NDK_HOME env var must be set")
-    print(f"\nANDROID_NDK_HOME found: {os.environ["ANDROID_NDK_HOME"]}")
+    print(f"\nANDROID_NDK_HOME found: {os.environ['ANDROID_NDK_HOME']}")
 
     cmake = shutil.which(args.exec_cmake)
     if not cmake:
@@ -382,7 +382,7 @@ def get_archive_name(args, unparsed_string):
     if not match_obj:
         raise Exception(f"Unsupported format of version string:\n{unparsed_string}")
 
-    print(f"Found host verison: {match_obj.group()}")
+    print(f"Found host version: {match_obj.group()}")
 
     return "dive-" + match_obj.group()
 
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -19,6 +19,7 @@ import common_dive_utils as dive
 import enum
 import os
 import platform
+import re
 import shutil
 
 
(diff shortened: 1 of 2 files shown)


scripts/build.py
```diff
@@ -382,17 +405,21 @@ def package(args):
         case "Windows":
             for dir_name in WINDOWS_UNNECESSARY_DIRS:
                 rmtree_if_exists(f"{args.root_build_dir}/{PKG_DIR}/host/{dir_name}")
+
+            cmd = [f"{args.root_build_dir}/{PKG_DIR}/host/dive_client_cli.exe", "--version"]
+            long_version_string = dive.run_and_return_output(cmd)
+            archive_name = get_archive_name(args, long_version_string)
```
footballhead, line 411: if `args.archive_name` is provided then the cmd is run but the output is immediately discarded. Is this intentional? You can avoid doing unnecessary work by checking for args.archive name here, instead of in the helper

```suggestion
            if args.archive_name:
                archive_name = args.archive_name
            else:
                cmd = [f"{args.root_build_dir}/{PKG_DIR}/host/dive_client_cli.exe", "--version"]
                long_version_string = dive.run_and_return_output(cmd)
                # Remove the args.archive_name check from this function
                archive_name = get_archive_name(long_version_string)
```

angela28chen: Hmm I think that having the version checked & thrown into stdout is useful regardless of if --archive-name is provided or not. I will change it to this now.

Also having the helper function check --archive-name was because I thought it'd be the same for linux/mac/windows implementation

angela28chen: Changed

scripts/build.py
```diff
@@ -363,6 +369,23 @@ def deploy_qt(args):
...
+    if args.archive_name:
+        return args.archive_name
+
+    pattern = r"(?<=" + re.escape(HOST_TOOLS_BUILD_NAME) + r"\:\s)\S+"
+    match_obj = re.search(pattern, unparsed_string)
+
+    if not match_obj:
+        raise Exception(f"Unsupported format of version string:\n{unparsed_string}")
+
+    print(f"Found host verison: {match_obj.group()}")
```
elviscapiaq: Typo:I think it is intended to say `version`.

scripts/build.py
elviscapiaq: Not related to this PR, but note that reusing double quotes inside a f-string expression is only valid in Python 3.12 and newer. Maybe we can have minimum python requirement or just change this single quotes to be safe?
` print(f"\nANDROID_NDK_HOME found: {os.environ['ANDROID_NDK_HOME']}")`

footballhead (dismissed):

scripts/build.py
angela28chen: I see, nice catch! I changed it to single quotes

elviscapiaq (approved):

footballhead (approved):

PR #876, opened 2026-04-27
Kokoro windows build fix.

hysw:
- Make sure CMake pick up python3.11 or higher.
- Do not spawn long running gradle daemon since that confuses python subprocess.


footballhead (dismissed):

scripts/build.py
```diff
@@ -140,6 +140,10 @@ def parse_args():
         "--visual-studio-name",
         default="Visual Studio 17 2022",
         help="Name of Visual Studio version to use on Windows platform")
+    parser.add_argument(
+        "--package-output-name",
```
angela28chen: I think this overlaps with the existing flag `--archive-name`. Priority is not clear and there is a difference in naming. If the intent is to change the output folder, maybe `--archive-dir` instead?

hysw: Reverted build.py change.

elviscapiaq (approved):

CMakeLists.txt
```diff
@@ -334,7 +262,7 @@ if(ANDROID)
     execute_process(
         COMMAND
             ${GRADLE_EXEC} layer:assembleDebug replay:assembleDebug
-            --console=${DIVE_GFXR_GRADLE_CONSOLE} -P${ANDROID_ABI}
+            --console=${DIVE_GFXR_GRADLE_CONSOLE} -P${ANDROID_ABI} --no-daemon
```
footballhead, line 339: Since the gradle daemon [improves build times](https://docs.gradle.org/current/userguide/gradle_daemon.html), what is the impact of not using it? This changes both the CI build and dev builds. Should it be configurable?

hysw: It should probably be configurable. Will add that later once CI build is setup properly. I don't want to do it now since there's still changes to all the CI scripts at the moment.

footballhead (approved):

angela28chen (approved):

PR #877, opened 2026-04-27
Implement linux deploy and package steps in the build script

angela28chen:
Other minor changes:
- Added `--archive-dir` flag so that users can specify an output directory for the archive
- Rearranged flags alphabetically

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -356,7 +356,7 @@ def deploy_qt(args):
     system_name = platform.system()
     match system_name:
         case "Linux":
-            print("\ndeploy_qt() is UNIMPLEMENTED for Linux...")
+            print("\ndeploy_qt() is a no-op for Linux, it's expected that the users install Qt...")
             return
 
         case "Darwin":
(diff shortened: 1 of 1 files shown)


scripts/build.py
```diff
@@ -410,25 +409,25 @@ def package(args):
...
-            shutil.move(f"{archive_name}.zip", final_location_zip)
 
         case _:
             raise Exception(f"Unrecognized platform: {system_name}")
 
+    long_version_string = dive.run_and_return_output(cmd)
+    archive_name = get_archive_name(args, long_version_string)
+
+    print(f"\nZipping into '{archive_name}.zip'...")
+    shutil.make_archive(f"{archive_name}", "zip", f"{args.root_build_dir}/{PKG_DIR}")
```
elviscapiaq: Should we directly use archive_name as it is already a string?

elviscapiaq (dismissed):
Just a minor comment.

scripts/build.py
angela28chen: Done

elviscapiaq (approved):

hysw (approved):

PR #878, opened 2026-04-27
Package libwrap into GFXR replay

footballhead:
In my quest to remove hardcoded paths that are copy-pasted in multiple locations, I found that wrap.sh can refer to binaries packaged inside the application. By packaging libwrap into the replay, we

- simplify deployment: don't need to push libwrap.so. This will be tackled in a later PR.
- improve compatibility: theoretically don't need root access anymore for replay with PM4 dump. This will be tackled in a later PR.
- improve maintainability: eliminates bugs like PR #705

The easiest way to add libwrap to the app is to rebuild it as part of the GFXR build. While this means that the lib is being built twice, the extra time should not be noticable since libwrap is so small.

What could be more interesting though is that GFXR is built for debug so this extra copy is built for debug as well. Given that libwrap was originally built for debug (see PR #867), this shouldn't be an issue.

### DIFF
--- a/third_party/gfxreconstruct/android/tools/replay/src/main/resources/lib/arm64-v8a/wrap.sh
+++ b/third_party/gfxreconstruct/android/tools/replay/src/main/resources/lib/arm64-v8a/wrap.sh
@@ -35,7 +35,7 @@ fi
 
 echo "capture_pm4_enabled is " $capture_pm4_enabled
 if [[ "$capture_pm4_enabled" == "true" || "$capture_pm4_enabled" == "1" ]]; then
-    export LD_PRELOAD=/data/local/tmp/dive/libwrap.so
+    export LD_PRELOAD="$(dirname $0)/libwrap.so"
 fi
 
 exec $cmd
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -542,6 +542,7 @@ absl::Status AndroidDevice::SetupDevice()
 
     if (!m_gfxr_capture_settings)
     {
+        // TODO: b/506128425 - We shouldn't need root for replay
         RETURN_IF_ERROR(RequestRootAccess());
         RETURN_IF_ERROR(DeployDeviceResource(Dive::DeviceResourcesConstants::kVkLayerLibName));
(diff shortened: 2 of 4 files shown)


third_party/gfxreconstruct/android/tools/replay/src/main/resources/lib/arm64-v8a/wrap.sh
```diff
@@ -35,7 +35,7 @@ fi
 
 echo "capture_pm4_enabled is " $capture_pm4_enabled
 if [[ "$capture_pm4_enabled" == "true" || "$capture_pm4_enabled" == "1" ]]; then
-    export LD_PRELOAD=/data/local/tmp/dive/libwrap.so
+    export LD_PRELOAD="$(dirname $0)/libwrap.so"
```
angela28chen, line 38: neat trick!

angela28chen (approved):
Just to confirm, the points mentioned in the PR description about not needing to use root access and not needing to push libwrap.so, those are intended for followup PRs and not included in this one?

footballhead:
> Just to confirm, the points mentioned in the PR description about not needing to use root access and not needing to push libwrap.so, those are intended for followup PRs and not included in this one?

Yes. These are more complicated to tackle. I didn't want to unnecessarily increase the size of this PR.

angela28chen:
> > Just to confirm, the points mentioned in the PR description about not needing to use root access and not needing to push libwrap.so, those are intended for followup PRs and not included in this one?
> 
> Yes. These are more complicated to tackle. I didn't want to unnecessarily increase the size of this PR.

ok, nit: can you tweak the PR description to make it clearer what this PR does VS what it paves the way for doing in the future? Maybe something like "By packaging libwrap into the replay, in the future we can:"

footballhead:
> > > Just to confirm, the points mentioned in the PR description about not needing to use root access and not needing to push libwrap.so, those are intended for followup PRs and not included in this one?
> > 
> > 
> > Yes. These are more complicated to tackle. I didn't want to unnecessarily increase the size of this PR.
> 
> ok, nit: can you tweak the PR description to make it clearer what this PR does VS what it paves the way for doing in the future? Maybe something like "By packaging libwrap into the replay, in the future we can:"

Done

GrantComm (approved):

PR #883, opened 2026-04-30
Cleans Up Analyze Dialog

GrantComm:
- Default-initialize members
- Mark constructor as explicit to avoid accidental conversions (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rc-explicit)
- Mark destructor with override (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rh-override)

### DIFF
--- a/ui/analyze_dialog.cpp
+++ b/ui/analyze_dialog.cpp
@@ -70,6 +70,10 @@ void AttemptDeletingTemporaryLocalFile(const std::filesystem::path& file_path)
     }
 }
 
+constexpr int kDataRole = Qt::UserRole + 1;
+constexpr int kDefaultFrameCount = 300;
+constexpr const char* kDefaultReplayButtonText = "Replay";
+
 }  // namespace
 
 // =================================================================================================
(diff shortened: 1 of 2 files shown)


ui/analyze_dialog.cpp
```diff
@@ -70,6 +70,10 @@ void AttemptDeletingTemporaryLocalFile(const std::filesystem::path& file_path)
     }
 }
 
+constexpr int kDataRole = Qt::UserRole + 1;
+constexpr int kDefaultFrameCount = 300;
+constexpr const char* kDefaultReplayButtonText = "Replay";
```
hysw: `string_view`

ui/analyze_dialog.cpp
```diff
@@ -167,15 +171,15 @@ AnalyzeDialog::AnalyzeDialog(ApplicationController& controller,
 
     // Enable GPU Time
```
hysw, line 157: At this point, what is preventing you from having something like

```cpp
QHBoxLayout* CreateGpuTimeLayout() {
  // ...
}
```

footballhead (dismissed):
LGTM once peter's comments are addressed

ui/analyze_dialog.cpp
GrantComm: Done, also modified the functions that used std::string to stringview

ui/analyze_dialog.cpp
GrantComm: Added CreateFrameCountGroupBox that is used to create m_gpu_time_replay_box and m_custom_replay_box

hysw (approved):

footballhead (approved):

PR #885, opened 2026-05-04
Fixes variable shadowing

GrantComm:
Fixes variable shadowing bug introduced in #879 which was causing a crash when advanced options are selected.

### DIFF
--- a/ui/trace_dialog.cpp
+++ b/ui/trace_dialog.cpp
@@ -167,7 +167,7 @@ TraceDialog::TraceDialog(ApplicationController& controller, QWidget* parent) : D
 
     // Capture Debuggable Applications Only Warning
     auto* capture_warning_layout = new QHBoxLayout();
-    auto* m_capture_warning_label = new QLabel(tr(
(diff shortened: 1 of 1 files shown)


footballhead:
Is there a clang-tidy rule for this that we can turn on?

footballhead (approved):

GrantComm:
> Is there a clang-tidy rule for this that we can turn on?

I think the clang-diagnostic-shadow (-Wshadow) rule is the one for this: 
https://releases.llvm.org/20.1.0/tools/clang/docs/DiagnosticsReference.html#wshadow

wangra-google (approved):

footballhead:
> > Is there a clang-tidy rule for this that we can turn on?
> 
> I think the clang-diagnostic-shadow (-Wshadow) rule is the one for this: https://releases.llvm.org/20.1.0/tools/clang/docs/DiagnosticsReference.html#wshadow

Thanks. Looks like I want -Wshadow-all which is not on by default. I'll fix that in a future PR

PR #888, opened 2026-05-05
Prioritizes active variant name over generic name

GrantComm:
Modifies the generation script to ensure the generic register name is not overwritten by variants and prioritize the active GPU variant name over the generic name. This ensures that we always find the correct offset for the current capture. This bug resulted in the trace stats not being fully populated, resulting in them not being displayed on Windows.

Manual Testing:

**Linux**

Open gfxr capture file, trace stats are loaded and displayed

**Windows**

Open gfxr capture file, trace stats are loaded and displayed

### DIFF
--- a/dive_core/generatePm4Info_adreno.py
+++ b/dive_core/generatePm4Info_adreno.py
@@ -914,12 +914,9 @@ def outputPacketInfo(pm4_info_file, registers_et_root, enum_index_dict, opcode_d
 		uint32_t gpu_variants = reg.first << shift_bits >> shift_bits;
 		uint32_t reg_offset = reg.first >> kGPUVariantsBits;
 
-    // Only set the generic name if it hasn't been set yet. 
-    // This prevents later variants from overwriting the base/default offset.
-    if (g_sRegNameToIndex.find(name) == g_sRegNameToIndex.end())
-    {
-        g_sRegNameToIndex[name] = reg_offset;
-    }
+		// Only set the generic name if it hasn't been set yet.
+		// This prevents later variants from overwriting the base/default offset.
+		g_sRegNameToIndex.insert({name, reg_offset});
 
 		if (gpu_variants != 0)
 		{
@@ -1002,15 +999,13 @@ uint32_t GetRegOffsetByName(const char *name)
 
     std::string str = std::string(name);
     std::string name_with_variant = str + "_" + GetGPUStr(g_sGPU_variant);
-    auto i = g_sRegNameToIndex.find(name_with_variant);
-    if (i != g_sRegNameToIndex.end())
+    if (auto i = g_sRegNameToIndex.find(name_with_variant); i != g_sRegNameToIndex.end())
     {
         return i->second;
     }
 
     // Fall back to the generic name if no variant match exists
-    i = g_sRegNameToIndex.find(str);
-    if (i != g_sRegNameToIndex.end())
+    if (auto i = g_sRegNameToIndex.find(str); i != g_sRegNameToIndex.end())
     {
         return i->second;
     }
--- a/dive_core/generatePm4Info_adreno.py
+++ b/dive_core/generatePm4Info_adreno.py
@@ -913,7 +913,14 @@ def outputPacketInfo(pm4_info_file, registers_et_root, enum_index_dict, opcode_d
 		const uint32_t shift_bits = 32 - kGPUVariantsBits;
 		uint32_t gpu_variants = reg.first << shift_bits >> shift_bits;
 		uint32_t reg_offset = reg.first >> kGPUVariantsBits;
-		g_sRegNameToIndex[name] = reg_offset;
+
+    // Only set the generic name if it hasn't been set yet. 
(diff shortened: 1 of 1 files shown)


wangra-google (dismissed):

dive_core/generatePm4Info_adreno.py
```diff
@@ -913,7 +913,14 @@ def outputPacketInfo(pm4_info_file, registers_et_root, enum_index_dict, opcode_d
...
 		uint32_t gpu_variants = reg.first << shift_bits >> shift_bits;
 		uint32_t reg_offset = reg.first >> kGPUVariantsBits;
-		g_sRegNameToIndex[name] = reg_offset;
+
+    // Only set the generic name if it hasn't been set yet. 
+    // This prevents later variants from overwriting the base/default offset.
+    if (g_sRegNameToIndex.find(name) == g_sRegNameToIndex.end())
+    {
+        g_sRegNameToIndex[name] = reg_offset;
+    }
```
footballhead: avoid redundant lookups by using [insert()](https://cppreference.com/cpp/container/unordered_map/insert). insert never assigns; it only mutates the map if the key doesn't already exist

```c++
g_sRegNameToIndex.insert({name, reg_offset});
```

wangra-google:
Thanks for fixing this! Without your fix, it was purely on luck if the generic name slot would get the correct offset. My guess is that since on Linux the iteration order is different from Windows, the correct offset is accidentally put to the generic name slot which hides this bug.

dive_core/generatePm4Info_adreno.py
```diff
@@ -994,17 +1001,20 @@ def outputFunctionsCpp(pm4_info_file):
     }
 
     std::string str = std::string(name);
-    auto i = g_sRegNameToIndex.find(str);
-    if (i == g_sRegNameToIndex.end())
+    std::string name_with_variant = str + "_" + GetGPUStr(g_sGPU_variant);
+    auto i = g_sRegNameToIndex.find(name_with_variant);
+    if (i != g_sRegNameToIndex.end())
```
footballhead: nit: can use if-with-initializer to limit the scope of the iterator

```
if (auto i = g_sRegNameToIndex.find(name_with_variant); i != g_sRegNameToIndex.end())
{
    return i->second;
}
```

dive_core/generatePm4Info_adreno.py
```diff
@@ -994,17 +1001,20 @@ def outputFunctionsCpp(pm4_info_file):
...
-        if (i == g_sRegNameToIndex.end())
-        {
-            return kInvalidRegOffset;
-        }
+        return i->second;
+    }
+
+    // Fall back to the generic name if no variant match exists
+    i = g_sRegNameToIndex.find(str);
+    if (i != g_sRegNameToIndex.end())
```
footballhead: nit: can use if-with-initializer to limit the scope of the iterator

```
if (auto i = g_sRegNameToIndex.find(str); i != g_sRegNameToIndex.end())
{
    return i->second;
}
```

dive_core/generatePm4Info_adreno.py
```diff
@@ -994,17 +1001,20 @@ def outputFunctionsCpp(pm4_info_file):
     }
 
     std::string str = std::string(name);
-    auto i = g_sRegNameToIndex.find(str);
-    if (i == g_sRegNameToIndex.end())
+    std::string name_with_variant = str + "_" + GetGPUStr(g_sGPU_variant);
+    auto i = g_sRegNameToIndex.find(name_with_variant);
```
footballhead: +1 for reducing nesting!

dive_core/generatePm4Info_adreno.py
```diff
@@ -913,7 +913,14 @@ def outputPacketInfo(pm4_info_file, registers_et_root, enum_index_dict, opcode_d
 		const uint32_t shift_bits = 32 - kGPUVariantsBits;
 		uint32_t gpu_variants = reg.first << shift_bits >> shift_bits;
 		uint32_t reg_offset = reg.first >> kGPUVariantsBits;
-		g_sRegNameToIndex[name] = reg_offset;
+
+    // Only set the generic name if it hasn't been set yet.
```
footballhead: nit: the indentation looks off between this and the surrounding code. is this a tabs vs  space issue?

GrantComm: Fixed the indentation.

dive_core/generatePm4Info_adreno.py
GrantComm: Done, added if-with-initializer

dive_core/generatePm4Info_adreno.py
GrantComm: Done, added if-with-initializer

dive_core/generatePm4Info_adreno.py
GrantComm: Done, uses insert now

footballhead (approved):

wangra-google (approved):

PR #889, opened 2026-05-05
Update scripts/build.py

hysw:
- Change ActionType into Action object
  - Make it possible to update list of actions for other script
- Use append argument for host configure args
  - It require less quoting
- Run autopep8
- Update quote formats

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -161,7 +161,7 @@ def parse_args():
         action="append",
         default=[],
         help="String of additional CMake flags to pass directly to cmake in the "
-        "configure_host stage",
+        "configure_host action",
     )
     parser.add_argument(
         "--host-configure-additional-flags", default="", help="Deprecated"
@@ -259,7 +259,7 @@ def check_environment(args):
 
 
 def copy_plugins(args):
-    """Implements COPY_PLUGINS stage"""
+    """Implements COPY_PLUGINS action"""
 
     if not os.path.exists(PROFILING_PLUGIN_PATH):
         print(f"\nDir {PROFILING_PLUGIN_PATH} does not exist, skipping")
@@ -276,7 +276,7 @@ def copy_plugins(args):
 
 
 def configure_host(args):
-    """Implements CONFIGURE_HOST stage, uses cmake to generate
+    """Implements CONFIGURE_HOST action, uses cmake to generate
     build files targeting the host
     """
     print("\nGenerating build files with cmake...")
@@ -294,7 +294,7 @@ def configure_host(args):
 
 
 def build_host(args):
-    """Implementing BUILD_HOST stage"""
+    """Implementing BUILD_HOST action"""
     if not args.build_via_generator:
         print("\nBuilding host libraries by invoking cmake...")
         cmd = [
@@ -332,7 +332,7 @@ def build_host(args):
 
 
 def install_host(args):
-    """Implementing INSTALL_HOST stage"""
+    """Implementing INSTALL_HOST action"""
     # TODO: b/482108095 - This must be done regardless of clean build because
     # install folders are re-used between different build types and the
     # timestamps could affect whether files within are correctly replaced or not
@@ -352,7 +352,7 @@ def install_host(args):
 
 
 def all_device(args):
-    """Implementing ALL_DEVICE stage"""
+    """Implementing ALL_DEVICE action"""
     print("\nGenerating build files with cmake...")
     android_ndk_home = os.environ["ANDROID_NDK_HOME"]
     cmd = [
@@ -402,7 +402,7 @@ def all_device(args):
 
 
 def deploy_qt(args):
-    """Implements DEPLOY_QT stage, which uses a QT tool to add QT
+    """Implements DEPLOY_QT action, which uses a QT tool to add QT
     libraries to the same dir as the binary that depends on them in preparation
     for packaging and distribution
     """
@@ -461,7 +461,7 @@ def get_archive_name(args, unparsed_string):
 
 
 def package(args):
-    """Implements PACKAGE stage, which will package this Dive build
+    """Implements PACKAGE action, which will package this Dive build
     into a single file that is easy to transfer
     """
     # TODO: b/504654590 - Investigate if there's a way to identify extraneous files
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -22,7 +22,6 @@ import platform
 import re
 import shutil
 
-
 # GLOBAL CONSTANTS
 PROFILING_PLUGIN_PATH = "plugins/external/dive_profiling_plugin"
 PKG_DIR = "pkg"
@@ -38,12 +37,12 @@ HOST_TOOLS_BUILD_NAME = "Host Tools Build"
 
 class ActionType(enum.StrEnum):
     CONFIGURE_HOST = enum.auto()
-    BUILD_HOST = enum.auto()        # separated to make VS builds using the UI easy
+    BUILD_HOST = enum.auto()  # separated to make VS builds using the UI easy
     INSTALL_HOST = enum.auto()
-    ALL_DEVICE = enum.auto()        # configure, build, and install device libraries
-    COPY_PLUGINS = enum.auto()      # copy external plugins into pkg dir
+    ALL_DEVICE = enum.auto()  # configure, build, and install device libraries
+    COPY_PLUGINS = enum.auto()  # copy external plugins into pkg dir
     DEPLOY_QT = enum.auto()
-    PACKAGE = enum.auto()           # removes some extraneous files and packages the release
+    PACKAGE = enum.auto()  # removes some extraneous files and packages the release
 
 
 class BuildType(enum.StrEnum):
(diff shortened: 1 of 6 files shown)


scripts/build.py
```diff
@@ -135,16 +142,15 @@ def parse_args():
...
-        "--host-configure-additional-flags",
+        "--host-configure-flag",
+        action="append",
+        default=[],
         help="String of additional CMake flags to pass directly to cmake in the "
         "configure_host stage")
+    parser.add_argument("--host-configure-additional-flags",
+                        default="",        help="Deprecated")
     parser.add_argument(
-        "--list-actions",
```
angela28chen, line 142: Is it intentional to remove `--list-actions`? I think this is a useful flag and there's still a lot of references to it in this file or maybe BUILD.md.

scripts/build.py
```diff
@@ -273,22 +266,24 @@ def configure_host(args):
         f"-DDIVE_RELEASE_TYPE={args.dive_release_type}",
         f"-DDIVE_STR_BUILD={args.root_build_dir}",
     ]
-    if args.host_configure_additional_flags:
-        for arg in args.host_configure_additional_flags.split(" "):
-            if arg:
-                cmd.append(arg)
+    cmd.extend(args.host_configure_flag)
+    cmd.extend(filter(None, args.host_configure_additional_flags.split(" ")))
```
angela28chen, line 292: If `--host-configure-additional-flags` is still functional, I think it makes more sense to mark it as "to be deprecated" rather than "deprecated" in the help message

scripts/build.py
```diff
@@ -273,22 +266,24 @@ def configure_host(args):
...
     if not args.build_via_generator:
         print("\nBuilding host libraries by invoking cmake...")
         cmd = [args.exec_cmake,
                "--build", f"{args.root_build_dir}/host",
                "--config", f"{args.build_type}"
                ]
+        # Force cmake to run parallel build on Windows.
+        # Linux/macOS uses ninja, which do parallel builds by default.
+        if platform.system() == "Windows":
+            cmd.append(f"-j{os.cpu_count()}")
```
angela28chen, line 310: I think this should be added to the PR description bullet list

Also, did you verify that this works on Windows? Reading the comments I see this is intended for use with VS generator, I thought that this flag doesn't work properly to do a parallel build.

scripts/build.py
```diff
@@ -481,27 +493,8 @@ def main():
...
-            with dive.Timer(ActionType.COPY_PLUGINS):
-                copy_plugins(args)
-        if ActionType.DEPLOY_QT in args.actions:
-            with dive.Timer(ActionType.DEPLOY_QT):
-                deploy_qt(args)
-        if ActionType.PACKAGE in args.actions:
-            with dive.Timer(ActionType.PACKAGE):
-                package(args)
+        for action in actions:
+            action(args)
```
angela28chen: Is the separate timer logic gone? IMO that is pretty useful to see in the output

angela28chen (commented):
Wow, I learned a lot about newer python features. 

Also, if you want to have a standard tool/process for formatting python files within this repo, we should probably at minimum put some info about it inside CONTRIBUTING.md.

scripts/build.py
hysw: ~`build.py --help` already prints the list~

looks like I have the wrong version on github

scripts/build.py
hysw: It might depend on CMake version

hysw:
> Also, if you want to have a standard tool/process for formatting python files within this repo, we should probably at minimum put some info about it inside CONTRIBUTING.md.

It's part of python standard process PEP8 (Python Enhancement Proposal 8), see  https://peps.python.org/pep-0008/

Many tools will apply that: autopep8 / [black](https://github.com/psf/black) / [pyink](https://github.com/google/pyink)

scripts/build.py
hysw: It's deprecated, just that it will be removed after I fix the kokoro builds stuff

scripts/build.py
hysw: So, make argparse do the checking

```log
$ python3 scripts/build.py --action=not_found 
usage: build_android [-h] [--action {configure_host,build_host,install_host,all_device,copy_plugins,deploy_qt,package}] [--actions ACTIONS]
                     [--archive-dir ARCHIVE_DIR] [--archive-name ARCHIVE_NAME] [--build-type {Debug,RelWithDebInfo,Release}]
                     [--build-via-generator | --no-build-via-generator] [--ci | --no-ci] [--dive-release-type DIVE_RELEASE_TYPE]
                     [--exec-cmake EXEC_CMAKE] [--exec-codesign EXEC_CODESIGN] [--exec-deployqt EXEC_DEPLOYQT] [--exec-msbuild EXEC_MSBUILD]
                     [--exec-ninja EXEC_NINJA] [--host-configure-flag HOST_CONFIGURE_FLAG]
                     [--host-configure-additional-flags HOST_CONFIGURE_ADDITIONAL_FLAGS] [--mac-sign | --no-mac-sign]
                     [--prereq-checks | --no-prereq-checks] [--root-build-dir ROOT_BUILD_DIR] [--host-cmake-generator HOST_CMAKE_GENERATOR]
build_android: error: argument --action: invalid choice: 'not_found' (choose from configure_host, build_host, install_host, all_device, copy_plugins, deploy_qt, package)
```

scripts/build.py
hysw: Fixed.

scripts/build.py
angela28chen: LGTM thanks!

scripts/build.py
angela28chen: Yes, I just mentioned it because I assume you're going to come back to this python script and remove this line after that, which is when I think it makes sense to change "to be deprecated" -> "deprecated"

But it's a nitpick so either way is fine

scripts/build.py
```diff
@@ -165,33 +174,20 @@ def parse_args():
...
-            actions.append(action)
-    else:
-        actions = [x for x in ActionType]
-
-    str_actions = ", ".join(actions)
-    print(f"\nQueued actions, not necessarily listed in order: {str_actions}")
-
-    args.actions = actions
+    # Getting an ordered list of actions
+    return [action for action in available_actions if str(action) in actions]
```
footballhead: nit:  Consider using:

- `allowed_action_names` instead of `allowed_actions`
- `action_names` instead of `actions`.

Since this function deals with both actions as strings and as `Action` objects, it might be less confusing when we're dealing with `Action` objects later.

```suggestion
    allowed_action_names = set(str(action) for action in available_actions)
    action_names = set(args.action) | set(filter(None, args.actions.split(',')))
    for action_name in action_names:
        if action_name not in allowed_action_names:
            raise Exception(f"Invalid action specified: '{action_name}'")
    if not action_names:
        action_names = allowed_action_names

    # Getting an ordered list of actions
    return [action for action in available_actions if str(action) in action_names]
```

scripts/build.py
```diff
@@ -36,14 +37,19 @@
...
-    CONFIGURE_HOST = enum.auto()
-    BUILD_HOST = enum.auto()        # separated to make VS builds using the UI easy
-    INSTALL_HOST = enum.auto()
-    ALL_DEVICE = enum.auto()        # configure, build, and install device libraries
-    COPY_PLUGINS = enum.auto()      # copy external plugins into pkg dir
-    DEPLOY_QT = enum.auto()
-    PACKAGE = enum.auto()           # removes some extraneous files and packages the release
+class Action:
+    def __init__(self, name: str, func: typing.Callable[[argparse.Namespace], None]):
+        self.name = name.lower()
```
footballhead: why not just create the action with a lower case name?

```suggestion
        self.name = name
```

scripts/build.py
```diff
@@ -165,33 +174,20 @@ def parse_args():
     return parser.parse_args()
 
 
-def list_actions():
-    """Lists all of ActionType in the order they would be executed
+def parse_actions(args) -> list[str]:
+    """Convert args.actions to an ordered list of actions
```
footballhead: nit: consider using descriptive style to be consistent

```suggestion
    """Converts args.actions to an ordered list of actions
```

scripts/build.py
footballhead, line 206: optional: might be able to express more succinctly using `filter`

scripts/build.py
hysw: Applied, and added an assert check.

scripts/build.py
hysw: Done.

scripts/build.py
hysw: List comprehension and filter seems to be similar complexity in this one. The `filter(None, ...)` case is probably slightly cleaner.

scripts/build.py
```diff
@@ -245,8 +259,7 @@ def check_environment(args):
 
 
 def copy_plugins(args):
-    """Implements ActionType.COPY_PLUGINS stage
-    """
+    """Implements COPY_PLUGINS stage"""
```
footballhead: nit: consider using "action" instead of stage to match the Action object and the `--action` commandline argument

```suggestion
    """Implements COPY_PLUGINS action"""
```

footballhead (dismissed):
In the future, consider formatting in a separate CL so that it's easier for me to review the pure formatting changes compared to the functionality changes.

hysw:
> In the future, consider formatting in a separate CL so that it's easier for me to review the pure formatting changes compared to the functionality changes.

Formatting should be in a separate commit already?

scripts/build.py
hysw: Done.

footballhead (approved):

angela28chen (approved):

PR #893, opened 2026-05-07
Symbol exporting for the DeviceManager singleton

GrantComm:
Adds symbol exporting for the DeviceManager singleton on Windows. This prevents the plugin from using a different instance from the host on windows.


capture_service/CMakeLists.txt
```diff
@@ -102,6 +99,11 @@ else()
         dive_renderdoc_files
     )
 
+    if(MSVC)
+        target_compile_definitions(device_mgr PUBLIC CAPTURE_SERVICE_BUILD)
```
wangra-google: Shouldn't here be PRIVATE instead of PUBLIC? Otherwise CAPTURE_SERVICE_BUILD will be defined for other targets that link against device_mgr?

capture_service/capture_service_export.h
wangra-google, line 1: i am wondering if this could be auto generated by adding these in the cmake file:
```
include(GenerateExportHeader)

generate_export_header(device_mgr
    EXPORT_MACRO_NAME CAPTURE_SERVICE_EXPORT
    EXPORT_FILE_NAME ${CMAKE_CURRENT_BINARY_DIR}/capture_service_export.h
)
```

capture_service/CMakeLists.txt
```diff
@@ -79,10 +79,7 @@ if(ANDROID)
     target_link_libraries(service PRIVATE wrap)
 else()
     add_library(device_mgr device_mgr.cc android_application.cc)
```
wangra-google: just to make sure i understand, do we build static or dynamic lib here?

GrantComm: By default, it is static

https://cmake.org/cmake/help/latest/command/add_library.html

capture_service/capture_service_export.h
GrantComm: Added this and deleted the capture_service_export.h file

capture_service/CMakeLists.txt
wangra-google: sorry i am a bit confused, if it is static, what is the point of having capture_service_export.h?

GrantComm: Statically linking it into both dive_gui_lib.dll and system_trace.dll results in two completely separate instances of device_mgr, which resulted in the crash. 

To fix the crash, both DLLs needed share a single instance of DeviceManager. This requires compiling device_mgr as a shared library on Windows so the singleton crosses the DLL boundary. Having capture_service_export.h in place is what allows GenerateExportHeader to apply the correct __declspec(dllexport) / __declspec(dllimport) attributes when it is built as a shared library.

wangra-google: ok, so this is the part i was confused. As you mentioned: `by default, it is static`, so i guess it probably inherits some configurations of being dynamic from a global setting somewhere to make it dynamic? would it be better to explicit use SHARED here to make sure it is dyanimcally linked?

capture_service/CMakeLists.txt
wangra-google: to follow up on the above discussion, if this is PUBLIC, dive_gui_lib.dll/system_trace.dll will inherit the macro and try to export the singleton instead of importing it?

wangra-google (dismissed):

wangra-google (dismissed):

wangra-google (approved):

hysw (approved):
Approve for now, although this is not the best way of designing ABI.

PR #894, opened 2026-05-13
Use common warnings in more places

footballhead:
Create a CMake function to apply warnings more consistently. Apply to:

- capture_service
- gfxr_ext/decode
- gpu_time
- ui

Apply at using directory-based commands so that we can't forget to apply the config to a particular target.

This should catch more bugs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,7 +276,7 @@ endif()
 
 message(CHECK_START "Generate build files for third_party/abseil-cpp")
 list(APPEND CMAKE_MESSAGE_INDENT "  ")
-add_subdirectory(third_party/abseil-cpp)
+add_subdirectory(third_party/abseil-cpp SYSTEM)
 list(POP_BACK CMAKE_MESSAGE_INDENT)
 message(CHECK_PASS "done")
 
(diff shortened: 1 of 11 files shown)


hysw (approved):

CMakeLists.txt
```diff
@@ -276,7 +276,7 @@ endif()
 
 message(CHECK_START "Generate build files for third_party/abseil-cpp")
 list(APPEND CMAKE_MESSAGE_INDENT "  ")
-add_subdirectory(third_party/abseil-cpp)
+add_subdirectory(third_party/abseil-cpp SYSTEM)
```
wangra-google, line 279: oh, nice trick, i didn't know this before. to confirm, this is to suppress the warnings within abseil-cpp, right?

wangra-google (approved):

CMakeLists.txt
footballhead: Exactly. `ABSL_FLAG(absl::Duration)` was causing [C4312](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4312?view=msvc-170) in the guts of the Abseil library. With `SYSTEM`, CMake instructs MSVC to use [/external](https://learn.microsoft.com/en-us/cpp/build/reference/external-external-headers-diagnostics?view=msvc-170) for the Abseil headers with a different warning level. This suppresses the warnings from Abseil which we don't care about.

PR #896, opened 2026-05-14
Adds recursive force deletion of manifest folder path

GrantComm:
Changes the call to -rf prevents the command from failing if the folder has already been deleted or was never created.


wangra-google (approved):

footballhead (approved):
LGTM. This is consistent with other cleanup code

PR #897, opened 2026-05-19
Hide What-ifs by default

wangra-google:
- Hide the What-ifs feature behind Show Advanced Options as it is quite buggy

### DIFF
--- a/ui/trace_dialog.cpp
+++ b/ui/trace_dialog.cpp
@@ -459,7 +459,10 @@ void TraceDialog::OnShowAdvancedOptions(bool show)
     if (!show)
     {
         m_pkg_list_options = Dive::AndroidDevice::PackageListOptions::kDebuggableOnly;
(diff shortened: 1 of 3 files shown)


ui/trace_dialog.cpp
```diff
@@ -459,7 +459,10 @@ void TraceDialog::OnShowAdvancedOptions(bool show)
     if (!show)
     {
         m_pkg_list_options = Dive::AndroidDevice::PackageListOptions::kDebuggableOnly;
-        UpdatePackageList();
+        if (isVisible())
```
wangra-google, line 462: this is to fix a case when user looking at a trace without connecting to any devices and the trace dialog is closed, when the Show Advanced Option is unchecked, Dive is trying to connect to the device, and showing "Failed to connect".

GrantComm (approved):

angela28chen (approved):

PR #898, opened 2026-05-20
Fix Runtime What-ifs

wangra-google:
- Move `ForwardFirstAvailablePort()` from ` AndroidDevice::SetupDevice()` to the end of `AndroidApplication::RuntimeWhatIfSetup()`. (Before this fix, we call `adb root` after calling `ForwardFirstAvailablePort()` in ` AndroidDevice::SetupDevice()` which wipes out all existing adb forward rules associated with the device.)
- Disable runtime what-ifs before capturing

### DIFF
--- a/capture_service/device_mgr.cc
+++ b/capture_service/device_mgr.cc
@@ -514,6 +514,10 @@ absl::Status AndroidDevice::ForwardFirstAvailablePort()
 {
     for (int p = kFirstPort; p <= kFirstPort + kPortRange; p++)
     {
+        // Explicitly remove any existing forward for this port to avoid "port already in use"
+        // or stale rules issues after adb root.
+        Adb().Run(absl::StrFormat("forward --remove tcp:%d", p)).IgnoreError();
+
         auto res = Adb().RunAndGetResult(
             absl::StrFormat("forward tcp:%d localabstract:%s", p,
                             Dive::DeviceResourcesConstants::kUnixAbstractPath));
@@ -532,7 +536,6 @@ absl::Status AndroidDevice::SetupDevice()
 {
     if (m_runtime_what_if_enabled)
     {
-        RETURN_IF_ERROR(ForwardFirstAvailablePort());
         return absl::OkStatus();
     }
 
@@ -1398,6 +1401,11 @@ absl::Status AndroidDevice::TriggerScreenCapture(
         on_device_screenshot_dir / Dive::kCaptureScreenshotFile;
 
     std::string on_device_capture_screen_shot = full_capture_path.generic_string();
+    std::string on_device_capture_dir = full_capture_path.parent_path().generic_string();
+
+    // Ensure the directory exists before taking the screenshot.
+    RETURN_IF_ERROR(m_adb.Run(absl::StrFormat("shell mkdir -p %s", on_device_capture_dir)));
+    RETURN_IF_ERROR(m_adb.Run(absl::StrFormat("shell chmod 777 %s", on_device_capture_dir)));
 
     absl::Status ret =
         m_adb.Run(absl::StrFormat("shell screencap -p %s", on_device_capture_screen_shot));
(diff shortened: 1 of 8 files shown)


capture_service/device_mgr.cc
```diff
@@ -514,6 +514,10 @@ absl::Status AndroidDevice::ForwardFirstAvailablePort()
 {
     for (int p = kFirstPort; p <= kFirstPort + kPortRange; p++)
     {
+        // Explicitly remove any existing forward for this port to avoid "port already in use"
+        // or stale rules issues after adb root.
+        Adb().Run(absl::StrFormat("forward --remove tcp:%d", p)).IgnoreError();
```
footballhead: if you use `tcp:0` then adb will assign a port and print it to stdout. Should we prefer that behavior instead? It could simplify this code by removing the for loop as well as the possible error conditions

elviscapiaq: I took a look and that is correct. I agree with using just `tcp:0`.

wangra-google: good point, fixed

elviscapiaq (dismissed):

capture_service/device_mgr.cc
```diff
@@ -512,27 +513,31 @@ absl::StatusOr<std::vector<std::string>> AndroidDevice::ListPackage(PackageListO
...
     {
-        auto res = Adb().RunAndGetResult(
-            absl::StrFormat("forward tcp:%d localabstract:%s", p,
-                            Dive::DeviceResourcesConstants::kUnixAbstractPath));
-        if (res.ok())
+        Adb().Run(absl::StrFormat("forward --remove tcp:%d", m_port)).IgnoreError();
+        m_port = 0;
+    }
+
+    auto res = Adb().RunAndGetResult(absl::StrFormat(
```
footballhead: nit:

- consider using guard clauses to reduce nesting (https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html)
- consider using a more descriptive variable like `new_port` rather than `res`

```c++
auto new_port = Adb().RunAndGetResult(absl::StrFormat(
    "forward tcp:0 localabstract:%s", Dive::DeviceResourcesConstants::kUnixAbstractPath));
if (!new_port)
{
    return new_port.status();
}

if (!absl::SimpleAtoi(*new_port, &m_port)) {
    return absl::InternalError(
        absl::StrFormat("Failed to parse port from adb output: %s", *new_port));
}

return absl::OkStatus();
```

capture_service/device_mgr.h
```diff
@@ -227,7 +227,7 @@ class CAPTURE_SERVICE_EXPORT AndroidDevice
     std::unique_ptr<AndroidApplication> m_app;
     std::optional<GfxrCaptureSettings> m_gfxr_capture_settings;
     bool m_runtime_what_if_enabled = false;
-    int m_port = kFirstPort;
+    int m_port = 0;
```
footballhead: nit: avoid sentinels, consider a std::optional instead (https://abseil.io/tips/171) This has more semantic value which aids the reader in understanding when the value is initialized or not.

```suggestion
    std::optional<int> m_port;
```

I'd also suggest changing the `AndroidDevice::Port` accessor to return `std::optional<int>` instead as well. Consumers are currently making the assumption that the port is forwarded, which may not be true. Fixing these cases may be as simple as logging or returning failure when the optional doesn't have a value. This should make it more obvious when an error is due to a port that is not forwarded or not.

footballhead (dismissed):

footballhead (dismissed):

capture_service/device_mgr.cc
wangra-google: fixed

capture_service/device_mgr.h
wangra-google: fixed

footballhead (approved):

elviscapiaq (approved):

PR #899, opened 2026-05-21
Disable DerivePointerAlignment in clang-format

wangra-google:
- The default value is true, which means that if it detects that the file predominantly uses right-aligned pointers/references, it will silently override the `PointerAlignment: Left` setting.


hysw (approved):
Apparently that's a clang-format-18 issue. Top of tree clang-format already have it set to false with based on google style.

footballhead (approved):

PR #901, opened 2026-05-27
Ensures dialogs are always in the foreground

GrantComm:
Ensures dialogs are always in the foreground and can be brought to the foreground on mac at all times.

**Manual Testing:**
Linux, MacOS, Windows

Minimize main window, load gfxr capture, open analyze dialog, maximize main window. Ensure dialog is still in the foreground.

### DIFF
--- a/ui/analyze_dialog.cpp
+++ b/ui/analyze_dialog.cpp
@@ -84,7 +84,9 @@ AnalyzeDialog::AnalyzeDialog(ApplicationController& controller,
     : DeviceDialog(parent), m_controller(controller), m_available_metrics(available_metrics)
 {
     qDebug() << "AnalyzeDialog created.";
-
+#if defined(__APPLE__)
+    setWindowFlags(windowFlags() | Qt::Tool);
+#endif
     m_overlay = new OverlayHelper(this);
 
     // Metrics List
@@ -439,7 +441,7 @@ void AnalyzeDialog::OnAnalyzeCaptureStarted(const QString& file_path)
 
     UpdateDeviceList();
     // Open the dialog for users to initiate analysis
-    show();
+    ShowAndRaise();
 }
 
 //--------------------------------------------------------------------------------------------------
--- a/ui/analyze_dialog.cpp
+++ b/ui/analyze_dialog.cpp
@@ -84,9 +84,7 @@ AnalyzeDialog::AnalyzeDialog(ApplicationController& controller,
     : DeviceDialog(parent), m_controller(controller), m_available_metrics(available_metrics)
 {
     qDebug() << "AnalyzeDialog created.";
-#if defined(__APPLE__)
-    setWindowFlags(windowFlags() | Qt::Tool);
-#endif
+
     m_overlay = new OverlayHelper(this);
 
     // Metrics List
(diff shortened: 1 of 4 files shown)


ui/analyze_dialog.cpp
```diff
@@ -84,7 +84,9 @@ AnalyzeDialog::AnalyzeDialog(ApplicationController& controller,
     : DeviceDialog(parent), m_controller(controller), m_available_metrics(available_metrics)
 {
     qDebug() << "AnalyzeDialog created.";
-
+#if defined(__APPLE__)
```
wangra-google: just curious, why macos needs to be handled differently?

ui/analyze_dialog.cpp
```diff
@@ -84,7 +84,9 @@ AnalyzeDialog::AnalyzeDialog(ApplicationController& controller,
...
+    setWindowFlags(windowFlags() | Qt::Tool);
```
wangra-google: why do we apply this in both DeviceDialog and AnalyzeDialog constructor?

ui/device_dialog.cpp
```diff
@@ -80,6 +83,24 @@ void DeviceDialog::UpdateDeviceList()
...
+    {
+        showNormal();
+    }
+    else
+    {
+        if (isVisible())
+        {
+            hide();
+        }
+        show();
```
wangra-google, line 94: would this cause issues on what-ifs?
From what i understand, calling show() after calling hide() could trigger showEvent() which could trigger showEvent()? if that is the case, since WhatIfConfigureDialog inherits from DeviceDialog and overrides  showEvent  in what_if_configure_dialog.cpp:
```
    void WhatIfConfigureDialog::showEvent(QShowEvent* event)
    {
        QDialog::showEvent(event);
        PopulateLiveFilters();
    }
```
Whenever showEvent is triggered, it calls PopulateLiveFilters() which triggers GetLivePSOs() and GetLiveRenderPasses() which doesn't seem to be cheap.

wangra-google: actually do you need this part? i guess removing this part could prevent showEvent() being triggered?

```
if (isVisible())
{
    hide();
}
```

ui/analyze_dialog.cpp
GrantComm: The Qt::Tool window flag creates a utility window. On macOS, utility windows are explicitly layered to always float on top of their parent window. This flag guarantees that the TraceDialog, AnalyzeDialog, and the What-If dialog remain visible and instantly accessible above the main window

ui/analyze_dialog.cpp
GrantComm: Removed the duplication

ui/device_dialog.cpp
GrantComm: Good point! Refactored DeviceDialog::ShowAndRaise() so that if the dialog is already visible, raise() and activateWindow() are triggered to bring it back to the foreground

wangra-google (approved):

ui/analyze_dialog.cpp
angela28chen: If this is a no-op on windows and linux, what about just removing the `#if defined(__APPLE__)` and having it present for all platforms?

angela28chen (approved):

PR #905, opened 2026-06-30
Do not crash on unknown packet.

hysw:


dive_core/command_hierarchy.cpp
```diff
@@ -1288,11 +1307,20 @@ uint64_t CommandHierarchyCreator::AddPacketNode(const IMemoryManager& mem_manage
...
+                                       header.type7.count, append_extra_dwords, packet_info_ptr,
+                                       packet_node_index);
+            }
+            else
+            {
+                if (FirstTimeSeenBadOpCode(header.type7.opcode))
+                {
+                    DIVE_ERROR_MSG("PacketInfo for opcode %d not found\n", header.type7.opcode);
+                }
+            }
```
footballhead: nit: could reduce the level of nesting here

```suggestion
            else if (FirstTimeSeenBadOpCode(header.type7.opcode))
            {
                DIVE_ERROR_MSG("PacketInfo for opcode %d not found\n", header.type7.opcode);
            }
```

dive_core/command_hierarchy.cpp
```diff
@@ -35,6 +36,24 @@
 
 namespace Dive
 {
+namespace
+{
+bool FirstTimeSeenBadOpCode(uint32_t opcode)
+{
+    constexpr int kOpcodeBits = 7;
+    static std::bitset<(1 << kOpcodeBits)> seen = {};
+    if (opcode > seen.size())
```
wangra-google: shouldn't this be `opcode >= seen.size())`?

dive_core/common/common.h
```diff
@@ -21,6 +21,17 @@ void DIVE_LOG_INTERNAL(const char* file, int line, const char* format, ...);
 #include <assert.h>
 
 #define DIVE_ERROR_MSG(...) DIVE_LOG_INTERNAL(__FILE__, __LINE__, __VA_ARGS__)
+// Note: this is not thread safe, but it does not really matter to log more than once
+#define DIVE_ERROR_MSG_ONCE(...)                                \
+    do                                                          \
+    {                                                           \
+        static bool log_once_guard##__LINE__ = false;           \
```
wangra-google: wouldn't this create `log_once_guard__LINE__`? i guess you want to have the line number?

dive_core/common/common.h
```diff
@@ -21,6 +21,17 @@ void DIVE_LOG_INTERNAL(const char* file, int line, const char* format, ...);
 #include <assert.h>
 
 #define DIVE_ERROR_MSG(...) DIVE_LOG_INTERNAL(__FILE__, __LINE__, __VA_ARGS__)
+// Note: this is not thread safe, but it does not really matter to log more than once
```
wangra-google, line 24: i guess if we really want to have a clean log, we could use std::call_once?

dive_core/common/common.h
hysw: Good point, I just don't want the same variable shadowing, removed __LINE__.

dive_core/command_hierarchy.cpp
hysw: Done. Although technically this error condition will never reach due to opcode having only 7 bits.

dive_core/common/common.h
hysw: Not sure if this header is supposed to be compatible with C.

Otherwise, call_once is unnecessary as local lambda works just fine
```cpp
static int var = [](){ DIVE_LOG_INTERNAL(); return 0; }();
```

dive_core/command_hierarchy.cpp
footballhead: Good catch. I totally missed this

